PR #4, opened 2017-10-05
Proof of Concept of Python Process Forking on metaoptd (daemon)

dendisuhubdy:
- adding working CMakeLists.txt
- tested build nightly on Ubuntu 14.04
- added metaopt.cpp for client
- metaoptd.cpp for daemon service that does process fork on Linux"

TODO
- adding ring buffer data structure from LMAX Disruptor to metaoptd, to efficiently receive orders and executes experiments from `metaopt` client 
- adding UDP communication broadcast support
- adding MongoDB database

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,21 @@ addons:
     # install toolchains
     - gcc-5
     - g++-5
+    - gcc-6
+    - g++-6
+    - gcc-7
+    - g++-7
     - clang-3.6
+    - clang-5.0
+    # Boost
+    - libboost-chrono1.55-dev
+    - libboost-date-time1.55-dev
+    - libboost-test1.55-dev
+    - libboost-system1.55-dev
+    - libboost-filesystem1.55-dev
+    - libboost-timer1.55-dev
+    - libboost-program-options1.55-dev
+    - libboost-thread1.55-dev
 
 matrix:
   include:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,23 +1,20 @@
 cmake_minimum_required(VERSION 3.8.0)
 
-project (evolve)
+project (METAOPT)
 
 set(CMAKE_BUILD_TYPE Release)
 set (CMAKE_CXX_STANDARD 11)
 
-set(BOOST_INCLUDE_DIR /usr/local/Cellar/boost/1.65.1/include)
-set(BOOST_LIBRARY_DIR /usr/local/Cellar/boost/1.65.1/lib)
-
-find_package( Boost 1.65 COMPONENTS system REQUIRED )
+FIND_PACKAGE( Boost 1.55 COMPONENTS program_options REQUIRED )
 
 include_directories( 
-	${BOOST_INCLUDE_DIR}
-	${FOOD_SOURCE_DIR}/include
+	${Boost_INCLUDE_DIR}
+	${METAOPT_SOURCE_DIR}/include
 )
 
 link_directories(
-	${BOOST_LIBRARY_DIR}
-	${FOOD_SOURCE_DIR}/lib
+	${Boost_LIBRARY_DIR}
+	${METAOPT_SOURCE_DIR}/lib
 )
 
 set(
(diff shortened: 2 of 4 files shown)


dendisuhubdy:
@tsirif @bouthilx based on the separation between daemon and client, does this skeleton include what is imagined?

tsirif:
Some things I would like to clarify, just to be sure we are on the same page:

1. The daemon is going to run in the background at all times, whether there are workers currently hyperoptimizing or not.
2. A worker/client (`metaopt.cpp` executable) can start up independently and it will communicate with the same node's daemon to have it registered in the pool of workers that the user selects through the cmd line. If the pool id it requested for does not exist, then the daemon creates a one.
3. The daemon responds to the worker/client with information about how workers in the same pool are going to communicate and about the database they are going to use to push/pull stats.
4. For the inter-node case, different daemons spawned across machines are going to communicate with each in order to have the same information about currently existing pools and workers in the programme.
5. A worker/client forks a new process which will execute the user script's given in `metaopt`'s arguments. This will happen for each optimization episode.
6. The worker/client process needs to be decoupled from the user's script in the following ways:
    * Hyperoptimization method (input: history, output: next hparams) is going to be executed in the worker process.
    * Hyperoptimization objective function is going to be evaluated inside the user's script. Motivation for this are:
      1. To give ability to the user to write their own hyperobjective functions, apart from some already implemented basic ones that we will offer.
      2. Evaluation sets for hyperoptimization are going to reside in user's script only.
   * So, the user's script will receive only the next episode's hyperparameters as inputs and output statistics useful for hyperoptimization (mainly info about a current episode's resulting hyperoptimization function value, and/or hypergradients and/or etc), and probably some metadata/flags for the sake of the procedure (probably not to be logged in the db).
7. A point, I think is important to be aware of, is that we would like not to disturb the user from the usual workflow. So, that the same user script is able to run both with our software and as a standalone. This is because lots of people probably would like to debug/test new stuff they code locally first, without any purpose whatsoever for a hyperoptimization procedure. For this reason I propose that we support the following ways to pass input and get output:
    * **input**, worker process must understand which way of passing inputs is used by the procedure:
       1. offering a Python API:
           * for declaring hyperparameters to be used and their search space, etc.
           * for pulling hyperparameters from the database.
       2. otherwise the user can still have a common way to pass hyperparameter values to his script (cmd args, yaml, configparser files, json), but instead of passing a specific value of a hparam, he would describe in that place the options for that specific hparam (search space, etc). These arguments/files are going to be given initially to `metaopt` and we will use them for initialization and declaration of the task's hparams and as a template for passing specific suggested hparam values at each training episode.
   * similarly for **output**:
      1. offering a Python API, which internally:
          * either pushes info about the hyperobjective function to the db.
          * or prints this info to a temporary file, named uniquely for this user script execution, e.g. `mopt_<script pid>.out`. These 2 options depend on how the user has preferred to take the input.
8. We should probably have to reserve some keywords for the logging elements that are useful for the hyperoptimization:
    * **inputs**: some thing like [these from hyperopt](https://github.com/hyperopt/hyperopt/blob/master/hyperopt/hp.py) when declaring variables through the API or when searching for hparams in args/configs.
   * **outputs**: a key for hyperobjective (`metaval`??) value and a key for hypergradients (`metagrad`??) to use when calling the Python API. The rest of the keys can be whatever the user likes to name the things to be logged.

What do you think about these? @bouthilx

dendisuhubdy:
I agree with all of your points up there @tsirif. One remark though about execution of this plan, on you point

> Hyperoptimization objective function is going to be evaluated inside the user's script. Motivation for this are:
> 
> - To give ability to the user to write their own hyperobjective functions, apart from some already implemented basic ones that we will offer.
> 
> - Evaluation sets for hyperoptimization are going to reside in user's script only.
> 
> So, the user's script will receive only the next episode's hyperparameters as inputs and output statistics useful for hyperoptimization (mainly info about a current episode's resulting hyperoptimization function value, and/or hypergradients and/or etc), and probably some metadata/flags for the sake of the procedure (probably not to be logged in the db).

For that, we need a Python parser or a call to the Python.h, in which the user specifies in his script a class or a function to do hyperparameter search. Call it `def grid` or `def random` which means when our parser hits that line it triggers the backend to automatically do the grid/random search.

Then regarding your concern that it would alter the users (researchers) workflow if we introduce more code inside his/her script, there's a simple solution. When he/she would want to execute the script, he/she will `python script.py` which we could just return no callback from our Python package because it's in non-hyperopt mode.

Thoughts? @tsirif @bouthilx

PR #6, opened 2017-10-06
patching error Boost program options not found

dendisuhubdy:
passing in Boost_libraries instead in CMakeLists


dendisuhubdy:
Only one build check left, merging.

Checking at https://travis-ci.org/mila-udem/metaopt/builds/284274516

PR #7, opened 2017-11-08
Setup repo for install+devel and add config parsing module

tsirif:


setup.py
```diff
@@ -1,66 +1,82 @@
...
-extra_compile_args = ['-std=c++1z', '-Wno-write-strings',
-                      # Python 2.6 requires -fno-strict-aliasing, see
-                      # http://legacy.python.org/dev/peps/pep-3123/
-                      '-fno-strict-aliasing']
-
-cmdclass = {
-    'build': build,
-    'develop': develop,
-    'install': install,
-    'clean': clean,
```
dendisuhubdy: I tried clean, and it works fine :)

setup.py
```diff
@@ -1,66 +1,82 @@
...
+    license=metaopt.__license__,
+    author=metaopt.__author__,
+    author_email=metaopt.__author_email__,
+    url=metaopt.__url__,
+    packages=find_packages('src'),
+    package_dir={'': 'src'},
+    include_package_data=True,
+    data_files=find_data_files(),
+    scripts=['scripts/mopt'],
+    install_requires=['six', 'PyYAML'],
```
dendisuhubdy: We're finally going for YAML?

setup.py
```diff
@@ -1,66 +1,82 @@
...
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 2',
+    'Programming Language :: Python :: 3',
+    'Topic :: Scientific/Engineering',
+    'Topic :: Scientific/Engineering :: Artificial Intelligence',
+    ]
+
+
+if __name__ == '__main__':
+    setup(**setup_args)
```
dendisuhubdy: Sounds good

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+    os.path.join(metaopt.dirs.site_data_dir, 'moptconfig.yaml.example'),
+    os.path.join(metaopt.dirs.site_config_dir, 'moptconfig.yaml'),
+    os.path.join(metaopt.dirs.user_config_dir, 'moptconfig.yaml')
+    ]
+
+# list containing tuples of
+# (environmental variable names, configuration keys, default values)
+ENV_VARS_DB = [
+    ('METAOPT_DB_NAME', 'name', 'MetaOpt'),
+    ('METAOPT_DB_TYPE', 'type', 'MongoDB'),
```
dendisuhubdy: In the case `PostgreSQL` or `Mysql` would be chosen, this would represent the class name inside `metaopt/io/database`?

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+                for vk, vv in six.iteritems(v):
+                    expconfig[k][vk] = vv
+            else:
+                expconfig[k] = v
+
+    for k, v in six.iteritems(cmdargs):
+        if v is not None:
+            expconfig[k] = v
+
+    return expconfig
```
dendisuhubdy: Looks good

scripts/mopt
```diff
@@ -0,0 +1,56 @@
...
+from __future__ import print_function, absolute_import
+import getpass
+import datetime
+import pprint
+pp = pprint.PrettyPrinter(indent=4)
+
+from metaopt.io import parsing
+
+if __name__ == '__main__':
+    starttime = datetime.datetime.utcnow()
```
dendisuhubdy: yep using `utcnow()`!  :)

bouthilx:
@tsirif I started a review but I won't have time to finish before the meeting. I'll post the review this afternoon.

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # get some defaults
+    defcfg['exp_name'] = DEF_CMD_EXP_NAME[1].format(
+        user=user,
+        starttime=starttime.isoformat())
+    defcfg['max_trials'] = DEF_CMD_MAX_TRIALS[0]
+    defcfg['pool_size'] = DEF_CMD_POOL_SIZE[0]
+
+    # get default options for some managerial variables (see `ENV_VARS`)
+    for signif, evars in six.iteritems(ENV_VARS):
```
bouthilx: Significant seems confusing. At least if confused but me and Fred. Any other idea how to name this?

bouthilx: Also `signif` and `evars` names needs to much though to decrypt. `significant` and `env_vars` are more explicit.

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+                if cfg is None:
+                    continue
+                # implies that yaml must be in dict form
+                for k, v in six.iteritems(cfg):
+                    if k in ENV_VARS:
+                        for vk, vv in six.iteritems(v):
+                            defcfg[k][vk] = vv
+                    else:
+                        defcfg[k] = v
+        except IOError as e:  # default file could not be found
```
bouthilx: Do we really want to log this as debug(e) and continue on?

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+            log.debug(e)
+        except AttributeError as e:
+            log.warn("Problem parsing file: %s", cfgpath)
+            log.warn(e)
+
+    return defcfg
+
+
+def env_vars(config):
```
bouthilx: Oh I see, you have a function named `env_vars`, hence the `evars`. Well, `env_vars` is a nice variable name, not a function name. :wink: The doc states it clearly: "Fetches env vars", so was about `fetch/get_env_vars`?

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+       * 'optimizer': {optimizer module name : method-specific configuration}
+       * 'dynamic': {dynamic module name : method-specific configuration}
+
+       .. seealso:: Method-specific configurations reside in `/config`
+
+    """
+    expconfig = deepcopy(config)
+
+    for cfg in (dbconfig, cmdconfig):
+        for k, v in six.iteritems(cfg):
```
bouthilx: Isn't there a similar loop inside `default_options`?

src/metaopt/io/parsing.py
tsirif: Not really, I put it to check things out

bouthilx: By not really, you mean we don't really want to log.debug(e)? But we want the process to go on? 

Oh ok I think I get it. That means the default value is not defined in the global configuration but we won't complain unless it is not defined in the experiment configuration. Is that it? Then yeah log.debug(e) is good to have. I was confused because I though that was a major exception.

scripts/mopt
```diff
@@ -0,0 +1,56 @@
...
+"""
+from __future__ import print_function, absolute_import
+import getpass
+import datetime
+import pprint
+pp = pprint.PrettyPrinter(indent=4)
+
+from metaopt.io import parsing
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
tsirif: So:
1. instead of using 'kwarg' `scripts=['scripts/mopt']`, you say that we should use:
```
entry_points={
        'console_scripts': ['mopt = metaopt.scripts.main:main']
    }
```
2. and that that I should move `/scripts` folder inside `/src/metaopt`.
3. `main` as a named function

Questions:
* Should `metaopt.scripts` be enlisted as a package in `/setup.py`?
* What is the difference between using `scripts=[]` and using `entry_points={'console_scripts':[]}`??
* Would that remove any need for using `scripts=[]` at all? In which cases are they still useful?

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+log = logging.getLogger(__name__)
+
+################################################################################
+#                 Default Settings and Environmental Variables                 #
+################################################################################
+
+# Default settings for command line arguments (option, description)
+DEF_CMD_MAX_TRIALS = (-1, 'inf/until preempted')
+DEF_CMD_POOL_SIZE = (10, str(10))
+DEF_CMD_EXP_NAME = (None, '{user}_{starttime}')
```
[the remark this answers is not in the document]
tsirif: Yeah I was wondering about that. Maybe it should be the only required argument and maybe we should enforce passing it only as a cmd arg or through cmd provided moptconfig (perhaps not even that). I do not see benefits from being able to pass it through default files.

If someone wants to resume an experiment, should explicitly type somewhere the appropriate name. Do you agree?

scripts/mopt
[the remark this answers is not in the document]
bouthilx: scripts=[] would not be required anymore unless we add others scripts. One of the main advantage for entry_points=[] vs scripts=[] is that we can more easily test and reuse functions in the script. I find scripts=[] useful for things like bash scripts. For python, it's better if we can access functions directly and interpret then as scripts using entry_points.

src/metaopt/io/parsing.py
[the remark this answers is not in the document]
bouthilx: That's another thing I'm not sure about. I don't think experiments should be definable *globally*. If that wasn't the case, then the user could be forced to define the experiment name in either the experiment configuration file or command-line.

I'm not sure yet what the interface would be to resume experiments but if the users needs to provide the experiment name which is by default user_starttime then that will certainly be annoying for the user to look up and provide.

src/metaopt/io/parsing.py
```diff
@@ -0,0 +1,257 @@
...
+
+    if args['verbose'] == 1:
+        logging.basicConfig(level=logging.INFO)
+    elif args['verbose'] == 2:
+        logging.basicConfig(level=logging.DEBUG)
+
+    return args, config
+
+
+def default_options(user, starttime):
```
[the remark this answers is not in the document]
tsirif: > To make things clear,...MetaOpt configuration optional.

That's my intention too. I think it successfully does that.

> For this distinction, I would... script outputs, configurations, etc.

Yeah, I could not think of a better name 😛. Probably `resolveconfig` (even within `metaopt.io`) would be more appropriate I think. Do you agree?

Regarding config files, I was thinking something like this:

```yaml
database: {
    # some global definitions
}
resources: {
    # some global definitions
}
optimizer: {
    # `Optim`, `Dynamic` stuff here
}
# other common experiment metadata follow here (each one overrides the previous one in hierarchy)
```

I would not like parameters of [black-box] functions to be coupled to optimization procedures. So that the same procedure definition, i.e. cmd-provided moptconfig file + default files, can be re-used across different problems, i.e. userscript/[black box] function + hyperparameters/function domain. Although, in theory some optimization procedures are more appropriate for some function classes, I think we should act as if we were ignorant towards this and allow the user to re-use definitions according to their beliefs for "appropriate" correspondences.

I think that user provided defaults and config files are meant to represent his particular beliefs about best optimization procedures (`optimizer` key), his usual resource preferences (`resources` key - such as scheduler args for which queue he usually submits his jobs), and personal workflow (`database` key - mby he keeps a personal database).

I am pretty positive that we can be config-type agnostic towards hparam space definition, so that we let user keep whatever they used to use, and use their config file both as declaration and as template + their script's cmd args. Yes, I agree that any hparams declaration found will definitely be logged in experiment's db entry and thus there would be a dictionary representing that at some point.

src/metaopt/io/parsing.py
tsirif: I thought of this as `signifier`. I took note of that, I will try to make more descriptive names in general.

src/metaopt/io/parsing.py
tsirif: When I was writing the flow in `'__main__'`, I had routines reading like: `parsing.env_vars(...)` so that would make sense there.

src/metaopt/io/parsing.py
tsirif: Indeed

tsirif: That loop is over 'configuration paths'. I think it's an overkill to abstract it to a function.

src/metaopt/io/parsing.py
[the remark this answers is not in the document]
tsirif: If `'exp_name'` key hits an entry in database, then we can suppose that an experiment is to be resumed (or extended if they want to use the same history of trials, but different optimization). This will work here, because database overrides defaults, but does not override cmd-provided moptconfig.

I think I will let it as it is for now and we can switch it off, if we find it not to be convenient. (I am kinda biased, in that all keyword arguments should be optional in a cli)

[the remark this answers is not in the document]
bouthilx: I agree the majority of the options should be optional, but there are some that should not be. I see it like a git commit message. Users can write crappy commit messages, to their own misfortune, but git developers are not encouraging this by making the argument optional. Experiments name/description seems similar to me. We could even have a similar mechanism, opening an editor to write one-line name with a description below if exp_name was not given explicitly by command-line argument or from configuration files. That would be a feature for later however.

src/metaopt/io/parsing.py
bouthilx: Again, semantically, `env_vars` reads like a parameter of parsing, not like a method/function.

src/metaopt/io/parsing.py
bouthilx: Fine. I just wanted to make sure there is no avoidable duplications.

src/metaopt/io/parsing.py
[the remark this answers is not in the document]
bouthilx: > Yeah, I could not think of a better name :stuck_out_tongue:. Probably resolveconfig (even within metaopt.io) would be more appropriate I think. Do you agree?

Yes, resolve_config would be even better. :wink: 

It think there is two concepts here which are currently entangled but should not dictate the design of each other. There is the concept of *hierarchical* or *distributed* configuration which enables global definitions of DB, resources and optimizer to favor reuse over many experiments. The second concept is the support for any type of text-based configuration file for the hyper-parameter space. 

### *Hierarchical* or *distributed* configuration

I do also believe having support for flexible configuration is important. However, I also believe default options for Optimizer will address most of the use cases. Furthermore, from my experience, people tend to keep experiment definitions centralized and prefer to copy-paste them and edit rather than refactoring it. By centralizing, I mean having all the configuration in a single file rather than many layers of global to exp-specific definitions. We could argue like for the git-commit message/exp-name what is in their best interest. In this case, the main argument for keeping config centralized is for having something easy to reproduce and also a simple reference for later. DB and ressources config and not crucial for reproducibility, for optimizer is. I find this workflow well adapted to research and I don't think we should discourage it.

I should also note that some argument of the optimizer are very problem specific. There is for exemple the type of optimizer (minimize, maximize), the type of task to optimizer (noiseless, gaussian, etc), some constraints on the hyper-parameters like in [spearmint's tasks](https://github.com/HIPS/Spearmint/blob/master/examples/constrained/config.json#L19-L32). All of them can be optional, with default values covering a large part of the use cases, but when users need to set them, that is unlikely to be very reusable. 

To be clear, I would not recommend to drop support for hierarchical/distributed configuration, but it is my belief that it should not be enforced on users. They should be able to use a single config file if they prefer without the complexity hassle of hierarchical/distributed config. It should be transparent. I believe your code presently support this beside hyper-parameter space, and this brings us to next point. 

### Support for any text-based hyper-parameter space definition file type

This feature is very nice. I can think of python scripts for instance, something I've seen often in others pipelines, that could be templates for hyper-parameters settings. Very neat. This feature however should not force us to keep hyper-parameter space definitions outside of the configuration file. Again, I believe a very common use case will be to simply use the experiment configuration file with the optimizer definition beside the hyper-parameter space definition. This would not cause any problem for agnostic support for text-based configs. We can simply update the `hp_space` dictionary from the experiment configuration file with a h-p space configuration file given in command-line. 

All in all, those two concepts are nice and I agree should be part of the project. They are however features that a minimal viable product can live without. In case of the hierarchical configuration, it seems you are close to having it built entirely already. In that case it might be worth finishing right now. For the agnostic h-p space parser I would leave it for later. You should open new issues for each features like this and give them a detailed description so that we do not lose the idea.

bouthilx:
I'll try to review tomorrow, otherwise it will be on Monday.

src/metaopt/cli.py
```diff
@@ -0,0 +1,84 @@
...
+
+"""
+from __future__ import absolute_import
+
+import datetime
+import getpass
+import logging
+import sys
+
+from metaopt import resolve_config as resolvconf
```
bouthilx: Why do you rename it? Would we agree to follow [pep8 naming's convention](https://www.python.org/dev/peps/pep-0008/#function-names)?

tsirif: Yes, we will follow pep8 naming convention. However, this is not against pep8. I noticed that in many projects (even in Python's stl) ppl really avoid underscored names if it is clear enough, e.g. `configparser`, `argparse`, `importlib`, `modulefinder`, `pkgutil`, etc.

Eitherways, `resolvconf` is a commonly recognisable abbreviation due to being used in linux jargon (https://duckduckgo.com/?q=resolvconf&ia=web). I renamed it to be shorter when I use it 😛 .

tsirif: I actually prefer that modules avoid having underscores in their names, but this is due to my personal linting preferences. 😝

bouthilx: Yes pep8 is flexible 

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.

However I don't like relying on outside's context for variable name clarity. I must say I do not understand the argument for short variable names nowadays. Every text editor, even vim, has auto-completion add-ons so we virtually never type a full variable name more than once. But god knows how many times we read them. :stuck_out_tongue:

PR #8, opened 2017-11-12
MongoDB - class and tests

dendisuhubdy:


src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
+import json
+import bson
+import pymongo
+
+from pymongo import MongoClient
+from datetime import datetime
+from abstractdb import AbstractDB
+
+try:
+    from StringIO import StringIO
```
bouthilx: In which case StringIO is not available?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+class MongoDB(AbstractDB):
+    def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
+                                            username=None, password=None):
+        self.dbhost = str(dbhost)
+        self.dbname = str(dbname)
+        self.username = str(username)
+        self.password = str(password)
+
+        # we might have a different port for the database port
+        # so here we do a check if the port is there
```
bouthilx: Is that a TODO comment?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.dbport = str(dbport)
+
+        # set the connection string
+        # this should be mongodb://mongodb0.example.net:27017
+        # according to MongoDB documentation
+        # https://docs.mongodb.com/getting-started/python/client/
+
+        self.uri = 'mongodb://' + self.dbhost + ':' + self.dbport
```
bouthilx: Building the DB url (you typed uri by the way, was it intentional?) can get complicated if we support many options. That would deserve it's own method. It could be worth making it a dynamic `@property`.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.conn = self.initiate_connection()
+        self.db = self.conn[self.dbname]
+
+    def initiate_connection(self):
+        try:
+            conn = MongoClient(self.uri)
+            self.conn = conn
+        except pymongo.errors.ConnectionFailure:
+            raise Exception("Could not connect to server")
```
bouthilx: Could ConnectionFailure connect important information that we would like to pass on?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+            #print("Connection Successful!")
+            return True
+        else:
+            #print("Please check your connection")
+            return False
+
+    def close_connection(self):
+        try:
+            close = self.conn.close()
+        except pymongo.errors.ConnectionFailure:
```
bouthilx: Idem

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+            raise Exception("Could not close socket connection")
+        return close
+
+    def is_locked(self):
+        try:
+            is_locked = self.conn.is_locked()
+        except pymongo.errors.ConnectionFailure:
```
bouthilx: Idem

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+            raise Exception("Could not determine if connection is locked")
+        return is_locked
+
+    def unlock(self):
+        try:
+            unlock = self.conn.unlock()
+        except pymongo.errors.ConnectionFailure:
```
bouthilx: Idem

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+            raise Exception("Could determine unlock due to connection")
+
+    def insert(self, experiment_name, trial_field, dictionary, field_filters=None):
+
+        # check field_filters value
+        if field_filters is None:
+            field_filters = {}
+
+        dbcollection    = self.db[experiment_name][trial_field]
```
bouthilx: pep8

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def unlock(self):
+        try:
+            unlock = self.conn.unlock()
+        except pymongo.errors.ConnectionFailure:
+            raise Exception("Could determine unlock due to connection")
+
+    def insert(self, experiment_name, trial_field, dictionary, field_filters=None):
+
+        # check field_filters value
```
bouthilx: Useful comment?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+            raise Exception("Could not determine if connection is locked")
+        return is_locked
+
+    def unlock(self):
+        try:
+            unlock = self.conn.unlock()
+        except pymongo.errors.ConnectionFailure:
+            raise Exception("Could determine unlock due to connection")
+
+    def insert(self, experiment_name, trial_field, dictionary, field_filters=None):
```
bouthilx: trial_field is used as a collection name. It's confusing. Is trial_field supposed to be a string inside the collection `experiment_name` pointing to a dictionary of trials ({id: trial})? Shouldn't that trial_field be a `CONSTANT` variable then? 

Nevertheless, I'm not sure DB classes should have experiment and trial semantic tied to them. From Christos' UMLs, it seems to me that Experiment and Trial should have their own interface, with each using AbstractDB internally (which can be MongoDB or PostgreSQL, etc). In some way, workers and producers do not need to *know* they are communicating with a DB, anything could be hidden behind Experiment and Trial objects; Database, Sockets, whatever. @tsirif Any thoughts on that?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+        return result
+
+    def update(self, experiment_name, trial_field, update_dictionary, field_filters=None):
+        # note here that update needs to have
+        # MongoDB comparison operations
+        # for example update={"$set": {'hyperparam_name': 'hidden units',
+        #                               'hyperparam_type': 'int',
+        #                               'hyperparam_value': '200'}
+
+        # check field_filters value
```
bouthilx: Idem

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,137 @@
...
+        if field_filters is None:
+            field_filters = {}
+
+        dbcollection    = self.db[experiment_name][trial_field]
```
bouthilx: pep8

src/metaopt/tests/mongodb_test.py
```diff
@@ -0,0 +1,54 @@
...
+                "value": "400"
+            }
+        }
+
+    def mongodb_connection_test(self):
+        conn = self.db.check_connection()
+        return conn
+
+    def mongodb_insert_test(self):
+        result = self.db.insert(self.experiment, self.trial, self.query_test, field_filters=None)
```
bouthilx: ok I see here what trial_field was intended to be. I'd prefer to combine them inside a single dict inside experiments otherwise they kind of *pollute* the experiment row.

src/metaopt/tests/mongodb_test.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+
+from metaopt.io.database.mongodb import MongoDB
+
+class MongoDB_test():
```
bouthilx: Why not using unittest.TestCase and tests directly inside the class rather than gluing them together in a function later?

src/metaopt/io/database/mongodb.py
dendisuhubdy: In Python 3.0 `StringIO` and `cStringIO` modules are gone. See https://stackoverflow.com/questions/11914472/stringio-in-python3

src/metaopt/io/database/mongodb.py
dendisuhubdy: I leaved it for a IF block before, I should've removed it. Noted. Thanks

src/metaopt/io/database/mongodb.py
dendisuhubdy: I used the full URI until our discussion last week with Fred that for current localhost testing purposes we shouldn't put a default username and password because it would get passed on. Anyway, I'll comment out this URI and use the one following this line instead.

src/metaopt/io/database/mongodb.py
dendisuhubdy: Like the enumeration of what error type it would be? So I guess `raise Exeception("Exception caught %s, e)` instead?

dendisuhubdy: I don't see how we would need it? If `pymongo.errors.ConnectionFailure as e` we pass on (like in C), the assert value for the error is null, thus outside of the `try exception` block we would always return a valid connection object.

src/metaopt/io/database/mongodb.py
bouthilx: Ah! You are speaking to an ignorant of python3 indeed.

src/metaopt/io/database/mongodb.py
```diff
@@ -26,72 +26,77 @@ def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
...
         self.username = str(username)
         self.password = str(password)
+        self.dbport = str(dbport)
 
-        # we might have a different port for the database port
-        # so here we do a check if the port is there
+        # try to initialize connection
+        # setting this as a function allows to make sure reconnection
+        # is properly done in the case of an unreliable network
+        self.uri = property(self.get_uri, self.set_uri, \
```
bouthilx: Why not using this syntax?

```python
@property
def uri(self):
    # ...

@uri.setter
def uri(self, value):
    # ...
```

src/metaopt/io/database/mongodb.py
```diff
@@ -26,72 +26,77 @@ def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
...
-        # try to initialize connection
-        # setting this as a function allows to make sure reconnection
-        # is properly done in the case of an unreliable network
-
-        self.conn = self.initiate_connection()
-        self.db = self.conn[self.dbname]
+        uri = "mongodb://%s:%s@%s:%s" % (quote_plus(self._username), \
+                                         quote_plus(self._password), \
+                                         self._dbhost, self._dbport)
+        self._uri = uri
```
bouthilx: `uri` passed to the setter is not used. Do we support editing dbhost, dbport, etc, or editing uri? Or both? I think we should support both, but for simplicity I would only support setting dbhost, dbport and others for now. If we allow setting the uri, than we need to parse it and set dbhost and others accordingly. We can keep that for later.

src/metaopt/io/database/mongodb.py
```diff
@@ -26,72 +26,77 @@ def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
...
-            raise Exception("Could not connect to server")
+        except pymongo.errors.ConnectionFailure as e:
+            raise Exception("Could not connect to server: %s", e)
         # return connection string for checking purposes
         return conn
 
     def check_connection(self):
         if self.db:
-            #print("Connection Successful!")
+            print("Connection is successful!")
```
bouthilx: logger.info please :)

src/metaopt/io/database/mongodb.py
```diff
@@ -26,72 +26,77 @@ def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
...
             return True
         else:
-            #print("Please check your connection")
+            print("Please check your connection")
```
bouthilx: idem

src/metaopt/io/database/mongodb.py
```diff
@@ -26,72 +26,77 @@ def __init__(self, dbhost='localhost', dbname='metaopt', dbport=27017, \
...
             return False
 
     def close_connection(self):
         try:
             close = self.conn.close()
-        except pymongo.errors.ConnectionFailure:
-            raise Exception("Could not close socket connection")
+        except pymongo.errors.ConnectionFailure as e:
```
bouthilx: I was not saying we should do this. I was just wondering what kind of information is passed through ConnectionFailure. Is there any relevant information that we should not lose? Additionally, I believe raising Exception like this might break the stack trace. Look at [this](https://github.com/mila-udem/blocks/blob/master/blocks/utils/utils.py#L32) for instance. For simplicity, could the ConnectionFailure error message be good enough to just let it pass with no try-except clause?

bouthilx:
@dendisuhubdy It would be preferable to use informative message for the commits. Taken out of context, "adding suggested changes from code review" doesn't mean anything. Even then, the first though I had reading the commit's name was "What changes did I asked exactly? I don't remember". :slightly_smiling_face: 

A good reading about commit messages: http://who-t.blogspot.ca/2009/12/on-commit-messages.html

src/metaopt/io/database/mongodb.py
bouthilx: By passing on, I meant passing the error information in the raised error so that it finds its way in the stack trace printed to the user.

src/metaopt/io/database/mongodb.py
dendisuhubdy: Gotcha :+1:

src/metaopt/io/database/mongodb.py
dendisuhubdy: Hmmmn conn itself contains enough information to let the user know if things aren't right

src/metaopt/tests/mongodb_test.py
dendisuhubdy: Gotcha I adjusted this per assert in the functions inside the class

src/metaopt/tests/mongodb_test.py
tsirif: I am not sure that I understand what you mean. Could you give 2 cases?

src/metaopt/io/database/mongodb.py
tsirif: http://python-future.org/compatible_idioms.html

You can use `from io import StringIO`.

In general, these sites are good references along with the `six` module:

http://python-future.org/compatible_idioms.html
https://docs.python.org/3/howto/pyporting.html
https://pythonhosted.org/six/

src/metaopt/io/database/mongodb.py
tsirif: I think that the `@property` is a good idea. I am not sure how we should ask for the password. There are multiple ways of authentications here: https://api.mongodb.com/python/current/examples/authentication.html

Also, there this useful stl in 2.7, >=3.5: https://docs.python.org/3/library/getpass.html

src/metaopt/io/database/mongodb.py
tsirif: It is not a good idea to raise or catch pure Exception objects; there is no way to differentiate between cases. I agree with Dendi's thought to reraise an exception here, but it should be an Exception of our own w.r.t. common database problems, so that we can wrap in the future other libraries as well.

However, I think we should still keep the traceback though, Pythonic stuff. `raise wrapped_e from e` clauses are the case here, but unfortunately however beautiful (I find) they are, they are Py3 only. But there is this:
http://python-future.org/compatible_idioms.html#raising-exceptions (`from future.utils`)
https://pythonhosted.org/six/#six.raise_from (`from six` although they do not implement the correct impl of this functn; there's a current PR about this)

For truthful reference: https://www.python.org/dev/peps/pep-3134/#rationale
`.. seemore:: difference between implicit and explicit exception chaining :P`

src/metaopt/tests/mongodb_test.py
tsirif: No need for subclassing in general: https://docs.pytest.org/en/latest/getting-started.html

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
         # setting this as a function allows to make sure reconnection
         # is properly done in the case of an unreliable network
 
         self.conn = self.initiate_connection()
         self.db = self.conn[self.dbname]
 
+        if __debug__:
+            self.logger = logging.getLogger('database')
+            self.fh = logging.FileHandler('database.log')
+            self.fh.setLevel(logging.DEBUG)
```
bouthilx: I understand why we would want a detailed log for the database but do we really want to have DEBUG by default? I guess that's like system's log that are always very verbose by default.

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
+                                        %(levelname)s - %(message)s')
+            self.fh.setFormatter(formatter)
+            self.logger.addHandler(self.fh)
+
+    @property
+    def uri(self):
+        return self._uri
+
+    @uri.setter
+    def uri(self):
```
bouthilx: That's not a correct setter interface. It should be 
```python
@uri.setter
def uri(self, value):
```
But that wouldn't make sense since we set it internally anyway. Maybe we just shouldn't have a setter.

src/metaopt/io/database/mongodb.py
bouthilx: @tsirif I referred to this [example](https://github.com/mila-udem/blocks/blob/master/blocks/utils/utils.py#L32) in another comment for a way to reraise an exception. I agree we should not reraise it as a pure Exception. But again, I'm not sure we would need to reraise it. Isn't the ConnectionFailure explicit enough?

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
         return conn
 
     def check_connection(self):
         if self.db:
-            #print("Connection Successful!")
+            self.logger.info("Connection Successful!")
             return True
         else:
-            #print("Please check your connection")
+            self.logger.info("Please check your connection")
```
bouthilx: It seems weird to me to print logging info inside the check_connection like this. If False, I guess the code calling the function might raise an error or something. Why should the check_connection print logging info then? Maybe that would be better to set them as debug. I don't know. @tsirif What do you think?

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
             return False
 
     def close_connection(self):
         try:
             close = self.conn.close()
         except pymongo.errors.ConnectionFailure:
             raise Exception("Could not close socket connection")
+            self.logger.info("Could not close socket connection")
```
bouthilx: This will never get executed

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
         return close
 
     def is_locked(self):
         try:
             is_locked = self.conn.is_locked()
         except pymongo.errors.ConnectionFailure:
             raise Exception("Could not determine if connection is locked")
+            self.logger.info("Could not determine if connection is locked")
```
bouthilx: This will also never get executed

src/metaopt/io/database/mongodb.py
```diff
@@ -19,95 +20,121 @@
...
-        # for example update={"$set": {'hyperparam_name': 'hidden units',
-        #                               'hyperparam_type': 'int',
-        #                               'hyperparam_value': '200'}
+        # for example update=
+        #   {
+        #       "$set": {'hyperparam_name': 'hidden units',
+        #               'hyperparam_type': 'int',
+        #               'hyperparam_value': '200'
+        #               }
+        #   }
```
bouthilx: That would be worth turning into a doctstring

src/metaopt/io/database/mongodb.py
dendisuhubdy: Doesn't the `if __debug__:` block only run if we run the script with the flag `python -O abc.py`? So it wouldn't run by default?

src/metaopt/io/database/mongodb.py
dendisuhubdy: Gotcha, just a setter would make more sense.

dendisuhubdy: I'm kinda confused if we do it this way what `value` should we pass, the value of the URI itself?

src/metaopt/io/database/mongodb.py
dendisuhubdy: Cool

src/metaopt/io/database/mongodb.py
bouthilx: That's the thing. It would not make sense and that's why I'm saying that maybe there should not be any setter method for uri.

dendisuhubdy: So there would only be a `@property`? I'm not that familiar with Python yet, so does it work without a setter? 

In the other case would a `self.uri` as the previous commits be preferable?

src/metaopt/io/database/mongodb.py
dendisuhubdy: Either both, I added it to the `__init__.py` in `utils.py` and redid all of the `exception` blocks.

src/metaopt/io/database/mongodb.py
bouthilx: It works without a setter. Using property can be more flexible if for instance the property is built using other arguments which can change over time. Since uri is built using other arguments I think we should keep @property.

src/metaopt/tests/mongodb_test.py
bouthilx: No but the assert methods coming with unittest.TestCase are quite useful. :slightly_smiling_face:

src/metaopt/tests/mongodb_test.py
bouthilx:
```
{ name: dummy,
  # some other properties,
  trial_1: {
      # properties of trial_1
   },
  # ...
  trial_n: {
      # properties of trial_n
   }
  # some other properties,
}
```

vs

```
{ name: dummy,
  # some other properties,
  trials: [
      {
          id: 1, 
          # properties of trial #1 
      },
      # ...
      trial_n: {
          # properties of trial_n
      }
   },
  # some other properties,
}
```

Trials could be either a list or a dictionary

src/metaopt/io/database/abstractdb.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.conn = self.initiate_connection()
+        self.db = None
+
+        self.logger = logging.getLogger('database')
+        self.fh = logging.FileHandler('database.log')
+        formatter = logging.Formatter('%(asctime)s - %(name)s -'
+                                      '%(levelname)s - %(message)s')
+        self.fh.setFormatter(formatter)
+        self.logger.addHandler(self.fh)
```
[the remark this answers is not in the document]
dendisuhubdy: Because it will be used if we would want to have a PostgreSQL or MySQL class down the line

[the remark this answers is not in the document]
bouthilx: They will change the file handler?

src/metaopt/tests/mongodb_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+def mongodb_load_test(db, experiment, trial, field_filters):
+    result = db.load(experiment, trial, field_filters)
+    assert result is not None
+
+def mongodb_remove_test(db, experiment, trial, field_filters):
+    result = db.remove(experiment, trial, field_filters)
+    assert result is None
+
+def mongodb_tests():
```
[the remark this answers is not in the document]
dendisuhubdy: I thought we agreed on working with `pytest` that's why we moved to these functions, now we decided on Unittest.TestCase and now we're moving back to the previous class?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,140 @@
...
+        self._uri = uri
+        return self._uri
+
+    def initiate_connection(self):
+        try:
+            conn = MongoClient(self.uri)
+            self.conn = conn
+        except pymongo.errors.ConnectionFailure as e:
+            raise_from(DatabaseError("Could not connect to server"), e)
+            self.logger.debug("Could not connect to server")
```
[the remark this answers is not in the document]
dendisuhubdy: I don't understand why this would never be executed.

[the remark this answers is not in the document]
bouthilx: Lines under `raise SomeError()` never get executed. Once reraising the error, the process will move up to the code calling this method.

src/metaopt/tests/mongodb_test.py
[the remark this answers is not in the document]
bouthilx: We can use pytest without unittest if Christos and you prefers this. From what I understand Christos' comment was no against nosetest but simply that pytest supports the class tests without inheriting from special classes like unittest.TestCase. You can avoid writing a test function to combine them all by structuring the tests in a class.

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
dendisuhubdy: Ok so let's settle for `raise Exception` then?

src/metaopt/io/database/abstractdb.py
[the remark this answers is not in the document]
dendisuhubdy: Shouldn't one prefer just one type of database, so we just need one type of `database.log` file 
 rather than `mongodb-database.log` thus one `database` filehandler? What do you think? :)

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
tsirif: He means that commands should be in reverse order 😛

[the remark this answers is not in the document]
tsirif: First `logger.debug` and then `raise_from`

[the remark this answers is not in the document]
dendisuhubdy: Oh lol :dancer:

src/metaopt/io/database/abstractdb.py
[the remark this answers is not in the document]
bouthilx: Yes, but then I still don't see why set self.fh. It seems to me that they can all use self.logger just fine without messing with self.fh.

src/metaopt/io/database/abstractdb.py
```diff
@@ -32,19 +32,19 @@ def close_connection(self):
         pass
 
     @abstractmethod
-    def insert(self, experiment_name, trial_field, dictionary, field_filters):
+    def insert(self, collection_name, dictionary, field_filters):
```
bouthilx: I don't think field_filters is necessary here

src/metaopt/io/database/abstractdb.py
```diff
@@ -32,19 +32,19 @@ def close_connection(self):
...
         pass
 
     @abstractmethod
-    def update(self, experiment_name, trial_field, update_dictionary, field_filters):
+    def update(self, collection_name, update_dictionary, field_filters):
```
bouthilx: I think `(self, collection_name, query, update_dictionary)` would be more clear.

src/metaopt/io/database/abstractdb.py
```diff
@@ -32,19 +32,19 @@ def close_connection(self):
...
         pass
 
     @abstractmethod
-    def load(self, experiment_name, experiment_field, field_filters):
+    def load(self, collection_name, field_filters):
```
bouthilx: I think `(self, collection_name, query, projection)` or `(self, collection_name, query, selection)` would be more clear. `projection` is the name used by mongodb for what is called a _selection_ in SQL DBs. I don't have a clear preference. Sticking to mongodb's nomenclature could make the interface more coherent but at the same time I always found `projection` to be less obivous than _selection_.

src/metaopt/io/database/abstractdb.py
```diff
@@ -32,19 +32,19 @@ def close_connection(self):
...
         pass
 
     @abstractmethod
-    def remove(self, experiment_name, trial_field, field_filters):
+    def remove(self, collection_name, field_filters):
```
bouthilx: I think (self, collection_name, query) would be more clear.

src/metaopt/io/database/mongodb.py
```diff
@@ -108,33 +105,26 @@ def update(self, collection_name, update_dictionary, field_filters=None):
...
-        if field_filters is None:
-            field_filters = {}
+    def load(self, collection_name, query, selection):
 
         dbcollection = self.db[collection_name]
-        dbdocs = list(dbcollection.find(field_filters))
+        dbdocs = list(dbcollection.find(query, selection))
 
-        if len(dbdocs) == 0:
+        if len(dbdocs) is 0:
```
bouthilx: Why switch from == to is?

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,130 @@
...
+            unlock = self.conn.unlock()
+        except pymongo.errors.ConnectionFailure as e:
+            self.logger.debug("Could not determine unlock due to connection")
+            raise_from(DatabaseError("Could not determine unlock due to connection."), e)
+        return unlock
+
+    def insert(self, collection_name, query, dictionary):
+
+        dbcollection = self.db[collection_name]
+        result = dbcollection.insert(query, dictionary)
```
bouthilx: AFAIK there is no such thing as insert(query, dict) in pymongo. It should be insert_many(query, objects) where objects is a list of dictionaries. Also I can't see why we would want to query something when we do an insert. It's not an update.

src/metaopt/io/database/mongodb.py
dendisuhubdy: Technically `is` works on object comparison and `==` works on value comparison. I guess in this case it's a comparison between the lengths I was just trying to make the code to look more functional.

src/metaopt/io/database/mongodb.py
dendisuhubdy: Insert does exists [What's the difference between insert(), insertOne() and insertMany() methods on MongoDB](https://stackoverflow.com/a/36794912/3492631). I've double checked it with a local execution on my local computer with MongoDB too.

dendisuhubdy: Although you're right, it's deprecated, I'll do some changes, although `object` does sound ambiguous, I propose `dictionaries`?

src/metaopt/io/database/mongodb.py
bouthilx: I know the difference :) and I would agree if it was meant for `is False` but here it's really about integers.

src/metaopt/io/database/mongodb.py
bouthilx: Oh! I though it did not exist looking at the doc [here](http://api.mongodb.com/python/current/api/pymongo/collection.html?_ga=2.24207676.782214609.1511212041-2115054929.1508380031#pymongo.collection.Collection.insert_one). Still, insert does not take a query as first argument. I don't understand how your tests can pass :thinking: 

I agree objects is not very clear but dictionaries only tells about the container not the content. What about document_dictionaries?

tsirif:
How do you find this?

Public Interface
--------------
- [ ] **Read *experiment*** document **by** name **from** *db* **as** a dict: `exp_name` -> `exp_config`

1. Query for existing entry with experiment name, `exp_name`.
2. Find document and return it as a dictionary (perhaps project to all except 'trials' key).

- [ ] **Write new *experiment*** document **with** configuration [dict]: `exp_config` -> `experiment document handle?`

1. Insert entry with new **unique** experiment name, `exp_config['exp_name']`.
2. Entry must have all key-value pairs which are contained in exp_config + `'last_modified'` time attribute set initially to the time inserted in DB. 

- [ ] **Read data** **by** attribute name **from** *db*, *experiment* **as** "value"/"list of values": [`exp_name`, `attr_name`, `**filter`] -> `data`

1. Search in experiment, with name `exp_name`, for attribute, with name `attr_name` and return a "value" (number, string or dict) or a "list of values", corresponding to that attribute.
2. If attribute, with name `attr_name`, **is known to** correspond to a "list of values", then `**filter` **can** contain `key=(how, value)` or `key=value` entries:
   * `key` := attribute of an object in that list
   * `how` := way of comparing; one of `['=', '>=', '<=', '>', '<']` ?
   * `value` := reference "value"
   * A `key=value` entry assumes `how == '='`.

- [ ] **Write data** **by** attribute name **to** *db*, *experiment* **as** "value"/"list of values": [`exp_name`, `attr_name`, `data`, `**filter`] -> `data`

1. `len(filter) != 0` assumes an **update** at a "list of values" attribute, with name `attr_name`. `data` is a "value".
2. `len(filter) == 0` assumes:
    * If attribute, with `attr_name`, **is known to** correspond to a "list of values", then **insert (many)** `data` in that list.
    * Else, **(over)write** `data` of that attribute.

tsirif:
@dendisuhubdy To be able not to brake CI from now on, it would be better if, instead of merging master back, you would rebase on top of master (**after** #7 is merged ^_^)

bouthilx:
@tsirif I don't understand why we would need 2 read/write interfaces for the DB rather than a single one. For now I would only rely on MongoDB's methods for doing complex queries and updates. All the information for this can be contained in the dictionaries passed to the DB methods. We can later port those dictionaries to SQL statements.

tsirif:
Initially I thought that it would be convenient to have two different sets, one for experiment entry and one for an experiment's attributes. But I see your point, you are right. I am restating that.

tests/core/unittests/mongodb_test.py
```diff
@@ -0,0 +1,64 @@
...
+def moptdb():
+    """Return MongoDB wrapper instance initiated with test opts."""
+    moptdb = MongoDB(dbname='metaopt-test', username='user', password='pass')
+    return moptdb
+
+
+def test_bad_connection():
+    """Raise when connection cannot be achieved."""
+    with pytest.raises(DatabaseError) as exc_info:
+        MongoDB(dbhost='asdfada', username='uasdfaf', password='paasdfss')
```
bouthilx: How do you test this? You don't have a database right?

tests/core/unittests/mongodb_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+from pymongo import MongoClient
+import pytest
+import yaml
+
+from metaopt.io.database.abstractdb import DatabaseError
+from metaopt.io.database.mongodb import MongoDB
+
+
+@pytest.fixture(scope='module')
```
bouthilx: Wow, I didn't know about pytest fixtures, that's nice!

.pylintrc
```diff
@@ -51,7 +51,7 @@ confidence=
...
+disable=redefined-outer-name,fixme,F0401,intern-builtin,nonzero-method,parameter-unpacking,backtick,raw_input-builtin,dict-view-method,filter-builtin-not-iterating,long-builtin,unichr-builtin,input-builtin,unicode-builtin,file-builtin,map-builtin-not-iterating,delslice-method,apply-builtin,cmp-method,setslice-method,coerce-method,long-suffix,raising-string,import-star-module-level,buffer-builtin,reload-builtin,unpacking-in-except,print-statement,hex-method,old-octal-literal,metaclass-assignment,dict-iter-method,range-builtin-not-iterating,using-cmp-argument,indexing-exception,no-absolute-import,coerce-builtin,getslice-method,suppressed-message,execfile-builtin,round-builtin,useless-suppression,reduce-builtin,old-raise-syntax,zip-builtin-not-iterating,cmp-builtin,xrange-builtin,standarderror-builtin,old-division,oct-method,next-method-called,old-ne-operator,basestring-builtin
```
bouthilx: Why do we disable so much checks?

.travis.yml
```diff
@@ -29,7 +29,7 @@ install:
   - pip install tox
 before_script:
   - sleep 15
-  - mongo metaopt --eval 'db.createUser({user:"user",pwd:"pass",roles:["readWrite"]});'
+  - mongo metaopt_test --eval 'db.createUser({user:"user",pwd:"pass",roles:["readWrite"]});'
```
bouthilx: Ah! Here is the test database. Fine.

src/metaopt/io/database/mongodb.py
```diff
@@ -11,17 +11,9 @@
...
-    from urllib.parse import quote_plus
-except ImportError:
-    # Python 2.x
-    from urllib import quote_plus
-
 class MongoDB(AbstractDB):
-    def __init__(self, dbhost='localhost', dbname='metaopt',
-                dbport=27017, username=None, password=None):
-        super(MongoDB, self).__init__(dbhost, dbname, dbport, username, password)
+    def __init__(self, uri):
```
bouthilx: It seems my comment on slack wasn't clear. You did the exact opposite of what I proposed. :laughing: 

I believe the simplest functional interface for now would be `_init__(uri, with, all, the, sugar, already, there, being, optional)`, but inside `__init__` only `self._uri = uri` would be set. Then, the uri getter would be as you wrote. However, there would be no setter for now as all the sugar already there are not set in the database yet. 

I know it's tempting to already implement all of it. It's what, just 2 or 3 more functions and a couple of attributes after all? But then we need to test all use cases and make sure passing uri works fine just as much as passing attributes, and think about what to do if user pass uri + other arguments. How do we handle the setter for uri? How do we handle the setter for the other arguments? It's a lot more work than it looks like, enough to make us miss the Friday's milestone for the DB. Furthermore, we would have to maintain this code while implementing the rest of the minimum viable product, hence spending even more time maintaining extra features.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def unlock(self):
+        try:
+            unlock = self.conn.unlock()
+        except pymongo.errors.ConnectionFailure as e:
+            self.logger.debug("Could not determine unlock due to connection")
+            raise_from(DatabaseError("Could not determine unlock due to connection."), e)
+        return unlock
+
+    def insert(self, collection_name, query, dictionaries):
```
bouthilx: I believe previous discussion still holds: https://github.com/mila-udem/metaopt/pull/8#discussion_r152126959

tests/core/unittests/mongodb_test.py
dendisuhubdy: So for the wrong host we would assert that the connection will definitely fail.

src/metaopt/io/database/mongodb.py
dendisuhubdy: I don't get it really. If you mean we use the function parser, we need to pass a `uri` into it, before it was like this, it was `__init__(self, dbhost, dbname, dbport, username, password))` which you said we shouldn't do this and we should just use the parser instead. If that get's parsed again inside the `MongoClient` we should just pass in the `uri`, which it would look like just that. 

I don't know how this is incorrectly. You don't need to pass through init both `uri` and all of the `dbhost`, `dbname`, `dbpassword`, and `dbport` all over again, it's inefficient.

tests/core/unittests/mongodb_test.py
bouthilx: Yes I was confused because I didn't know we had a mongo database setup for the tests. I realized when looking at the CI setup files.

src/metaopt/io/database/mongodb.py
bouthilx: I see why you are confused, I'm sorry I didn't detail enough my thoughts. 

I believe we should eventually support both 

```python
db = Database(uri)
```

and 

```python
db = Database(dbhost, dbname, dbport, username, password, **options)
```

Just like pymongo's MongoClient does. But for now only one of the two would be enough for the MVP. I believe `Database(uri)` would be the simplest one to implement first. You can keep the arguments in the init as they are optional, but they wouldn't be used for now.

We could use the same mechanism than MongoClient:
> class pymongo.mongo_client.MongoClient(host='localhost', port=27017, ...)
> ...
> The host parameter can be a full mongodb URI, in addition to a simple hostname

src/metaopt/io/database/mongodb.py
```diff
@@ -86,10 +86,10 @@ def unlock(self):
             raise_from(DatabaseError("Could not determine unlock due to connection."), e)
         return unlock
 
-    def insert(self, collection_name, query, dictionaries):
+    def insert(self, collection_name, query, document_dictionaries):
```
bouthilx: What about the query? Based on the [doc](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.insert_many), it seems it should be `insert_many(document_dictionaries)`. I don't understand what a query would be used for in during an insertion.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,132 @@
...
+                    }
+        }
+        """
+
+        dbcollection = self.db[collection_name]
+
+        # see http://api.mongodb.com/python/current/api/pymongo/collection.html
+        # for more details
+
+        result = dbcollection.update_one(query, update_dictionary)
```
bouthilx: I guess we should support update_many here.

dendisuhubdy:
@tsirif @bouthilx are we going to implement https://github.com/mila-udem/metaopt/pull/8#issuecomment-345928728 or this thin wrapper is enough?

bouthilx:
@dendisuhubdy I believe the wrapper is enough to handle all of this. We can write _load, _update and _save methods in Experiment and Trial if we want to save us the burden of always specifying the collection_name and id. This way we have a clear separation between databases and meta optimization specific classes.

@tsirif What do you think?

src/metaopt/io/database/mongodb.py
```diff
@@ -50,6 +50,7 @@ def initiate_connection(self):
         try:
             conn = MongoClient(self.uri)
             self.conn = conn
+            self.conn.admin.command('ismaster')
```
bouthilx: That's a good example of why detailed commit messages are necessary. I believe that would be a good time to read and apply what I suggested [in this comment](https://github.com/mila-udem/metaopt/pull/8#issuecomment-344464658). In this case, the commit message should explain why we need to set ismaster and maybe also the alternatives and why command('ismaster') is better. 

Writing detailed commit messages does slow down the process of committing code but it significantly speeds up review and maintenance. Also, you will sometimes realize while writing the commit message that something is wrong or missing in the commit. Many of your commits in this PR have been undo/redo, which is a sign that what you were doing was not totally clear to you. It was my fault many times, because my comments were not detailed enough, but if you took the time to write down the commit message you could have realized that informations was missing and you would have asked me for clarifications before committing the changes.

src/metaopt/io/database/mongodb.py
```diff
@@ -79,43 +79,94 @@ def is_locked(self):
...
+
+        result = dbcollection.find_one_and_update(filter=query,
+                                         update=update_dictionary,
+                                         projection=selection,
+                                         upsert=upsert,
+                                         sort=sort,
+                                         return_document=return_document)
+        """
 
+        Finds a single document and updates it, returning either the original or the updated document.
```
bouthilx: Why a string here? It won't be registered as a docstring.

src/metaopt/io/database/mongodb.py
```diff
@@ -79,43 +79,94 @@ def is_locked(self):
...
+        although a find_and_modify is exactly that we would want
+        but the interface is deprecated so we use find_and_update() instread
 
-        result = dbcollection.update_one(query, update_dictionary)
+        """
+
+        result = dbcollection.find_one_and_update(filter=query,
+                                         update=update_dictionary,
+                                         projection=selection,
+                                         upsert=upsert,
```
bouthilx: `upsert` is a nice additon, but why limit ourselves to one update? Please don't make a new commit with update_many. :slightly_smiling_face:  I'd prefer we discuss it and we'll see if we settle on find_one_and_update or update_many.

@tsirif Any thoughts on this?

src/metaopt/io/database/mongodb.py
```diff
@@ -188,4 +188,8 @@ def remove(self, collection_name, query):
         Returns:
         result -- the document/documents
         """
-        self.db[collection_name].remove(query)
+        result = self.db[collection_name].find_one_and_delete(filter=query,
```
bouthilx: To complement this [comment](https://github.com/mila-udem/metaopt/pull/8#discussion_r152851453), I feel like deletion are very likely to be requested on many documents and we are getting our hands tied here.

src/metaopt/io/database/mongodb.py
dendisuhubdy: We recently encountered a problem by doing a nested query inside a dictionary, and by the way `update` is deprecated http://api.mongodb.com/python/current/api/pymongo/collection.html (at the bottom of the page)

src/metaopt/io/database/mongodb.py
dendisuhubdy: Idem. But I'd rather return the document first at the end of the function rather than using `delete_one` or `delete_many`.

src/metaopt/io/database/mongodb.py
bouthilx: update is deprecated but update_{one, many} are not.

Do you have an example for the problem? I'd like to understand it better. It's not clear to me that update_one rather than update_many would be the best fix for that kind of problem.

dendisuhubdy: Also you can not update two documents at once with a MongoDB query. You will always have to do that in two queries. You can of course set a value of a field to the same value, or increment with the same number, but you can not do two distinct updates in MongoDB with the same query. As a reference https://stackoverflow.com/questions/17207183/way-to-update-multiple-documets-with-different-values

src/metaopt/io/database/mongodb.py
bouthilx: Being able to recover the deleted document could be handy but I'm not sure we are likely to encounter such a use case.

dendisuhubdy: Yes, although rarely. I'm seeing a case where we want to simply remove a `node` from the set of `resources` if suddenly it becomes idle, we need to do a remove from the `resources`, but then if the `node` shows up again on our ping, we might need to return it back again. Just removing it might be dangerous.

bouthilx: I don't know. If ressources are represented as objects which can easily register themselves in the database then it would do no harm to remove it and register again using the objects info rather than the deleted rows info. To me the best point in favor of find_and_remove is if we want to delete a row and make sure nobody changes it meanwhile. If we do db.load() and then db.remove() it could have changed in between. But again, how is that likely to happen given our use cases and is it worth limiting the flexibility of our interface? 

@dendisuhubdy @tsirif What would you think about setting base methods (insert, load, update, remove) as non-atomic ones supporting many operations and latter on we extend the interface to find_and_(update, remove) if needed?

src/metaopt/io/database/mongodb.py
bouthilx: Yes sure. Different updates need different calls to update(). I don't see that as an argument against update_many. The only argument I see in favor of find_one_and_update is the atomicity of the operation. I'm not sure however it is worth the cost for flexibility.

src/metaopt/io/database/mongodb.py
```diff
@@ -88,26 +90,16 @@ def unlock(self):
...
-               selection, upsert=False, sort=None,
-               return_document=ReturnDocument.BEFORE):
-        """ Finds a document and modify it
+    def write(self, collection_name,
+                    query,
+                    dictionary,
+                    selection,
+                    atomic=False,
+                    upsert=False,
+                    sort=None,
```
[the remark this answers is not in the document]
bouthilx: Forget about last comment. Based on the discussion here #10 it seems atomic operations on many rows is something we should forget about.

Then I think the behaviors for atomic is True or False are too different to be crammed inside the write() method. Since the default behavior is *not* atomic then we could postpone the addition of atomic operation to write/remove and finish this PR with only `write(collection_name, query, dictionary, selection)` (and maybe `upsert` if we decide to make it definable). We can decide later if we integrate atomicity inside write() to keep only write/read/remove or if we add new explicit methods for atomic operations.

src/metaopt/io/database/mongodb.py
```diff
@@ -88,26 +90,16 @@ def unlock(self):
...
-    def update(self, collection_name, query, update_dictionary,
-               selection, upsert=False, sort=None,
-               return_document=ReturnDocument.BEFORE):
-        """ Finds a document and modify it
+    def write(self, collection_name,
+                    query,
+                    dictionary,
+                    selection,
+                    atomic=False,
+                    upsert=False,
```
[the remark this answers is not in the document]
dendisuhubdy: If we always set it to `True` we'd lose power over just updating or doing an insert.

src/metaopt/io/database/mongodb.py
```diff
@@ -88,26 +90,16 @@ def unlock(self):
...
+                    sort=None,
+                    return_document=ReturnDocument.BEFORE,
```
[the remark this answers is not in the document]
dendisuhubdy: I'm suspicious that for MySQL or PostgreSQL the interface might be different, I'll do research for now and get back to you for that.

src/metaopt/io/database/mongodb.py
```diff
@@ -88,26 +90,16 @@ def unlock(self):
...
+                    collation=None):
```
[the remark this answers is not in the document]
dendisuhubdy: Ah I see :+1:

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
bouthilx: No, upsert only inserts if the document is not found using its id. A very common use case will be to write new documents so it would be weird to always have to set database.write(new_documents, update=True). Also, in our case updates will usually be done by the objects representing the rows (Experiments, Trials, etc) so it will be simple to pass to `write()` the id so that `write()` does an update rather than an insert. All of that without specifying `upsert`.

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
bouthilx: The interface will probably be different but we could handle that inside the method. If we define our own codes for return_document we could use the same interface for all type of databases. It could be something like this inside the abstractdb python file.

```python
class ReturnDocument(object):
    NONE = 0
    BEFORE = 1
    AFTER = 2
```

For simplicity I would only implement NONE for now.

.pylintrc
tsirif: Most of the them are default exclusions, I will try to introduce small amount of them at a time.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+                raise_from(DatabaseError("Connection Failure: database not found on "
+                                         "specified uri"), e)
+            except pymongo.errors.OperationFailure as e:
+                self._logger.error("Could not verify user, %s, on database, %s",
+                                   self.username, self.dbname)
+                raise_from(DatabaseError("Authentication Failure: bad credentials"), e)
+
+            self._conn = conn
+            self._db = db
+            return conn
```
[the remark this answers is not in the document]
bouthilx: I'm in favor of the second too.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+           to contain a unique *_id* key.
+
+        .. note::
+           In the case of an update operation, if `query` fails to find a
+           document that matches, insert of `data` will be performed instead.
+        """
+        dbcollection = self._db[collection_name]
+
+        if query is None:
+            # We can assume that we do not want to update.
```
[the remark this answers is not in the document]
bouthilx: Thanks for the explanation. I agree with you, `insert_many` is better.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+                data = [data]
+            result = dbcollection.insert_many(documents=data)
+            return result.acknowledged
+
+        update_data = {'$set': data}
+        if atomic is False:
+            result = dbcollection.update_many(filter=query,
+                                              update=update_data,
+                                              upsert=True)
+            return result.acknowledged
```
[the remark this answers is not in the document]
bouthilx: ok good.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+        elif atomic is True:
+            selected_one_key, selection = _get_selected_one_key(selection)
```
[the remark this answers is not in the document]
bouthilx: Ok I understand the intent. Then we would not need to do `db.read({}, {'arg':1})['arg']`. But isn't a bit weird? I'd be surprised as a user if making a selection of a single value suddenly returns items rather than dictionaries. It's convenient, but it makes the method's behavior more complex. What if some function makes dynamic reads and at some point its dictionary gets unitary. It would need to take into account that at if its dictionary gets unitary than it needs to rebuild the dictionary because the method suddenly returns an item.

Personally, I would favor coherence and simplicity over a convenience like this. Also, accumulation of small tweaks like that adds on the burden of testing and maintenance, thus slowing down development. Writing `db.read({}, {'arg':1})['arg']` is less elegant but doesn't have such an impact.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+            result = dbcollection.find_one_and_update(filter=query,
+                                                      update=update_data,
+                                                      projection=selection,
+                                                      upsert=True,
+                                                      sort=[('$natural', -1)],
+                                                      return_document=ReturnDocument.AFTER)
+            return _get_leaf_value(selected_one_key, result) if selected_one_key else result
```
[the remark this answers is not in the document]
bouthilx: See comment [above](https://github.com/mila-udem/metaopt/pull/8/files/9c1d7e4d3ca4dca8588715e6211111b78428945b#r153378283)

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+        dbcollection = self._db[collection_name]
+
+        selected_one_key, selection = _get_selected_one_key(selection)
+
+        if atomic is False:
+            cursor = dbcollection.find(query, selection)
+            dbdocs = list(cursor)
+        elif atomic is True:
+            # Return the most recent of the query
+            dbdocs = list(dbcollection.find(query, selection, sort=[('$natural', -1)], limit=1))
```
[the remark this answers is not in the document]
bouthilx: In this case `atomic` means that the operation is safe for concurrency. The atomicity is about the operation. The operation could return many documents.

With the current implementation we are force to return a single document in `write()` because of MongoDB particularities, but in general `write()` could be applied on many documents _atomically_. We are forcing `read()`s behavior to stay coherent on `write()`s behavior, while the later was already forced because of a specific database particularity. That doesn't sound like a good design to me.

It is my belief that we should give ourselves some time to figure out what would be the best way to integrate atomicity. Not having it won't hurt us for the MVP.

tests/core/unittests/mongodb_test.py
```diff
@@ -0,0 +1,394 @@
...
+        loaded_config = moptdb.read('experiments',
+                                    {'exp_name': 'supernaedo2',
+                                     'metadata.user': 'tsirif',
+                                     'metadata.datetime': datetime(2017, 11, 22, 23, 0, 0)})
+        assert loaded_config == exp_config[0][0]
+
+    @pytest.mark.skip(reason="sugar")
+    def test_ctx_read_experiment(self, exp_config, moptdb):
+        """Fetch a whole experiment entry using an experiment context (always atomic)."""
+        with moptdb('experiments', {'exp_name': 'supernaedo2',
```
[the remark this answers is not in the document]
bouthilx: Yes I saw the sugar skip. :laughing: 

If I may ask you, could you please remove tests that are for sugar features. You can keep them in a branch on your fork to avoid loosing the code. I would also encourage you to make an issue where you explain your idea of context manager if you do spend some time on that. Nevertheless, we are past the milestone for the DB and are gonna soon be past the milestone for Experiment/Trials so I would recommend we focus on the base first.

src/metaopt/io/database/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    port : int
+       Port that database server listens to for requests.
+    username : str
+        Name of user with write/read permissions to database with name `dbname`.
+    password : str
+        Secret phrase of user, `username`.
+
+    """
+
+    def __init__(self, uri=None, dbname='metaopt', host='localhost',
```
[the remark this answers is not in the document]
bouthilx: I love it. Defining the attributes separately in the configuration file is more natural than the URI. The URI is more convenient for passing as an argument in command line.

So as you propose I would make the `host`/`uri` the first and only positional argument and all the arguments would be passed to MongoClient rather than the URI. That would also address Pascal's concern about the username/password in the URI. In that case, I would drop everything related to URI; the property and self._uri. Until we add support for URI, MongoDB("some_uri") would fail.

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+            fixed_pass = quote_plus(self.password)
+            fixed_dbname = quote_plus(self.dbname)
+            uri = 'mongodb://{0}{1}{2}{3}'.format(
+                fixed_user + ':' + fixed_pass + '@' if self.username else '',
+                self.host, ':' + str(self.port) if self.port else '',
+                '/' + fixed_dbname if self.dbname else '')
+        else:
+            uri = self._uri
+        ret = parse_uri(uri)
+        if self._uri is not None:
```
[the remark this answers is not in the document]
bouthilx: For reference, comments are [here](https://github.com/mila-udem/metaopt/pull/8/files/9c1d7e4d3ca4dca8588715e6211111b78428945b#r153276922)

src/metaopt/io/database/mongodb.py
```diff
@@ -0,0 +1,277 @@
...
+        ---------
+        collection_name : str
+           A collection inside database, a table.
+        data : dict or list of dicts
+           New data that will **be inserted** or that will **update** entries.
+        query : dict, optional
+           Assumes an update operation: filter entries in collection to be updated.
+        atomic : bool, optional
+           If True, update only the most recent entry that matches `query`.
+        selection : dict, optional
```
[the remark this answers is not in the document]
bouthilx: I do not like to have documents returned from times to times. :stuck_out_tongue: I think the simplest for now would be to never return documents, only (True, False) to assess the operation's success.

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
tsirif: With more dotted selections 'arg1.arg2.arg3' the more ugly it will get. But I agree about your point on coherence and simplicity. Will see what I can do 😉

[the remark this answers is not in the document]
bouthilx: We should finish this [discussion](https://github.com/mila-udem/metaopt/pull/8#discussion_r153380279) before spending time on `selected_one_key`, otherwise it might be a waste of time.

tsirif:
@dendisuhubdy @bouthilx I think this is done. What do you think? Dendi could you merge the last commits I made?

dendisuhubdy:
@tsirif I believe it is done, @bouthilx could you approve the PR? So @tsirif could merge it :)

bouthilx:
Overall yes but we didn't finish this [discussion](https://github.com/mila-udem/metaopt/pull/8#discussion_r153380279). When we are done discussing, I'll make a final review and we should be able to merge.

src/metaopt/io/database/mongodb.py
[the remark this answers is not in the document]
tsirif: I see your point. I think though that I will find useful searching for a single most recent document later on, for MVP. We could change the name of the optional argument though to something else. I believe that in general this is ok for now. `AbstractDB` is an internal API, we could change this later.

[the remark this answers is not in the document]
bouthilx: I understand having a way to fetch the most recent document can be handy, but I disagree this is how it should be done. If I read the following code snippet, it's not clear why the output would be the most recent row unless I read the doc.
```python
most_recent_row = db.read(some_query, atomic=True)
```

The following is less elegant but more explicit
```python
most_recent_row = next(sorted(db.read(some_query), key=lambda d: d["date"]))
```

And could later be turned into
```python
most_recent_row = db.read(some_query, sort={"date": 1})
```

Anyhow, we don't need sort for now, and adding it would not impact the interface so we don't need to rush for it. However, adding it now would affect our progress. When we need it, we can simply make a PR for sort only. 

It might feel like I'm arguing for details and slowing down our progress as I advocate for removing stuff rather than changing or adding, but we need to focus more. The tendency of adding extra features will hurt us more and more as we progress with a larger and more complicated code base. Every extra feature is adding burden on development and maintenance. When the code is designed properly, extra features are easy to integrate later on, so there is no reason to rush them in. What we should rather rush for is users' feedback. We cannot afford to design hastily for that, but we can spare time on extra features.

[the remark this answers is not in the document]
tsirif: I would just like an optional interface to return the most return recent thing. Databases can do that generally. I would not like a general `sort` argument, this can be done as you demonstrated. However, searching for the most recent something is probably going to be common. I am not sure.

I agree with your last comment. It takes unjustified motivation to remove stuff I tried right now to facilitate for my future coding. However, I will try to contain my urge to implement in front stuff before they become necessary, in the future.

What would you suggest that I should do? Should I remove it? I think the current tests describe pretty much what I will need to proceed. Interface can be polished later, no need to perfect it right now as you said.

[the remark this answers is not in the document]
bouthilx: > It takes unjustified motivation to remove stuff I tried right now to facilitate for my future coding. However, I will try to contain my urge to implement in front stuff before they become necessary, in the future.

I don't understand what you mean.

> What would you suggest that I should do? Should I remove it? I think the current tests describe pretty much what I will need to proceed. Interface can be polished later, no need to perfect it right now as you said.

I think the best would be to move the extra feature to a new branch and point to it in an issue. This way we do not loose trace of the code but do not have the burden of maintaining it. From what I see there should be 2 issues. One about atomicity and one about getting the most recent document only. Those two should not be coupled.

bouthilx:
@dendisuhubdy Why did you dismiss the review? In the main comment of the review I wrote "I would like this review to be the *penultimate* one". Please do not dismiss reviews. I will never dismiss your reviews either. This mechanism is there for a reason.

PR #17, opened 2017-11-30
Revert "MongoDB - class and tests"

bouthilx:
Reverts mila-udem/metaopt#8


bouthilx:
There is minor things I'll look at again when I have time. It does not deserve a revert and could be fixed with additional PRs.

dendisuhubdy:
@bouthilx I really apologize for dismissing your review. But I did it to speed up things that I believe might already have been addressed, you are surely free to dismiss my review too the next time. My main concern is that our pace is a little bit too slow looking at the deadline that you've set, and even though we (well mainly @tsirif) has implemented code which includes `atomicity` or `sort` that could be argued as `additional features` I feel it's even better to have them at start, even though it might be better if we discuss it ahead of time. His code does work and passes the tests. We could also discuss it later at our meeting.

bouthilx:
@dendisuhubdy 

The PR does not need to be merged to start a new branch for DB configuration or Experiment/Trial based on branch database. So merging it does not speed up things.

Adding additional features at start is the best way to risk wasting our time. Going too fast but in the wrong direction is worst then progressing slowly but in the good direction. The reasons I slowed us down until now is for planning and to make sure we take time to think thoroughly about the design. Another reason we have been slowed down until now is because of significant time invested on implementing features that could be added latter on, or even never be needed. Furthermore, adding many features at the same time makes it more difficult to decouple things and is risky to induce bad design decisions. What just happened with the last PR is a good example of that. The behavior of `db.write()` and `db.read()` methods became surprisingly complex as we tried to integrate many features at once even though those features could be introduced without affecting the behavior that much.

Convenient features might help us implement a bit faster given there is no errors, but bad design decisions can require tremendous efforts to be fixed. Adding features and discussing them later is a recipe for disaster. 

Imagine we implemented a DB with `write()` and `read()` only using a single type of methods (be it update_many, or find_one_and_update). When do you think we would have finished the PR? My bet is last Friday. So we would already have finished the DB configuration and would be now discussing the details of Experiment/Trials. The DB would be simple, yes, but have a clear behavior that we need to maintain and would be enough for or needs now. What do think is slowing us down?

Good design helps adding easily features latter on, but features does not help doing better design. So, given we have a limited time to work on this project, how should we make the best use of it? Make design hastily so we can add more features, or focus on a few primordial features so we can take time to discuss the design?

I will never dismiss your review unless you are out and cannot give feedback for weeks. If someone request a change, I cannot asses that my modifications will exactly address what was asked for because I cannot be 100% sure I understand perfectly what was meant. And even if the modification was correct, it sometime sheds light on another flaw that wasn't apparent before.

tl;dr Slow but steady wins the race.

PR #18, opened 2017-11-30
Feature: Implement database creator

tsirif:


src/metaopt/io/database/__init__.py
```diff
@@ -148,4 +148,70 @@ class DatabaseError(RuntimeError):
...
+    ----------
+    all : list of `AbstractDB` classes
+       Updated to contain `all` possible implementations currently. Check out code.
+    typenames : list of str
+       Names of implemented wrapper classes, correspond to possible ``dbtype``
+       values.
+
+    """
+
+    all = [AbstractDB, MongoDB]
```
bouthilx: Can't we make AbstractDB register automatically any children? In that case `all` could be replaced by something like `AbstractDB.types`. Lists like `all` are annoying to maintain. Test at line 195 would not be needed anymore.

src/metaopt/resolve_config.py
```diff
@@ -231,25 +231,34 @@ def merge_env_vars(config):
...
-       * **optimizer**: {optimizer module name : method-specific configuration}
-       * **dynamic**: {dynamic module name : method-specific configuration}
-
-       .. seealso:: Method-specific configurations reside in `/config`
+    Mopt Configuration
+    ------------------
+
+    name --  Experiment's name.
+
+       If you provide a past experiment**s name,
```
bouthilx: Why experiment**s?

src/metaopt/io/database/__init__.py
```diff
@@ -148,4 +148,70 @@ class DatabaseError(RuntimeError):
...
+    typenames = list(map(lambda x: x.__name__, all))
+
+    def __call__(cls, dbtype='AbstractDB', *args, **kwargs):
```
bouthilx: Shouldn't it be `dbtype=None`? I don't think we want AbstractDB to be ever instantiated.

tests/core/unittests/database_test.py
```diff
@@ -0,0 +1,37 @@
...
+    assert 'AbstractDB' in str(exc_info.value)
+
+
+def test_instatiation_and_singleton():
+    """Test create just one object, that object persists between calls."""
+    database = Database(dbtype='MongoDB', dbname='metaopt_test',
+                        username='user', password='pass')
+
+    assert Database() is database
+    assert Database('fire', [], {'doesnt_matter': 'it\'s singleton'}) is database
```
bouthilx: That seems error-prone. In my opinion we should allow `Database()` after first call to ease reuse of the database without knowing the details (type), but trying to get something different should raise an error. 

``` python
>>> database = Database("mongodb")
>>> database = Database("mysql")
>>> print type(database)
"MongoDB"
```
:thinking:

tests/core/unittests/mongodb_test.py
```diff
@@ -60,12 +60,14 @@ def test_bad_connection(self):
             MongoDB(host='asdfada', port=123, dbname='metaopt',
                     username='uasdfaf', password='paasdfss')
         assert "Connection" in str(exc_info.value)
+        MongoDB.instance = None  # Set singular instance to None for independent tests
```
bouthilx: Isn't there something similar to TestCast.tearDown() in pytest?

src/metaopt/io/database/__init__.py
tsirif: It cannot be, eitherways 😛

src/metaopt/resolve_config.py
tsirif: typo.

tests/core/unittests/database_test.py
tsirif: Good point. This is easy to implement.

tests/core/unittests/mongodb_test.py
tsirif: There is, I fix this in the next PR. About experiment and trials.

src/metaopt/io/database/__init__.py
tsirif: I could make something dirty here:
1. globbing for all Python files in database directory
2. try to import them
3. for each public name in imported module, check whether it is a subclasses of `AbstractDB`,
append to list of `types` if it is.

bouthilx: Yeah that's too dirty. :stuck_out_tongue:

Man, I just learned something so nice. :open_mouth: Did you know about ```SomeClass.__subclasses__()```? I just found that from [here](https://stackoverflow.com/a/5883218)

```Python
>>> class A(object):
...     pass
... 
>>> class B(A):
...     pass
... 
>>> A.__subclasses__()
[<class '__main__.B'>]
```

Magic! :tada:

src/metaopt/io/database/__init__.py
bouthilx: So why AbstractDB by default if it can't be? None seems less confusing to me, but I might miss something.

tests/core/unittests/mongodb_test.py
bouthilx: Why not fixing it here?

src/metaopt/io/database/__init__.py
tsirif: `None` raises an exception somewhere 'random'. In particular when doing `dbtype.lower()`. Passing 'AbstractDB' raises a more natural and informative error.

tests/core/unittests/mongodb_test.py
tsirif: I am grouping "session" setup/teardowns in a default file `conftest.py` by pytest. I did that in another branch.

tsirif:
@bouthilx @dendisuhubdy Call for final review (not needed immediately now, but I believe that it is finalized)

src/metaopt/io/database/__init__.py
```diff
@@ -20,7 +20,7 @@
 
 import six
 
-from metaopt.utils import AbstractSingletonType
+from metaopt.utils import (AbstractSingletonType, SingletonFactory)
```
dendisuhubdy: What is `SingletonFactory`?

src/metaopt/io/database/__init__.py
```diff
@@ -106,14 +106,14 @@ def write(self, collection_name, data,
         pass
 
     @abstractmethod
-    def read(self, collection_name, query, selection=None):
+    def read(self, collection_name, query=None, selection=None):
```
dendisuhubdy: So a call on read using an empty `query` means that we retrieve all of the documents inside? [but yeah we did agree on that]

src/metaopt/io/database/__init__.py
```diff
@@ -148,4 +148,11 @@ class DatabaseError(RuntimeError):
     pass
 
 
-from metaopt.io.database.mongodb import MongoDB  # noqa
+@six.add_metaclass(SingletonFactory)  # pylint: disable=too-few-public-methods,abstract-method
```
dendisuhubdy: Gotcha here

src/metaopt/resolve_config.py
```diff
@@ -72,9 +72,9 @@ def nesteddict():
 # list containing tuples of
 # (environmental variable names, configuration keys, default values)
 ENV_VARS_DB = [
-    ('METAOPT_DB_NAME', 'name', 'MetaOpt'),
-    ('METAOPT_DB_TYPE', 'type', 'MongoDB'),
-    ('METAOPT_DB_ADDRESS', 'address', socket.gethostbyname(socket.gethostname()))
+    ('METAOPT_DB_NAME', 'dbname', 'MetaOpt'),
```
dendisuhubdy: Yes these variable names are much more clear

src/metaopt/resolve_config.py
```diff
@@ -231,25 +231,32 @@ def merge_env_vars(config):
...
+    pool_size -- Number of workers evaluating in parallel asychronously
+
+       (default: 10 @ default resource). Can be a dict of the form:
+       {resource_alias: subpool_size}
+
+    database -- dict with keys: of_type, dbname, host, port, username, password
+
+    resources -- {resource_alias: (entry_address, scheduler, scheduler_ops)} (optional)
+
+    algorithm -- {optimizer module name : method-specific configuration}
```
dendisuhubdy: Neat :+1:

tests/core/unittests/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+
+import os
+
+import pytest
+import yaml
+
+from metaopt.io.database import Database
+from metaopt.io.database.mongodb import MongoDB
+
+TEST_DIR = os.path.dirname(os.path.abspath(__file__))
```
dendisuhubdy: Just to double check where does this point to?

tests/core/unittests/database_test.py
```diff
@@ -0,0 +1,48 @@
...
+    of a database by-itself (this on which every `Database` acts on as part
+    of its being, attributes of an `AbstractDB`) and for-itself (what essentially
+    differentiates one concrete `Database` from one other).
+
+    """
+
+    def test_empty_first_call(self):
+        """Should not be able to make first call without any arguments.
+
+        Hegelian Ontology Primer
```
dendisuhubdy: `Hegelian Ontology Primer` :)

tests/core/unittests/database_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_notfound_type_first_call(self):
+        """Raise when supplying not implemented wrapper name."""
+        with pytest.raises(NotImplementedError) as exc_info:
+            Database('notfound')
+        assert 'AbstractDB' in str(exc_info.value)
+
+    def test_instatiation_and_singleton(self):
+        """Test create just one object, that object persists between calls."""
+        database = Database(of_type='MongoDB', dbname='metaopt_test',
```
dendisuhubdy: cute we have an `of_type` now, so we could switch to `PostgreSQL` I assume

tests/core/unittests/mongodb_test.py
```diff
@@ -47,10 +36,12 @@ def clean_db(database, exp_config):
 @pytest.fixture(scope='module')
 def moptdb():
     """Return MongoDB wrapper instance initiated with test opts."""
+    MongoDB.instance = None
     moptdb = MongoDB(username='user', password='pass', dbname='metaopt_test')
     return moptdb
 
 
+@pytest.mark.usefixtures("null_db_instances")
```
dendisuhubdy: last question, what is `null_db_instances`?

bouthilx:
@tsirif I will look at it Monday morning.

src/metaopt/io/database/__init__.py
tsirif: `Factory` is a metaclass to create the `Database` class, a factory pattern aggregating within its creation both the argument denoting its implementation database type (see `of_type` argument), and the arguments to be passed for the creation/initialization of an implementation of the interface, defined by the base class of the Factory class to-be-created (in our case is the `AbstractDB`). See also the tests for how it is used.

`SingletonFactory` is the `Factory` metaclass wrapped by the `Singleton` metaclass, so that it is allowed to return only one instantiation of the created class.

tests/core/unittests/conftest.py
tsirif: This is the directory where *experiment.yaml* can be found, so that it can be loaded and used for the tests.

tests/core/unittests/database_test.py
tsirif: Yes. Basically if everything is alright then we have a wrapper officially.

tests/core/unittests/mongodb_test.py
tsirif: It is a setup fixture that resets `Singleton`-type classes internal singleton instance object to None, so that we can test in the same run every possible invocation of a Singleton's creation. We assume independency between two different test functions.

src/metaopt/io/database/__init__.py
tsirif: Oh damn I am not sure why I did that... see bad commit names. 😛 I am removing/stashing the commit tsirif@4570cd6 until it becomes necessary. @bouthilx agree?

bouthilx: I don't understand what is the problem here. The behavior ```db.read("collection_name")``` being similar to ```db.read("collection_name", {})``` seems fine the me.

src/metaopt/utils/__init__.py
```diff
@@ -19,16 +21,84 @@ def __init__(cls, name, bases, dictionary):
...
+
+    """
+
+    def __init__(cls, names, bases, dictionary):
+        """Search in directory for attribute names subclassing `bases[0]`"""
+        super(Factory, cls).__init__(names, bases, dictionary)
+
+        modules = []
+        base = import_module(cls.__base__.__module__)
+        py_files = map(lambda x: '.' + x.split('.')[0].split('/')[-1],
```
bouthilx: That will be a painful to test. Why not depending on imports for making available the subclasses? It makes sense to import all classes of a given type in the same ```__init__.py``` file, ```database/__init__.py``` for instance. Also, it makes it easier to drop subclasses if they have unmet dependencies for instance. That way Database(of_type="PostgreSQL") would not be possible if the dependencies are unmet (it's not imported). Maybe dropping them for that is not a good idea but it's an example of how more flexible it is. The downside is making obligatory imports in specific ```__init__.py``` but that is a common practice anyway.

tests/core/unittests/mongodb_test.py
```diff
@@ -47,10 +36,12 @@ def clean_db(database, exp_config):
 @pytest.fixture(scope='module')
 def moptdb():
     """Return MongoDB wrapper instance initiated with test opts."""
+    MongoDB.instance = None
```
bouthilx: Is it needed here because ```moptdb()``` is not a test so you cannot use ```null_db_instances``` fixture?

tsirif: The problem is that `null_db_instances` has `'function'` scope, while `moptdb` has `'module'` scope. If there's a need to create a `'module'` scope one I will do it, but currently there's no other needs for it. (It's kinda annoying, I know..)

src/metaopt/utils/__init__.py
tsirif: I did it like because if it's pain to append a class name to a `types` list, then I assume it would be also a pain to require importing every new class to `__init__.py`. This works. You can checks the test code. It assumes that all implementations lie in files that reside in the same directory as the base implementation. What else should be the case, in order to be considered that it is tested properly?

(Not too much) Later this will be augmented with searching through the `pkg_resources` for registered `entry_points` also, in order to include the case where the interface is exported as a plugin (I consider the database interface to be an internal though).

tests/core/unittests/mongodb_test.py
bouthilx: ok. Thanks for the explanation.

src/metaopt/utils/__init__.py
bouthilx: I was worried that we would try looking all around so there could be many corner cases to test, but what you explain here is simple. Let's try it that way.

bouthilx:
@tsirif Alright! Needs rebase on master and ready to merge.

tsirif:
@bouthilx Done!

dendisuhubdy:
Is this ok to merge?

tsirif:
@dendisuhubdy Yes!

PR #27, opened 2017-12-03
Close #24: Add MongoDB tests for using '_id' key

### LINKED ISSUE #24: Database: `AbstractDB` subclasses should speak to the same id-denoted key.
I am referring to this issue raised on meeting 2017/11/30 about the following concern:

MongoDB queries and returned objects have a unique identifier set in them by the MongoDB itself, without explicitly programming so. We find using a particular ID of this kind useful and thus we should make sure that this special key is framework-agnostic. Current MongoDB implementation has string `'_id'` as the key that denotes this unique document ID per collection.

I propose fixing this key `_id` as the standard key to be expected in our interface, and all of the other implementation would convert [to] their own, back & forth to that particular key `'_id'`.

That also, raises the issue of how to test this interface agreement. I think that issue #23 addresses this. There should be a test for each public interface {`read`, `write`, `remove`}:
   1. checking the agreed ID key for particular returned documents and,
   2. checking return documents for query using the agreed ID key.

tsirif:


tests/core/unittests/mongodb_test.py
```diff
@@ -182,15 +188,36 @@ def test_update_many_default(self, database, moptdb):
...
+        value = list(database.experiments.find(filt))
+        assert value[0]['pool_size'] == 36
+
 
 @pytest.mark.usefixtures("clean_db")
 class TestRemove(object):
     """Calls to :meth:`metaopt.io.database.MongoDB.remove`."""
 
-    def test_remove_many_default(self, database, moptdb):
+    def test_remove_many_default(self, exp_config, database, moptdb):
```
dendisuhubdy: Sounds good

tests/core/unittests/mongodb_test.py
```diff
@@ -182,15 +188,36 @@ def test_update_many_default(self, database, moptdb):
...
         assert database.experiments.count() == count_before - 2
+        assert database.experiments.count() == 1
+        assert list(database.experiments.find()) == [exp_config[0][2]]
+
+    def test_remove_with_id(self, exp_config, database, moptdb):
+        """Query using ``_id`` key."""
+        filt = {'_id': exp_config[0][0]['_id']}
+        count_before = database.experiments.count()
+        # call interface
+        assert moptdb.remove('experiments', filt) is True
```
dendisuhubdy: Sounds good

tests/core/unittests/mongodb_test.py
```diff
@@ -111,6 +111,12 @@ def test_read_experiment(self, exp_config, moptdb):
                                      'metadata.user': 'tsirif',
                                      'metadata.datetime': datetime(2017, 11, 22, 23, 0, 0)})
         assert loaded_config == exp_config[0][0]
+        assert loaded_config['_id'] == exp_config[0][0]['_id']
+
+    def test_read_with_id(self, exp_config, moptdb):
+        """Query using ``_id`` key."""
+        loaded_config = moptdb.read('experiments', {'_id': exp_config[0][2]['_id']})
```
dendisuhubdy: Sounds good. Same thing as we discussed before last week.

bouthilx:
@tsirif Is there any hard dependency on this PR for the other ones? I would review that one on Tuesday if possible so that I have time to review the other priority PRs. The PR is simple but I want to spend some time thinking and I'm too much in a hurry right now.

tsirif:
Nop there are not. Take your time.

tests/core/unittests/mongodb_test.py
```diff
@@ -182,15 +188,36 @@ def test_update_many_default(self, database, moptdb):
...
 
+    def test_update_with_id(self, exp_config, database, moptdb):
+        """Query using ``_id`` key."""
+        filt = {'_id': exp_config[0][1]['_id']}
+        count_before = database.experiments.count()
+        # call interface
+        assert moptdb.write('experiments', {'pool_size': 36}, filt) is True
+        assert database.experiments.count() == count_before
+        value = list(database.experiments.find(filt))
+        assert value[0]['pool_size'] == 36
```
bouthilx: You should check that other experiments' pool_size value didn't change.

tests/core/unittests/mongodb_test.py
```diff
@@ -182,15 +188,36 @@ def test_update_many_default(self, database, moptdb):
         assert value[1]['pool_size'] == 16
         assert value[2]['pool_size'] == 2
 
+    def test_update_with_id(self, exp_config, database, moptdb):
```
bouthilx: You should also test when using a fake ```_id``` that is not present in db. MongoDB should create a new row with the given id. That's different than a bare insert.

tsirif:
@bouthilx if travis is ok, I think this is done.

bouthilx:
@tsirif 1 failure :scream:

PR #35, opened 2017-12-13
Change packaging structure of MetaOpt package

tsirif:
This needs some explanation in order to be justified:

I am imagining that implementations of algorithms should not be in the same package as the core functionality. This is because:

1. It is possible that there will be many implementations of the same algorithm (e.g. bayesian from skopt and bayesian from somewhere else).
2. It will be difficult to track the development of many different implementation in the same distribution/repo.
3. User has to install explicitly an implementation to make it available, as it is considered an extension/plug-in. This has the advantage that user will know whether or not his installation supports an algorithm or not, and this explicit knowledge will help one configuring metaopt.

In addition, a meta-distribution could be released as the minimum recommended working setup of metaopt which will have hard dependencies on whatever we dim as the best hyperparameter solution working out of the box.

Proposed Package Structure
---------------------------------------
* **metaopt** (meta-package/namespace package): See [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#namespace-packages)
  * **core**: existing code base delivering production, consumption, database, hyperparameter space definition and embedding, configuration, and scheduling/job dispatching
  * **algo**: abstract algorithm definition, helper functions to create implementation, implementations of abstract algorithm, implementations of optimizing strategies
    * **base**: helper functions and abstract algorithm
    * **strategy**: impl of abstract algorithm that use other implementations, like global-to-local-search, or randomly-select-algorithm
    * <whatever implementation of `base.AbstractAlgorithm`>
  * **client**: interface to user's script, optional helper Python functions that facilitate communication with a worker

Proposed Distributions
-------------------------------
* **metaopt-core** will bring `metaopt.core` and `metaopt.algo.base`
* **metaopt-algo-strategy**; depends on **metaopt-core**
* **metaopt-algo-\<whatever implementation\>**; depends on **metaopt-core**
* **metaopt-client** will bring code to be used in a Python script; depends on **metaopt-core**
* **metaopt-minimal** will come bundled with default configuration file for the algorithm[s] (which will be installed in *share/metaopt/\<version\>/* subdirectory of the installation directory); depends on **metaopt-core**, **metaopt-algo-\<whatever-we-believe-it-is-best\>**


bouthilx:
@tsirif We can discuss this briefly at the meeting. Please fix the test failures. :slightly_smiling_face:

tsirif:
Ready 😄

PR #47, opened 2018-03-05
[WIP] Feature: Transformation of Dimensions

dendisuhubdy:
- [ ] Create transform class
- [ ] Create bijection transformations of
   - [ ] Real to Categorical
   - [ ] Categorical to Real
   - [ ] Integer to Categorical
   - [ ] Categorical to Integer
   - [ ] Real to integer
   - [ ] Integer to Real


bouthilx:
Do we need transformations from Real/Int to categorical? My understanding is that some algorithms do not work with Int/Categories so we need to transform to Real/Vectors of real.

Issue #42, opened 2018-02-26
Make trial selections atomic


bouthilx:
When a trial is selected, it will completely overwrite the database row when setting status to reserved. However, there could be a race condition meanwhile and another process could have written the row already. To avoid this, the query in write() should ensure that the status is the one we get from the first read(). If the query fails, the method needs to retry another reservation. We can assume that trials' db row cannot change if status is not 'reserved', so such a query would be sufficient to ensure atomicity (if underlying's database code is something like find_and_update).

```python
aknowledged = self._db.write('trials', selected_trial.to_dict(),
                             query={'_id': selected_trial.id,
                                    'status': selected_trial.status})
if not aknowledged:
    return self.reserve_trial(score_handle)

return selected_trial
```

nouiz:
This is an important problem. Hopefully it can be fixed soon.

Issue #14, opened 2017-11-30
Add `atomic` option in AbstractDB.{write,remove}


tsirif:
Relative [discussion](https://github.com/mila-udem/metaopt/pull/8#discussion_r153380279).

We would like to support **fetching multiple matched** documents on a write (meant as an update) or remove, **in an atomic way**. Preferably using a single command, if the database framework supports it. Else, probably a locking mechanism should be devised and used (see #10).

A first approach,**currently returning only one** document and applying the update or remove on that, lies at tsirif@ab159cf, on branch [`feature/db/atomic`](https://github.com/tsirif/metaopt/tree/feature/db/atomic)

Issue #10, opened 2017-11-24
Add global locks to databases


bouthilx:
Some procedures might require atomic operations on many rows. For that we would need a lock on the database. 

Regarding MongoDB, we can read in the [doc](https://docs.mongodb.com/manual/faq/concurrency/) that it supports many level of locks. 
> MongoDB uses multi-granularity locking [1] that allows operations to lock at the global, database or collection level, and allows for individual storage engines to implement their own concurrency control below the collection level

I don't know if global would be enough.

Here is a code snippet for mongodb from @dendisuhubdy 

```python
def lock(self):
    """Lock a database"""
    try:
        lock = self.conn.fsync(lock=True)
    except pymongo.errros.ConnectionFailure as e:
        self.logger.debug("Could not lock a database")
        raise_from(DatabaseError("Could not determine lock due to connection."), e)
    return lock

def unlock(self):
    """ Unlocking a database """
    try:
        unlock = self.conn.unlock()
    except pymongo.errors.ConnectionFailure as e:
        self.logger.debug("Could not determine unlock due to connection")
        raise_from(DatabaseError("Could not determine unlock due to connection."), e)
    return unlock
```

tsirif:
I think `fsync` should better be in the database. We read somewhere that it is intended for reserving time for maintenance/backup, could it be a slow operation?

dendisuhubdy:
@tsirif is referring to https://stackoverflow.com/a/11531818/3492631

bouthilx:
Alright so that should not be how we lock the database. I haven't read the doc yet. We should read it when we try to address this issue.

dendisuhubdy:
@bouthilx there's no other way to manually lock the database, although according to the MongoDB documentation, these operations cause `locking` https://docs.mongodb.com/manual/faq/concurrency/#which-operations-lock-the-database

bouthilx:
I see. There is no way of locking over many updates, unless we use fsync or a retry loop like explained [here](https://stackoverflow.com/a/11496809). It seems it is just better to forget about atomic operations on many documents at a time.

dendisuhubdy:
@bouthilx in that case, would it mean that the `if` block not be needed and also operations `updateMany` too? Or this interface would be just fine? Mind double checking again for code errors/logic so that both of you can approve it for merging by tomorrow? :)

PR #53, opened 2018-03-15
 Fix race condition for experiment trial selection

bouthilx:
Related to #10, #14 and #42
Needs #36 to be merged.

1. Add `read_and_write()` to `Database`
2. Use `read_and_write()` instead of `write()` to reserve trials inside `Experiment.reserve_trial()`

Why:

Some operations require atomicity. For instance, when selecting a new trial, between the first read to database, the potential scoring and the random selection, another process could reserve it concurrently. To avoid such concurrency problems, we must ensure at the update time that the status was not altered during the selection.

How:

Using `read_and_write()` can ensure atomic operations. At update time, the process only proceed to update if the selected document remained unaltered. In our specific case, we could look at the status, since by design  only "reserved" trials can possibly change. If there was a race condition the "looser" process simply rollbacks and retry its operation. This solution is in line with 'mongodb''s spirit regarding concurrency.

Note:

My decision to add `read_and_write()` instead of adding an `atomic` argument to `write()` was to avoid increasing the complexity of `write()` for a simple use case. `write()` supports operations on many documents, MongoDB atomic operations do not. Furthermore, it seems our requirements for atomic operation are fairly rare, occurring only at registration time for experiments and reservation time for trials. In both cases, those operations only affects one document.

### DIFF
--- a/tests/unittests/core/test_experiment.py
+++ b/tests/unittests/core/test_experiment.py
@@ -48,6 +48,71 @@ def patch_sample2(monkeypatch):
 
 
 @pytest.fixture()
+def patch_sample_concurrent(monkeypatch, create_db_instance, exp_config):
+    """Patch ``random.sample`` to return the first one and check call."""
+    def mock_sample(a_list, should_be_one):
+        assert type(a_list) == list
+        assert len(a_list) >= 1
+        # Part of `TestReserveTrial.test_reserve_race_condition`
+        if len(a_list) == 4:
+            assert a_list[0].status == 'new'
+        if len(a_list) == 3:
+            assert a_list[0].status == 'new'
+        if len(a_list) == 2:
+            assert a_list[0].status == 'interrupted'
+        if len(a_list) == 1:
+            assert a_list[0].status == 'suspended'
+
+        assert should_be_one == 1
+
+        if len(a_list) > 3:
+            # Set row's status as 'reserved' just like if it was reserved by
+            # another process right after the call to moptdb.read()
+            create_db_instance.write(
+                "trials",
+                data={"status": "reserved"},
+                query={"_id": a_list[0].id})
+            trial = create_db_instance.read("trials", {"_id": a_list[0].id})
+            assert trial[0]['status'] == 'reserved'
+
+        return [a_list[0]]
+
+    monkeypatch.setattr(random, 'sample', mock_sample)
+
+
+@pytest.fixture()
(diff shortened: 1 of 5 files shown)


tsirif:
I will review this after we close #36. For now Travis complains about linting:
```
src/metaopt/core/io/database/mongodb.py:124:1: D202 No blank lines allowed after function docstring
    def read_and_write(self, collection_name, query, data, selection=None):
```
and
```
Warning, treated as error:
/home/travis/build/mila-udem/metaopt/src/metaopt/algo/base.py:docstring of metaopt.algo.base.BaseAlgorithm:51:Inline literal start-string without end-string.
```

bouthilx:
Fine, I will fix all those flake8 issues. It seems I have a different configuration locally. -_-

bouthilx:
@tsirif Problems fixed, ready for review. :)

tests/unittests/core/test_experiment.py
```diff
@@ -47,6 +47,71 @@ def mock_sample(a_list, should_be_one):
     monkeypatch.setattr(random, 'sample', mock_sample)
 
 
+@pytest.fixture()
+def patch_sample_concurrent(monkeypatch, create_db_instance, exp_config):
+    """Patch ``random.sample`` to return the first one and check call."""
+    def mock_sample(a_list, should_be_one):
+        assert type(a_list) == list
+        assert len(a_list) >= 1
+        # Part of `TestReserveTrial.test_reserve_success2`
```
tsirif:
```python
# Part of `TestReserveTrial.test_reserve_race_condition`
```

tests/unittests/core/test_experiment.py
```diff
@@ -47,6 +47,71 @@ def mock_sample(a_list, should_be_one):
...
+    monkeypatch.setattr(random, 'sample', mock_sample)
+
+
+@pytest.fixture()
+def patch_sample_concurrent2(monkeypatch, create_db_instance, exp_config):
+    """Patch ``random.sample`` to return the first one and check call."""
+    def mock_sample(a_list, should_be_one):
+        assert type(a_list) == list
+        assert len(a_list) >= 1
+        # Part of `TestReserveTrial.test_reserve_success2`
```
tsirif:
```python
# Part of `TestReserveTrial.test_reserve_dead_race_condition`
```

bouthilx:
@tsirif Done

PR #54, opened 2018-03-18
Decrease unittest time for database

bouthilx:
Why:

Unittest for bad connections takes several seconds, more than the sum of
all other unit tests. This is only due to a default large Timeout inside
pymongo when server does not respond.

How:

Mock pymongo's MongoClient to force serverSelectionTimeoutMS to 1.

### DIFF
--- a/tests/unittests/core/mongodb_test.py
+++ b/tests/unittests/core/mongodb_test.py
@@ -4,6 +4,7 @@
 
 from datetime import datetime
 
+from pymongo import MongoClient
 import pytest
 
 from metaopt.core.io.database import DatabaseError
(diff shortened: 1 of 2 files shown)


tsirif:
Cool!! I wanted to do that, but I forgot it :stuck_out_tongue_closed_eyes:

bouthilx:
I don't understand. If I run `tox -e pylint` locally, I get the same error. But if I run directly `pylint src/metaopt`, it works fine. 

(By the way, I fixed the flake8 issue)

bouthilx:
@tsirif A rebase on master after the merge of #36 fixed the problem. It seems it was coming from previous master, not from the current branch.

tests/unittests/core/mongodb_test.py
```diff
@@ -18,10 +19,22 @@ def moptdb():
...
 
+@pytest.fixture()
+def patch_mongo_client(monkeypatch):
+    """Patch ``pymongo.MongoClient`` to force serverSelectionTimeoutMS to 1."""
+    def mock_class(*args, **kwargs):
+        # 1 sec, defaults to 20 secs otherwise
+        kwargs['serverSelectionTimeoutMS'] = 1.
+        return MongoClient(*args, **kwargs)
+
+    monkeypatch.setattr('metaopt.core.io.database.mongodb.MongoClient', mock_class)
```
tsirif: I think you can patch `pymongo.MongoClient` instead. I think that effects are global.

bouthilx:
This is what I tried first, but it does not work because MongoClient is
imported as `from pymongo import MongoClient`. This is equivalent to
`import pymongo.MongoClient as MongoClient`, which forces us to point to
the module `[...]mongodb.MongoClient`.

bouthilx:
I guess another solution would be to replace `MongoClient` by `pymongo.MongoClient` in `metaopt.core.io.database.mongo`. I think I prefer this one, what do you think @tsirif ?

tsirif:
@bouthilx Let's do it like that!

bouthilx:
Will do it tonight!

bouthilx:
@tsirif Done

Issue #41, opened 2018-02-26
Add index method to DB


bouthilx:
When a query is issued to MongoDB, where we ask for specific <name, user>, recent trials, completed trials, sorted top 10 result, mongo needs to iterate through the entire collection.  If we specify indexes on <name, user> for instance, it can use its built indexes to quickly fetch documents matching <name, user>.

The choice of indexes is basically a compromise between efficiency and memory consumption.

Note: Specifying <experiment name, user> as a unique index will fix a race condition bug in the experiment creation.

PR #55, opened 2018-03-18
Add indexation to databases and fix race condition for experiment configuration

bouthilx:
Related to #10, #14 and #41 .
Needs #36 and #53 to be merged.

### Why:

Our main and only database back end is a NoSQL one and only `_id` is indexed by default. That means every `read` operation (including reads necessary in updates with queries) needs to parse the entire collection unless the query specifies the _id. This is highly inefficient. Adding indexes on keys which are often used in queries will significantly improve the database operations.

Additionally, we want to force some keys to be unique, namely the tuple (experiment name, user name). To do so we need a method to specify unique indexes.

- Experiments are believed to be mostly queried by (name, user) or status, so those keys are indexed.
- Trials are mostly queried by experiment, status, results, start_time or end_time.

#### Add `ensure_index` method to `AbstractDB`.

- Keys can be specified by name and optionally by sort order too.
- Keys can be forced to be unique.
- Generic sort orders are added to `AbstractDB`.
- `MongoDB` backend uses `dbcollection.create_index()`

#### Add generic `DuplicateKeyError` and a decorator for `MongoDB` methods.
- If the wrapped method raises `pymongo.errors.DuplicateKeyError`, it is converted to generic `DuplicateKeyError`.
- If the wrapped method raises `pymongo.errors.BulkWriteError`  containing at least one "duplicate key error", it is converted to  generic `DuplicateKeyError`.

Note: `pymongo.errors.BulkWriteError` can contain different type of errors. For conversion simplicity DuplicateKeyError` will have precedence over the others.

#### Add to `cli` handling of configuration failures due to race condition.

---

### Note:
Unittest configuration contained two experiments with the same tuple
(name, user). The addition of enforced uniqueness on this tuple makes
many unittest crash.

### DIFF
--- a/src/metaopt/core/io/database/mongodb.py
+++ b/src/metaopt/core/io/database/mongodb.py
@@ -16,12 +16,22 @@ from metaopt.core.io.database import (
     AbstractDB, DatabaseError, DuplicateKeyError)
 
 
+AUTH_FAILED_MESSAGES = [
+    "auth failed",
+    "Authentication failed."]
+
+DUPLICATE_KEY_MESSAGES = [
+    "duplicate key error"]
+
+
 def mongodb_exception_wrapper(method):
     """Convert pymongo exceptions to generic exception types defined in src.core.io.database.
 
     Current exception types converted:
     pymongo.errors.DuplicateKeyError -> DuplicateKeyError
-    pymongo.errors.BulkWriteError[DuplicateKeyError] -> DuplicateKeyError
+    pymongo.errors.BulkWriteError[DUPLICATE_KEY_MESSAGES] -> DuplicateKeyError
+    pymongo.errors.ConnectionFailure -> DatabaseError
+    pymongo.errors.OperationFailure(AUTH_FAILED_MESSAGES) -> DatabaseError
 
     """
     @functools.wraps(method)
@@ -33,8 +43,16 @@ def mongodb_exception_wrapper(method):
             raise DuplicateKeyError(str(e)) from e
         except pymongo.errors.BulkWriteError as e:
             for error in e.details['writeErrors']:
-                if "duplicate key error" in error["errmsg"]:
+                if any(m in error["errmsg"] for m in DUPLICATE_KEY_MESSAGES):
                     raise DuplicateKeyError(error["errmsg"]) from e
+
+            raise
+        except pymongo.errors.ConnectionFailure as e:
+            raise DatabaseError("Connection Failure: database not found on "
+                                "specified uri") from e
+        except pymongo.errors.OperationFailure as e:
+            if any(m in str(e) for m in AUTH_FAILED_MESSAGES):
+                raise DatabaseError("Authentication Failure: bad credentials") from e
 
             raise
 
@@ -59,6 +77,7 @@ class MongoDB(AbstractDB):
 
     """
 
+    @mongodb_exception_wrapper
     def initiate_connection(self):
         """Connect to database, unless MongoDB `is_connected`.
 
@@ -70,23 +89,13 @@ class MongoDB(AbstractDB):
 
         self._sanitize_attrs()
 
-        try:
-            self._conn = pymongo.MongoClient(host=self.host,
-                                             port=self.port,
-                                             username=self.username,
-                                             password=self.password,
-                                             authSource=self.name)
-            self._db = self._conn[self.name]
-            self._db.command('ismaster')  # .. seealso:: :meth:`is_connected`
-        except pymongo.errors.ConnectionFailure as e:
-            self._logger.error("Could not connect to host, %s:%s",
-                               self.host, self.port)
-            raise DatabaseError("Connection Failure: database not found on "
-                                "specified uri") from e
-        except pymongo.errors.OperationFailure as e:
-            self._logger.error("Could not verify user, %s, on database, %s",
-                               self.username, self.name)
-            raise DatabaseError("Authentication Failure: bad credentials") from e
+        self._conn = pymongo.MongoClient(host=self.host,
+                                         port=self.port,
+                                         username=self.username,
+                                         password=self.password,
+                                         authSource=self.name)
+        self._db = self._conn[self.name]
+        self._db.command('ismaster')  # .. seealso:: :meth:`is_connected`
 
     @property
     def is_connected(self):
--- a/src/metaopt/core/io/database/mongodb.py
+++ b/src/metaopt/core/io/database/mongodb.py
@@ -82,8 +82,43 @@ class MongoDB(AbstractDB):
         """
         self._conn.close()
 
-    def write(self, collection_name, data,
-              query=None):
+    def ensure_index(self, collection_name, keys, unique=False):
+        """Create given indexes if they do not already exist in database.
+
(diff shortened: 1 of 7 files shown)


bouthilx:
@tsirif Problems fixed, ready for review. :)

tsirif:
This will resolve #41 also!

tests/unittests/core/test_experiment.py
```diff
@@ -288,32 +354,57 @@ def test_get_after_init_plus_hit_no_diffs(self, exp_config):
...
         # Deliver an external configuration to finalize init
-        exp.configure(exp_config[0][1])
+        exp.configure(exp_config[0][0])
         assert exp._init_done is True
         with pytest.raises(RuntimeError) as exc_info:
-            exp.configure(exp_config[0][1])
+            exp.configure(exp_config[0][0])
         assert 'cannot reset' in str(exc_info.value)
 
+    def test_try_set_after_race_condition(self, exp_config, new_config):
```
tsirif: So this test demonstrates that if two Experiment perceive that `is_new is True`, then only one of them will actually configure and that the other one will raise a DuplicateError. The exception will get caught by `metaopt.cli.create_experiment` and creation will be retries. All second attempts for creation will succeed (assuming no forking conditions) because the second time `is_new is False`! Is this correct?

src/metaopt/core/io/database/mongodb.py
```diff
@@ -84,8 +114,44 @@ def close_connection(self):
...
+        if not isinstance(keys, (list, tuple)):
+            keys = [(keys, self.ASCENDING)]
+
+        converted_keys = []
+        for key, sort_order in keys:
+            converted_keys.append((key, self._convert_sort_order(sort_order)))
+
+        return converted_keys
+
+    def _convert_sort_order(self, sort_order):
```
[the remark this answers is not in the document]
bouthilx: Yes, exactly!

tests/unittests/core/test_experiment.py
```diff
@@ -37,16 +37,82 @@ def mock_sample(a_list, should_be_one):
...
 
 
+@pytest.fixture()
+def patch_sample_concurrent(monkeypatch, create_db_instance, exp_config):
+    """Patch ``random.sample`` to return the first one and check call."""
+    def mock_sample(a_list, should_be_one):
+        assert type(a_list) == list
+        assert len(a_list) >= 1
+        # Part of `TestReserveTrial.test_reserve_success2`
+        print(a_list)
```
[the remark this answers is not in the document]
bouthilx: :scream: Thanks! :laughing:

tests/unittests/core/mongodb_test.py
```diff
@@ -64,22 +66,130 @@ def test_singleton(self):
...
+
+        This is because _id inside exp_config[0][0] cannot be set. It is an
+        immutable key of the collection.
+
+        """
+        config_to_add = exp_config[0][0]
+
+        query = {'_id': exp_config[0][1]['_id']}
+
+        with pytest.raises(pymongo.errors.OperationFailure):
```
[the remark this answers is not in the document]
bouthilx: The thing is OperationFailure is a pretty broad exception and we don't want to convert exception causes that we don't know. If there is a specific OperationFailure we know and need to convert, then we should verify its message before converting. My main concern with this method for conversion is that pymongo's error messages might vary from versions to others. I'd say we try like it is and extent the catches if necessary later on.

tests/unittests/core/test_experiment.py
bouthilx: Yes! Quite accuracy. If I may say, all experiment racing will configure, but only one will be able to write its configuration to DB, while the others will hit a DuplicateKeyError. My intention was to test the retry inside the `cli` unittests, but maybe we should also test it at a lower level on `Experiment` directly, what do you think?

bouthilx:
@tsirif Done. Please don't merge before I squash 127a0e6 and 63e9522. I let them there to ease the flow of feedback and discussion until we consider it ready for merge.

tests/unittests/core/test_experiment.py
tsirif: and mark it xfail? why not? it will help polishing UI details later :stuck_out_tongue:

tests/unittests/core/mongodb_test.py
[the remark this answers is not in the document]
tsirif: @bouthilx I understand your point. So let the reraise happen exactly where the exception is raised; where the reason for `OperationFailure` cannot mean multiple stuff.

tests/unittests/core/test_experiment.py
bouthilx: But it won't fail. I've added it in 4bde5c9

tsirif: @bouthilx last reply was meant for #57...

I think that even if the first one will only write a new configuration with unique (exp_name, user_name), all the rest will eventually enter the branch where `is_new is False` and they will still call `_db.write(...)` but they are going to update it to be exactly the same thing.

As these two are the only options:
```python
        if is_new:
            self._db.write('experiments', final_config)
            # XXX: Reminder for future DB implementations:
            # MongoDB, updates an inserted dict with _id, so should you :P
            self._id = final_config['_id']
        else:
            self._db.write('experiments', final_config, {'_id': self._id})
```

It could verified by a mock at `_db.write` and testing against the order of the calls to that function and which arguments they call it with. Perhaps, it could be an informative test. I am not sure..

bouthilx: The `else`-clause will not update the document to exactly the same thing. Some changes do not trigger a fork, `max_trials` for instance. I would verify such modifications. Don't we have a unit test doing that?

tests/unittests/core/mongodb_test.py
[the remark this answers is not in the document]
bouthilx: Well, I tried something at 1a46674. What do you think of it?

src/metaopt/core/io/database/mongodb.py
```diff
@@ -16,12 +16,22 @@
...
+
 def mongodb_exception_wrapper(method):
     """Convert pymongo exceptions to generic exception types defined in src.core.io.database.
 
     Current exception types converted:
     pymongo.errors.DuplicateKeyError -> DuplicateKeyError
-    pymongo.errors.BulkWriteError[DuplicateKeyError] -> DuplicateKeyError
+    pymongo.errors.BulkWriteError[DUPLICATE_KEY_MESSAGES] -> DuplicateKeyError
+    pymongo.errors.ConnectionFailure -> DatabaseError
+    pymongo.errors.OperationFailure(AUTH_FAILED_MESSAGES) -> DatabaseError
```
tsirif: OCD here... above it is with square brackets, here it is with paretheses... sorry for that :nerd_face:

tsirif: @bouthilx this one

bouthilx: I would agree, but [] vs () has a meaning. BulkWriteError contains a list of errors, from which we convert the one with DUPLICATE_KEY_MESSAGES. OperationFailure has only a message.

PR #62, opened 2018-04-22
Make `random` algorithm be the default choice

tsirif:
Also, fix a minor bug in hierarchical configuration.


bouthilx:
@tsirif Can you add a unit test about this bug.

bouthilx:
@tsirif Needs rebase

PR #59, opened 2018-04-02
Fix README.rst, create website page, document using and developping 

tsirif:
Related to issue #39

Not yet completed. Pull requesting just to get the touch of things

PR #63, opened 2018-04-22
Again some renaming for metaopt to orion stuff

tsirif:


src/orion/core/resolve_config.py
```diff
@@ -75,7 +75,7 @@ def is_exe(path):
 # list containing tuples of
 # (environmental variable names, configuration keys, default values)
 ENV_VARS_DB = [
-    ('ORION_DB_NAME', 'name', 'Orion'),
+    ('ORION_DB_NAME', 'name', 'Oríon'),
```
lamblin: I'm not ideologically opposed to the accent in strings and so on, on the contrary (we have "Université de Montréal" at a couple places), but the name of the DB itself would be pushing it IMO.

docs/src/conf.py
```diff
@@ -60,8 +60,8 @@
 # General information about the project.
 project = u'orion'
 _full_version = orion.__version__
-author = u'Orion Team - MILA, Université de Montréal'
-copyright = u'2017, Orion Team - MILA, Université de Montréal'
+author = u'Oríon Team - MILA, Université de Montréal'
+copyright = u'2017, Oríon Team - MILA, Université de Montréal'
```
lamblin: 2017--2018?

docs/src/conf.py
```diff
@@ -140,7 +140,7 @@
 # -- Options for HTMLHelp output ------------------------------------------
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = 'Oriondoc'
+htmlhelp_basename = 'Oríondoc'
```
lamblin: Is that part of the URL? If so I would remove the accent.

bouthilx: I agree

bouthilx:
As we already have Universit**é** de Montr**é**al, why not having Or**í**on. I agree with Pascal for DB names and URLs.

tsirif:
Agreed

bouthilx:
@tsirif Quick fix and merge? :)

docs/src/conf.py
tsirif: I fixed this in 'feature/docs' #59

docs/src/index.rst
```diff
@@ -1,9 +1,9 @@
-.. Orion documentation master file, created by
+.. Oríon documentation master file, created by
    sphinx-quickstart on Mon Nov 13 19:08:05 2017.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to Orion's documentation!
+Welcome to Oríon's documentation!
 ===================================
```
lamblin: I missed this underline that's too long

lamblin:
Probably still failing because of sphinx 1.7.3

docs/src/index.rst
tsirif: Don't worry about it. I am changing it completely in #59

bouthilx:
@tsirif Needs rebase

PR #67, opened 2018-05-01
Add default value support for dimensions

DatCorno:
Why: Default values are necessary when inserting a trial manually and
giving only a subset of the space of an experiment as fixed dimensions.
When a dimension is missing, it'll go look inside the configuration and
see if the default value has been set, if not, it'll stop.

How: By adding the default value as a named parameters of the distribution,
the dimension can check if it belongs to its domain and cast it to the correct
type. This limits the actual code impact since the parsing of the parameters
was already done inside the _build function of DimensionBuilder.

### DIFF
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -86,6 +86,8 @@ class Dimension(object):
         """
         self._name = None
         self.name = name
+        self._default_value = None
+
         if 'random_state' in kwargs or 'seed' in kwargs:
             raise ValueError("random_state/seed cannot be set in a "
                              "parameter's definition! Set seed globally!")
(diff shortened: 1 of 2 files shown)


src/orion/algo/space.py
```diff
@@ -181,6 +187,21 @@ def name(self, value):
         else:
             raise TypeError("Dimension's name must be either string or None. "
                             "Provided: {}, of type: {}".format(value, type(value)))
+    
+    @property
+    def default_value(self):
+        return self._default_value
+
+    @default_value.setter
+    def default_value(self, value):
```
bouthilx: I'm not sure we should have a public setter. Dimensions object are basically immutable objects, with the exception of their random number generator in ```self.prior```.

PR #69, opened 2018-05-07
User can request managerial info, like trial string identifiers, from cli

tsirif:
Why:
   I need to way to identify saved stuff in local disk from the
   execution of my script given a particular trial. This way I can have
   a reproducible deterministic string identifier that I intend to pass
   appropriately to a command line argument (if specified from ``orion
   hunt``) of my executable. This will disambiguate saved states,
   snapshots, or logs (i.e. from tensorboard), among different trials of
   the same executable.

   The actual implementation of all save/load is left to the executable
   itself, as well as configuring the absolute path where the stuff will
   be saved to. Most (responsible) ML programs I have personally encountered
   have a cli way to specify some naming for the experiment to run.

### DIFF
--- a/src/orion/core/io/space_builder.py
+++ b/src/orion/core/io/space_builder.py
@@ -240,7 +240,7 @@ class SpaceBuilder(object, metaclass=SingletonType):
     USERARGS_NAMESPACE = r'\W*([a-zA-Z0-9_-]+)'
     USERARGS_CONFIG = '--config='
 
-    EXPOSED_PROPERTIES = ['trial.hash_name', 'trial.full_name']
+    EXPOSED_PROPERTIES = ['trial.hash_name', 'trial.full_name', 'exp.name']
 
     def __init__(self):
         """Initialize a `SpaceBuilder`."""
@@ -381,12 +381,7 @@ class SpaceBuilder(object, metaclass=SingletonType):
                 # It's a experiment/trial management command
                 namespace = '$'
                 namespace += ns_search[0] if ns_search else get_next_pos_ns()
-                try:
-                    objname, attrname = expression.split('.')
-                except ValueError as exc:
-                    raise ValueError("Expected dotted expression with two parts "
-                                     "'{{object}}.{{attribute}}'. "
-                                     "Got: '{}'".format(expression)) from exc
+                objname, attrname = expression.split('.')
                 self.commands_tmpl[namespace] = (objname, attrname)
                 self.userargs_tmpl[namespace] = prefix + '=' if ns_search else ''
             elif not ns_search or not expression or expression[0] == '/':
@@ -417,14 +412,18 @@ class SpaceBuilder(object, metaclass=SingletonType):
 
         return userconfig
 
-    def build_to(self, config_path, trial):
+    def build_to(self, config_path, trial, experiment=None):
         """Use templates saved from `build_from` to generate a config file (if needed)
         and command line arguments to correspond to specific parameter selections.
 
         :param config_path: Path in which the configuration file instance
            will be created.
         :param trial: A `orion.core.worker.trial.Trial` object with concrete
-           parameter values for the defined `Space`.
+           parameter values for the defined `Space`. It may be used to retrieve
+           managerial information from it.
+        :param experiment: An `orion.core.worker.experiment.Experiment` object.
+           It may be used to retrieve managerial information from it. See
+           :attr:`EXPOSED_PROPERTIES`.
 
         :returns: A list with the command line arguments that must be given to
            script's execution.
@@ -432,7 +431,7 @@ class SpaceBuilder(object, metaclass=SingletonType):
         """
         if self.userconfig:
             self._build_to_config(config_path, trial)
-        return self._build_to_args(config_path, trial)
+        return self._build_to_args(config_path, trial, experiment)
 
     def _build_to_config(self, config_path, trial):
         config_instance = copy.deepcopy(self.userconfig_tmpl)
@@ -459,10 +458,10 @@ class SpaceBuilder(object, metaclass=SingletonType):
 
         self.converter.generate(config_path, config_instance)
 
-    def _build_to_args(self, config_path, trial):
+    def _build_to_args(self, config_path, trial, experiment=None):
         cmd_args = []
         # objects whose properties can be fetched to fill a cli argument
-        exposed_objects = {'trial': trial}
+        exposed_objects = {'trial': trial, 'exp': experiment}
         param_names = [trial.params[i].name for i in range(len(trial.params))]
 
         for namespace, prefix in self.userargs_tmpl.items():
--- a/src/orion/core/io/space_builder.py
+++ b/src/orion/core/io/space_builder.py
@@ -66,6 +66,20 @@ def _real_or_int(kwargs):
     return Integer if kwargs.pop('discrete', False) else Real
 
 
+def replace_key_in_order(odict, key_prev, key_after):
+    """Replace `key_prev` of `OrderedDict` `odict` with `key_after`,
+    while leaving its value and the rest of the dictionary intact and in the
+    same order.
+    """
+    tmp = collections.OrderedDict()
+    for k, v in odict.items():
+        if k == key_prev:
+            tmp[key_after] = v
+        else:
+            tmp[k] = v
+    return tmp
+
+
 class DimensionBuilder(object, metaclass=SingletonType):
     """Create `Dimension` objects using a name for it and an string expression
     which encodes prior and dimension information.
@@ -220,18 +234,20 @@ class DimensionBuilder(object, metaclass=SingletonType):
 class SpaceBuilder(object, metaclass=SingletonType):
     """Build a `Space` object form user's configuration."""
 
-    USERCONFIG_OPTION = '--config='
+    # TODO Expose these 4 USER oriented goodfellows to a orion configuration file :)
     USERCONFIG_KEYWORD = 'orion~'
-    USERARGS_SEARCH = r'\W*([a-zA-Z0-9_-]+)~(.*)'
     USERARGS_TMPL = r'(.*)~(.*)'
+    USERARGS_NAMESPACE = r'\W*([a-zA-Z0-9_-]+)'
+    USERARGS_CONFIG = '--config='
+
     EXPOSED_PROPERTIES = ['trial.hash_name', 'trial.full_name']
 
     def __init__(self):
         """Initialize a `SpaceBuilder`."""
         self.userconfig = None
-        self.is_userconfig_an_option = None
         self.userargs_tmpl = None
         self.userconfig_tmpl = None
+        self.positional_args_count = 0
 
         self.dimbuilder = DimensionBuilder()
         self.space = None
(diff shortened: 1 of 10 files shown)


src/orion/core/io/space_builder.py
```diff
@@ -289,60 +322,114 @@ def _build_from_config(self, config_path):
...
 
     def _build_from_args(self, cmd_args):
+        """Build templates from arguments found in the original cli.
+
+        Rules for namespacing:
+           1. Prefix ``'/'`` is given to parameter dimensions.
+           2. Prefix ``'$'`` is given to substitutions from exposed properties.
+           3. Prefix ``'_'`` is given to arguments which do not interact with.
+              Suffix is a unique integer.
+           4. ``'config'`` is given to user's script configuration file template
```
bouthilx: There should be an example. The descriptions alone are not clear enough.

src/orion/core/io/space_builder.py
```diff
@@ -289,60 +322,114 @@ def _build_from_config(self, config_path):
...
+
+        User script configuration file argument is treated specially, trying to
+        recognise a ``--config`` option or by checking the first positional
+        argument, if not found elsewhere. Only one is allowed.
+
+        .. note:: Currently positional arguments cannot define a parameter
+           dimension. And I don't think they will; it wouldn't be good for
```
bouthilx: We should not use first person singular in docstrings. We can't use positional arguments because they don't have a name assigned to. I think that's enough justification.

src/orion/core/io/space_builder.py
```diff
@@ -218,25 +232,42 @@ def build(self, name, expression):
...
 
-    USERCONFIG_OPTION = '--config='
+    # TODO Expose these 4 USER oriented goodfellows to a orion configuration file :)
     USERCONFIG_KEYWORD = 'orion~'
-    USERARGS_SEARCH = r'\W*([a-zA-Z0-9_-]+)~(.*)'
     USERARGS_TMPL = r'(.*)~(.*)'
+    USERARGS_NAMESPACE = r'\W*([a-zA-Z0-9_-]+)'
+    USERARGS_CONFIG = '--config='
+
+    EXPOSED_PROPERTIES = ['trial.hash_name', 'trial.full_name', 'exp.name']
```
bouthilx: If I understand properly, if there is some `$trial.hash_name` in the command line of the user, this will be replaced by the actual trial.hash_name provided by orion. Is that right?

src/orion/core/worker/trial.py
```diff
@@ -192,8 +194,8 @@ def to_dict(self):
 
     def __str__(self):
         """Represent partially with a string."""
-        ret = "Trial(experiment={0}, status={1}, params.value={2})".format(
-            repr(self.experiment), repr(self._status), [p.value for p in self.params])
+        ret = "Trial(experiment={0}, status={1},\n      params={2})".format(
+            repr(self.experiment), repr(self._status), self.params_repr(sep=('\n' + ' ' * 13)))
```
bouthilx: What about using `pprint` if we want some fancy `repr`?

src/orion/core/worker/trial.py
```diff
@@ -236,6 +238,27 @@ def is_registered(self):
...
 
+    def params_repr(self, sep=','):
+        """Represent with a string the parameter contained in this `Trial` object."""
+        return sep.join(map(lambda param: "{0.name}:{0.value}".format(param), self.params))
+
+    @property
+    def hash_name(self):
+        """Generate a unique name with an md5sum hash for this `Trial`.
+
+        .. note:: Two trials that have the same `params` must have the same `hash_name`.
```
bouthilx: I disagree. Two trials with identical parameters could have different results. We can't assume the user's script is deterministic. We could, however, add a `seed` to the attributes of the Trial and pass it through an environment variable to the user's script. The user could seed the script using it, making it deterministic for a given pair of (params, seed).

tests/functional/demo/orion_config.yaml
```diff
@@ -6,6 +6,7 @@ max_trials: 100
 algorithms:
   gradient_descent:
     learning_rate: 0.1
+    # dx_tolerance: 1e-7
```
bouthilx: Why comment out? Should we remove it?

tests/functional/demo/test_demo.py
```diff
@@ -157,3 +159,55 @@ def test_workon(database):
...
+    how_many = 1000
+    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
+    process = subprocess.Popen(["orion", "--max-trials={}".format(how_many),
+                                "--pool-size=1",
+                                "--name=lalala",
+                                "--config", "./stress_gradient.yaml",
+                                "./dir_per_trial.py",
+                                "--dir={}".format(str(tmpdir)),
+                                "--other-name~exp.name",
+                                "--name~trial.hash_name",
```
bouthilx: We use `~` for both exposed variables and distributions? Wasn't it supposed to be `$` in this case?

tests/functional/demo/test_demo.py
```diff
@@ -157,3 +159,55 @@ def test_workon(database):
...
+    assert rcode == 0
+    # For contingent broken trials, which in this test means that a existing
+    # directory was attempted to be created, it means that it's not md5 or
+    # bad hash creation to blame, but the finite precision of the floating
+    # point representation. Specifically, it seems that gradient descent
+    # is able to reach such levels of precision jumping around the minimum
+    # (notice that an appropriate learning rate was selected in this stress
+    # test to create underdamped behaviour), that it begins to suggest same
+    # things from the past. This is intended to be shown with the assertions
+    # in the for-loop below.
```
bouthilx: Why not used random search if gradient descent causes side-effects you need to verify to make sure the problem is not md5 of bad hash creation.

tests/unittests/core/test_space_builder.py
```diff
@@ -221,18 +221,27 @@ def test_build_from_args_only(self, spacebuilder):
...
         assert len(space) == 2
         assert '/yolo' in space
         assert '/arch2' in space
-        assert len(spacebuilder.userargs_tmpl) == 3
-        assert spacebuilder.userargs_tmpl[None] == ["--seed=555",
-                                                    "--arch1=choices({'lala': 0.2, 'yolo': 0.8})"]
-        assert spacebuilder.userargs_tmpl['/yolo'] == '-yolo='
-        assert spacebuilder.userargs_tmpl['/arch2'] == '--arch2='
+        assert list(spacebuilder.userargs_tmpl.keys()) == ['_0', '/yolo', '_1', '/arch2',
+                                                           '$2', '_3', '_4', '_5']
```
bouthilx: Ah, now I understand what the `$` was for.

tests/functional/demo/test_demo.py
bouthilx: Nevermind

src/orion/core/io/space_builder.py
bouthilx: Nevermind

src/orion/core/worker/trial.py
tsirif: of course, but later

src/orion/core/worker/trial.py
tsirif: Is it the case that we may request from the script to evaluate a particular set of parameters twice? Then, that's a design issue that we did not take care about.

only if the `params` are the same, `hash_name` is going to be, not the `results`.

#71 is this the one that you opened because of this observation?

tests/functional/demo/orion_config.yaml
tsirif: To show that it is the default, but not used here

tests/functional/demo/test_demo.py
tsirif: For the reason that even in the case where `Δx_t` is the minimum possible in the finite float representation scenario, which can be reached by gradient descent for example, the demand for `diff hash_names => diff parameters` is true. By random search I wouldn't be able to have that stressed behaviour.

src/orion/core/worker/trial.py
bouthilx: Yes I opened #71 for this.

If two trials with identical parameters give different results then they should be considered different, hence have different hash names.

For hyper-parameter optimization, it would not make sense to run twice the exact same hyper-parameters unless they are all categoricals or very small interval of integers. If any is dimension is real or large interval of integers, than the optimization algorithm is better to poke around than running twice identical sets.

However, someone could want to run multiple trials with identical params to compute the variance of the performance. This is something we could support.

To make it clear, what I propose is to keep it like you did for now, our first algorithms won't use seed information anyway. But I believe we should add support for seeds later on, and include the seed information in what we use to compute the hash name of a trial. A change like that should not break anything so it's ok to postpone it. (it's possible to keep backward compatibility for prior trials after we add support for seeds)

tests/functional/demo/orion_config.yaml
bouthilx: ok, that's fine

src/orion/core/io/space_builder.py
```diff
@@ -337,10 +337,9 @@ def _build_from_args(self, cmd_args):
         recognise a ``--config`` option or by checking the first positional
         argument, if not found elsewhere. Only one is allowed.
 
-        .. note:: Currently positional arguments cannot define a parameter
-           dimension. And I don't think they will; it wouldn't be good for
-           your health down the line. Please be responsible and give a good
-           descriptive name to your inputs.
+        .. note:: Positional arguments cannot define a parameter
+           dimension, because no name, but an arbitrary (semantically) one, can
```
bouthilx: I understand what you mean but it seems to me that it could be explained in more simple words. Just dropping "but an arbitrary (semantically) one" would make it more simpler. I don't think the note needs to justify why we do not assign random strings as hyper-parameter names when none are explicitly given.

src/orion/core/worker/trial.py
tsirif: `hash_name` is supposed to be generated and given as an input to user's script so that it is able to disambiguate local directories that it may use (i.e. for snapshots or logs) among different hyperparameter trial sets. It is supposed to be generated by trials whose results are not yet known, as is the use case above. So this hash_name corresponds only to `trial.params` only. To make it able to run multiple experiments for the same set of hyperparameters:
1. yes we should provide the seed to the user's script. I predict a `~trial.seed` kind of command in the likes of what I've built here. So, `seed` is going to be an attribute of a `Trial` as well.
2. `hash_name` will be generated as a function of `(trial.params, trial.seed)` to ensure uniqueness of results and reproducibility.

tsirif:
We could have assigned `_5` as a name. Let's keep like this, please.

src/orion/core/io/space_builder.py
bouthilx: > We could have assigned `_5` as a name. Let's keep like this, please

Well that would be terrible. It would be very hard to understand what is `_5` when the user try to analyze the results latter on.

The documentation needs to be clear and concise. I'm sorry but this is not.

tsirif: `Positional arguments cannot define a parameter dimension, because no **meaningful** name can be assigned` ?

bouthilx: :ok_hand:

PR #74, opened 2018-05-10
Fix: Convert ndarrays to lists because they are not bsonable

tsirif:
**Case**:
Use random optimization for a single (3,)-shaped dimension.

**Reproduce**:
In bouthilx/orion.algo.skopt#3 run:
```
export ORION_DB_NAME=orion_test
export ORION_DB_ADDRESS=mongodb://user:pass@localhost
orion -n=test --pool-size=1 --max-trials=10 ./tests/rosenbrock.py -x~'uniform(-5, 5, shape=3)'
```

**Traceback**:
from ``worker.__init__.py``:line 42 (workon)
``producer.produce()``
from ``worker.experiment.py``:line 228 (register_trials)
``self._db.write('trials', trials_dicts)``
from ``mongodb.py``:line 41 (write)
``result = dbcollection.insert_many(documents=data)``

**Raises**:
``bson.errors.InvalidDocument: Cannot encode object: array([0.8, 1.3])``

### DIFF
--- a/src/orion/core/utils/format_trials.py
+++ b/src/orion/core/utils/format_trials.py
@@ -34,11 +34,17 @@ def tuple_to_trial(data, space):
     :type space: `orion.algo.space.Space`
     """
     assert len(data) == len(space)
-    params = [dict(
-        name=space[order].name,
-        type=space[order].type,
-        value=data[order]
-        ) for order in range(len(space))]
+    params = []
+    for i, dim in enumerate(space.values()):
+        try:
+            datum = data[i].tolist()  # if it is numpy.ndarray
+        except AttributeError:
+            datum = data[i]
+        params.append(dict(
+            name=dim.name,
+            type=dim.type,
+            value=datum
+            ))
     return Trial(params=params)
 
 
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -601,3 +601,52 @@ class Space(OrderedDict):
         """Represent as a string the space and the dimensions it contains."""
         dims = list(self.values())
         return "Space([{}])".format(',\n       '.join(map(str, dims)))
+
+
+def pack_point(point, space):
+    """Take a list of points and pack it appropriately as a point from `space`.
+
+    :param point: array-like or list of numbers
+    :param space: problem's parameter definition,
+       instance of `orion.algo.space.Space`
+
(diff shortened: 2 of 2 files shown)


src/orion/core/utils/format_trials.py
```diff
@@ -34,11 +34,17 @@ def tuple_to_trial(data, space):
...
+    params = []
+    for i, dim in enumerate(space.values()):
+        try:
+            datum = data[i].tolist()  # if it is numpy.ndarray
+        except AttributeError:
+            datum = data[i]
+        params.append(dict(
+            name=dim.name,
+            type=dim.type,
+            value=datum
```
bouthilx: But then trial_to_tuple will not have numpy arrays but lists. That could be problematic for algorithms. I believe there is a way to save numpy arrays in MongoDB. My laptop battery is about to die now. I'll look for this when I get home.

tsirif: I suppose that algorithms would like to have a way to **unpack** a `point in space` into a numpy.ndarray and **repack** a numpy.ndarray into a `point in space`. Points as they are shown in our test cases, are not favouring easy usage with numpy itself, but easy formatting them into the user script's input template. This is because a single point in a space with multiple Dimensions looks like: `(sp0, sp1, ..., spN)` where sp0 can be just a floating point, sp1 a flat list, and spN a numpy.array. Once you have the time, check out the two helper functions I introduced in bouthilx/orion.algo.skopt#3, they could be part of `orion.algo.base` I believe and their responsibility is to transform stuff into a single flat numpy array and back.

bouthilx: ok so this problem occurs when we have a dimension with shape greater than 1. I would agree to migrate `pack_point` and `unpack_point` to `orion.algo.base`.

Still, I don't like the fact that if some_trial_tuple contains some numpy array, then the conversions diverges:
```python
some_trial_tuple != trial_to_tuple(tuple_to_trial(some_trial_tuple))
```
(abuse of notation with !=)

tsirif: Oh that's what you mean. Yeah I dislike this too.

dendisuhubdy:
pulling it! Thanks @bouthilx

bouthilx:
Was about time!

PR #70, opened 2018-05-07
Feature/cli commands 2.0

DatCorno:
This PR changes the way Orion works by adding support to modular commands parsing and execution.

PR #76, opened 2018-05-14
Feature/cli commands 3.0

bouthilx:
Follow-up on #70
> This PR changes the way Orion works by adding support to modular commands parsing and execution.

Additions:
- Rework unit-tests to support mocking getpass.getuser for unit-tests.
- Modify parsers method names for more coherence
- Shield Real/Integer dimension `__contains__` methods against non-numeric values


bouthilx:
Unit-tests coverage dropped mostly because many functions about experiment configuration are copied over cli commands (hunt.py, insert.py, init_only.py). Those will be unified in a later PR so we merge this one even if coverage is low.

PR #80, opened 2018-05-16
Add full ref to README

bouthilx:
Note: PR on **develop**

Why:

Links for the documentation could not be inferred before the readthedocs
account was online. Now that we have those links, they are added to the
README file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,9 @@ now what you have to do looks like this:
 
 ``orion -n experiment_name ./run.py --mini-batch~'randint(32, 256)'``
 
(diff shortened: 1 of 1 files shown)


tsirif:
Just a completely minor question. Is there a reason for using `http` instead of `https` in the links?

bouthilx:
Not really, all links from readthedocs.io are http, I copy pasted. That's why.

bouthilx:
@tsirif Any request for https before we merge?

tsirif:
we could have it as `https`, but I am not having objections otherwise as well.

bouthilx:
Why not, I amended my commit.

PR #82, opened 2018-05-16
Add full ref to README

bouthilx:
Note: PR on **master**

Why:

Links for the documentation could not be inferred before the readthedocs
account was online. Now that we have those links, they are added to the
README file.


tsirif:
@bouthilx ok if we have already released. Now it's not useful; I was confused when I saw that again because it just hangs waiting for something.

bouthilx:
Alright, there is the auto-api to fix anyway before we release.

PR #83, opened 2018-05-16
Add adapters for experiment version control

bouthilx:
Depends on #86 

Why:

Experiment branches because of changes in the configuration or
the user's code. This make experiments incompatible with one another
unless we define adapters such that a branched experiment B can access
trials from parent experiment A.

How:

Adapters have two main methods, forward and backward. The method forward
defines how trials from parent experiment A are filtered or adapted to
be compatible with experiment B. Modifications are only applied at
execution time and are not saved anywhere in the database. Adapters only
provides a *view* on an experiment.

There is adapters for

* Dimension addition
* Dimension deletion
* Dimension renaming
* Change of dimension prior
* Change of algorithm
* Change of code
* Combining different adapters

Adapters all have a `to_dict` method which provides sufficient
information to rebuild the adapter. This is to facilitate save of
adapters in a database and retrieval.

Adapters can be build using the factory class Adapter(**kwargs) or using
Adapter.build(list_of_dicts).

### DIFF
--- a/src/orion/core/evc/adapters.py
+++ b/src/orion/core/evc/adapters.py
@@ -1,0 +1,694 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.core.evc.adapters` -- Adapters to connect experiments within the EVC system
+=======================================================================================
+
+.. module:: adapters
+   :platform: Unix
+   :synopsis: Adapters to connect experiments within the EVC system
+
+Experiment branches because of changes in the configuration or the user's code. This make
+experiments incompatible with one another unless we define adapters such that a branched experiment
+B can access trials from parent experiment A.
+
+Adapters have two main methods, forward and backward. The method forward defines how trials from
+parent experiment A are filtered or adapted to be compatible with experiment B. Modifications are
+only applied at execution time and are not saved anywhere in the database. Adapters only provides a
+view on an experiment.
+
+There is adapters for
+
+    * Dimension addition
+    * Dimension deletion
+    * Dimension renaming
+    * Change of dimension prior
+    * Change of algorithm
+    * Change of code
+    * Combining different adapters
+
+Adapters all have a `to_dict` method which provides sufficient information to rebuild the adapter.
+This is to facilitate save of adapters in a database and retrieval.
+
+Adapters can be build using the factory class `Adapter(**kwargs)` or using
+`Adapter.build(list_of_dicts)`.
+
+"""
+from abc import (ABCMeta, abstractmethod)
+import copy
+
+from orion.core.io.space_builder import DimensionBuilder
+from orion.core.utils import Factory
+from orion.core.worker.trial import Trial
+
+
+class BaseAdapter(object, metaclass=ABCMeta):
+    """Base class describing what an adapter can do."""
+
+    @abstractmethod
+    def forward(self, trials):
+        """Adapt trials of the parent experiment such that they are compatible to the child
+        experiment
+
+        Parameters
+        ----------
+        trials: list of :class:`orion.core.worker.trial.Trial`
+            List of :class:`orion.core.worker.trial.Trial` coming from the parent experiment
+
+        """
+        pass
+
+    @abstractmethod
+    def backward(self, trials):
+        """Adapt trials of the child experiment such that they are compatible to the parent
+        experiment
+
+        Parameters
+        ----------
+        trials: list of :class:`orion.core.worker.trial.Trial`
+            List of :class:`orion.core.worker.trial.Trial` coming from the child experiment
+
+        """
+        pass
+
+    @abstractmethod
+    def to_dict(self):
+        """Provide the configuration of the adapter as a dictionary
+
+        This method is intended for single adapters only.
+        For coherence, method configuration is used by all adapters, which is a list of dictionary
+        configurations.
+
+        .. seealso::
+
+            :meth:`orion.core.evc.adapters.BaseAdapter.configuration`
+        """
+        pass
+
+    @property
+    def configuration(self):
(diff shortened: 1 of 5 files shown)


src/orion/core/worker/trial.py
```diff
@@ -106,6 +106,10 @@ def to_dict(self):
                 )
             return ret
 
+        def __eq__(self, other):
```
tsirif: This should have been done earlier. It is useful fo the `hash_name` property as well. Shall I right a `__hash__` and change the implementation of the `hash_name`? I think I should. This will be in a future PR.

src/orion/core/worker/trial.py
```diff
@@ -132,13 +136,15 @@ class Result(Value):
         function or of an 'constraint' expression.
         """
 
+        __slots__ = ('name', '_type', 'value')
```
tsirif: Oh yes. Sorry for that. The tuple defining the slots can be still in `Trial.Value`. What needs to be done is to declare `__slots__ = ()` in `Trial.Param` and `Trial.Result`, so that a namespace `__dict__` won't be created for `Trial.Param` or `Trial.Result`. An object instantiated from the inherited classes will be able only have access to the non-blank namespace defined by the base class.

src/orion/core/worker/trial.py
```diff
@@ -132,13 +136,15 @@ class Result(Value):
...
         allowed_types = ('objective', 'constraint', 'gradient')
 
     class Param(Value):
         """Types for a `Param` can be either an integer (discrete value),
         floating precision numerical or a categorical expression (e.g. a string).
         """
 
+        __slots__ = ('name', '_type', 'value')
```
tsirif: same here

tests/unittests/core/test_utils.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Test base fonctionnalities of :mod:`orion.core.utils`."""
```
tsirif: Oh typo here, bad me

src/orion/core/worker/trial.py
bouthilx: Oh! I will try that.

src/orion/core/worker/trial.py
bouthilx: Yes sure if there is hashing that is certainly related to __eq__.

src/orion/core/evc/adapters.py
```diff
@@ -0,0 +1,639 @@
...
+
+            :meth:`orion.core.evc.adapters.BaseAdapter.to_dict`
+        """
+        ret = dict(
+            of_type=self.__class__.__name__.lower(),
+            param=self.param.to_dict())
+        return ret
+
+
+class DimensionDeletion(DimensionAddition):
```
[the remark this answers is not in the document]
bouthilx: I'll rework it a bit.

src/orion/core/evc/adapters.py
```diff
@@ -0,0 +1,639 @@
...
+        adapted_trials = []
+
+        for trial in trials:
+            if apply_if_valid(self.param.name, trial, raise_if_not=False):
+                raise RuntimeError("Provided trial does not have a compatible configuration. "
+                                   "A dimension named '%s' was already present.\n %s" %
+                                   (self.param.name, trial))
+
+            adapted_trial = copy.deepcopy(trial)
+            adapted_trial.params.append(self.param)
```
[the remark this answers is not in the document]
bouthilx: Well, a `Trial.Param` instance is basically a `dimension.sample()` turned into a param. So, if we have a default value, `Trial.Param` is a very convenient and natural way to encapsulate such default value. Furthermore, for DimensionAddition and DimensionDeletion we need to add a `Trial.Param` object to each trials, so using such encapsulation for default value is very very convenient.

tests/unittests/core/test_utils.py
bouthilx: Bad me? I did the typo, no?

src/orion/core/evc/adapters.py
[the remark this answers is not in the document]
tsirif: I agree. This means that `default_value` should be a slotted attribute of `Trial.Param`?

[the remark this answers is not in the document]
tsirif: #88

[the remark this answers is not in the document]
bouthilx: I don't think it should. The default_value is encapsulated as the *value* of `Trial.Param`. You could have either
```python
Trial.Param(dimension.name, dimension.type, dimension.sample())
```
or 
```python
Trial.Param(dimension.name, dimension.type, dimension.default_value)
```

We don't need to know if it was sampled or if it is the default value. It simply *is* the value of the param.

src/orion/core/worker/trial.py
tsirif: #89

src/orion/core/evc/adapters.py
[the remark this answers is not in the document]
tsirif: Oh I see now. I didn't see it. :stuck_out_tongue_closed_eyes:

bouthilx:
@tsirif I'm glad you enjoyed reading it ;) Thanks for your support :stuck_out_tongue: 

I think I addressed everything you mentioned now.

DatCorno:
Can this be merged?

bouthilx:
@DatCorno @tsirif It's up to you. You both approve the PR in its current state?

tsirif:
I am waiting for corno's input. I approved this message.

DatCorno:
From what I can see everything's good.

PR #85, opened 2018-05-16
Feature/Vizualization

dendisuhubdy:
- [x] interface with networkx and matplotlib (for tex and image base graphs)
- [x] add functionality given experiment forking histories, output graph
- [x] add functionality given experiment forking histories, output tex-based graph
- [x] write extensive tests
- [ ] write cli interface
- [ ] fix correct MD5 hashes
- [ ] correct pylints, coverages and travis CI approval


dendisuhubdy:
@bouthilx I was thinking that NetworkX might be better to manipulate graphs (interface is nice here, code is clean https://github.com/networkx/networkx/blob/master/examples/graph/expected_degree_sequence.py) while pydot is a blob of code in one place (https://github.com/erocarrera/pydot/blob/master/pydot.py)

bouthilx:
So you would like to use something like [this](https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pydot.to_pydot.html#networkx.drawing.nx_pydot.to_pydot)?

Maybe using matplotlib with networkx would be better indeed ([example](https://networkx.github.io/documentation/stable/auto_examples/drawing/plot_weighted_graph.html#sphx-glr-auto-examples-drawing-plot-weighted-graph-py)). @nouiz @lamblin What was your experience for Theano graphs?

dendisuhubdy:
@bouthilx actually it was my first CCW supervised by Fred (https://github.com/dendisuhubdy/ccw1_trainstats/blob/master/tutorial/visdom/6.compgraphs.md)

src/orion/viz/evc/graph/graph.py
```diff
@@ -0,0 +1,223 @@
...
+            # its children connected with the current node
+            # is connected with an edge
+            self.graph.add_edge(node, node.next())
+            # we've built our tree
+        
+        '''
+        Was considering using the map, but giving the node
+        the corresponding edges might not be straightforward
+        '''
+        #self.evc_node.map(self.graph.add_node, self.evc_node)
```
dendisuhubdy: @bouthilx is this the right way to ue map? I used an iteration instead of this for now, I want to get it right

dendisuhubdy: Nevermind `https://github.com/mila-udem/orion/commit/525bb73ce92d3a40eebd8f3537b24f5c50dd41b0`

bouthilx: I think that would work.
```python
def add_node(node, children):
    self.graph.add_node(node)
    for child in children:
        self.graph.add_edge(node, child)

    return node, children

self.evc_node.map(add_node, self.evc_node.children)
```

src/orion/viz/evc/graph/graph.py
```diff
@@ -0,0 +1,218 @@
...
+        an EVC node is passed through.
+        '''
+        self.evc_node = evc_node
+        
+        '''
+        Create a networkx graph and allocate
+        it in memory
+        '''
+        self.graph = nx.Graph()
+
```
dendisuhubdy: Call `self.build_graph()` here

dendisuhubdy:
@bouthilx @DatCorno should I close this?

bouthilx:
Yes we can close the PR but please do not delete the branch. We will need it when you and @DatCorno integrates the EVC visualizations with the new CLI.

dendisuhubdy:
@bouthilx ok

PR #86, opened 2018-05-17
Extend Factory to find subclasses recursively

bouthilx:
Why:

If a subclass of OptimizationAlgorithm is extended, the meta-class
Factory will not find it's implementation. Let's say for instance that
we define a global class ScikitOptimizationAlgorithm which integrates
generalities for all different algorithms from scikit-optimize, then all
children of ScikitOptimizationAlgorithm would not be found by the
Factory unless they are specifically registered by entry points.

How:

Call recursively on cls.__subclasses__() and list them all.

### DIFF
--- a/src/orion/core/utils/__init__.py
+++ b/src/orion/core/utils/__init__.py
@@ -91,7 +91,16 @@ class Factory(ABCMeta):
                       entry_point.dist.project_name, entry_point.dist.version)
 
         # Get types visible from base module or package, but internal
-        cls.types = cls.__base__.__subclasses__()
+        def get_all_subclasses(parent):
+            """Get set of subclasses recursively"""
+            subclasses = set()
+            for subclass in parent.__subclasses__():
+                subclasses.add(subclass)
+                subclasses |= get_all_subclasses(subclass)
+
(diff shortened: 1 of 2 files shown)


src/orion/core/utils/__init__.py
```diff
@@ -91,7 +91,15 @@ def __init__(cls, names, bases, dictionary):
                       entry_point.dist.project_name, entry_point.dist.version)
 
         # Get types visible from base module or package, but internal
-        cls.types = cls.__base__.__subclasses__()
+        def get_all_subclasses(parent):
+            """Get subclasses recursively"""
+            subclasses = []
+            for subclass in parent.__subclasses__():
+                subclasses += [subclass] + get_all_subclasses(subclass)
```
tsirif: So I have a minor thing that there will be duplicates entries with the same class here. So we may use set here (classes are hashable), and a `list(get_all_subclasses(cls.__base__))`.

tsirif:
This needs to be rebased right?

bouthilx:
No why would it need to? I amended to commit to fix the type in test_utils.py.

PR #91, opened 2018-05-18
 Add ExperimentView and ReadOnlyDB wrapper

bouthilx:
Why:

With experiment version control there is the possibility of having many
experiments instantiated at the same time but only one should be
editable, the _main_ one. All the others should not be able to make
modifications in the database. This commit is an attempt at reducing
write errors by wrapping the experiment with a read-only wrapper which
defines which attributes can be accessed and which not.

Note:

It will still be possible to access the write methods of the experiment,
by explicitly fetching it with by accessing the sub attribute of the
wrapper (ex: `experiment_view._experiment.register_trials()`).
Nonetheless the current wrapper makes it less likely to do so by error.

How:

Create an empty class with a `valid_attributes`. list which defines which
attribute and method may be accessed. The method `.__getattr__` filters out
the request on the wrapper based on `valid_attributes`.

The property `is_done` is slightly modified in the view since
`experiment.is_done` would normally do a write on the database. If
`wrapper.is_done` returns True, it will not update the database like the
normal experiment would.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -495,13 +495,12 @@ class ExperimentView(object):
     def __init__(self, name):
         """Initialize viewed experiment object with primary key (:attr:`name`, :attr:`user`).
 
-        Try to find an entry in `Database` with such a key and config this object
-        from it import, if successful. Else, init with default/empty values and
-        insert new entry with this object's attributes in database.
+        Build an experiment from configuration found in `Database` with a key (name, user).
 
         .. note::
 
             A view is fully configured at initialiation. It cannot be reconfigured.
+            If no experiment is found for the key (name, user), a `ValueError` will be raised.
 
         :param name: Describe a configuration with a unique identifier per :attr:`user`.
         :type name: str
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -488,7 +488,7 @@ class ExperimentView(object):
     #                     Attributes
     valid_attributes = (["_id", "name", "refers", "metadata", "pool_size", "max_trials", "status"] +
                         # Properties
-                        ["space", "algorithms", "stats"] +
+                        ["space", "algorithms", "stats", "configuration"] +
                         # Methods
                         ["fetch_completed_trials"])
 
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -501,13 +501,19 @@ class ExperimentView(object):
 
         .. note::
 
-            A view is **note** initializable. It is meant to access trials and statistics from the
-            experiment..
+            A view is fully configured at initialiation. It cannot be reconfigured.
 
         :param name: Describe a configuration with a unique identifier per :attr:`user`.
         :type name: str
         """
         self._experiment = Experiment(name)
+
+        if self._experiment.status is None:
+            raise ValueError("No experiment with given name '%s' for user '%s' inside database, "
+                             "no view can be created." %
+                             (self._experiment.name, self._experiment.metadata['user']))
+
+        self._experiment.configure(self._experiment.configuration)
         self._experiment._db = ReadOnlyDB(self._experiment._db)
 
     def __getattr__(self, name):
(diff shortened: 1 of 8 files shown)


src/orion/core/worker/experiment.py
```diff
@@ -471,3 +471,66 @@ def _is_different_from(self, config):
...
+                        # Properties
+                        ["space", "algorithms", "stats"] +
+                        # Methods
+                        ["fetch_completed_trials"])
+
+    def __init__(self, name):
+        """Initialize viewed experiment object with primary key (:attr:`name`, :attr:`user`).
+
+        Try to find an entry in `Database` with such a key and config this object
+        from it import, if successful. Else, init with default/empty values and
```
DatCorno: I feel like there is a missing period here, it should probably read :
> Try to find an entry in `Database` with such a key and configure this object from it. Import if succesful.

src/orion/core/worker/experiment.py
```diff
@@ -471,3 +471,66 @@ def _is_different_from(self, config):
...
+        insert new entry with this object's attributes in database.
+
+        .. note::
+
+            A view is **note** initializable. It is meant to access trials and statistics from the
```
DatCorno: Should this read `not`? What exactly does this sentence mean? I feel like saying a class is not initializable inside its __init__ function is a bit misleading. We could maybe go with not mutable.

src/orion/core/worker/experiment.py
```diff
@@ -471,3 +471,66 @@ def _is_different_from(self, config):
...
+        :param name: Describe a configuration with a unique identifier per :attr:`user`.
+        :type name: str
+        """
+        self._experiment = Experiment(name)
+        self._experiment._db = ReadOnlyDB(self._experiment._db)
+
+    def __getattr__(self, name):
+        """Get attribute only if valid"""
+        if name not in self.valid_attributes:
+            raise AttributeError("Cannot access attribute %s on view-only experiments." % name)
```
DatCorno: We could provide a deepcopy of the attribute if it is not meant to be modified instead of throwing

src/orion/core/worker/experiment.py
bouthilx: Yes the sentence is terrible :sweat_smile: . I have commits in #92 for ExperimentView, I think I will import them in this PR. It will be better for the review.

src/orion/core/worker/experiment.py
bouthilx: Yes I fixed it in the other PR's commits. The ExperimentView will now be fully configured at initialization.

src/orion/core/worker/experiment.py
```diff
@@ -471,3 +471,66 @@ def _is_different_from(self, config):
...
+        .. seealso::
+            :attr:`orion.core.worker.experiment.Experiment.is_done`
+        """
+        query = dict(
+            experiment=self._id,
+            status='completed'
+            )
+        num_completed_trials = self._experiment._db.count('trials', query)
+
+        return num_completed_trials >= self.max_trials
```
DatCorno: Missing check `self._init_done and self.algorithms.is_done`, is this normal?

src/orion/core/worker/experiment.py
bouthilx: Well, the invalid attributes are meant for writing, this also includes writing **methods** of the viewed class. So, they should not be provided at all.

src/orion/core/worker/experiment.py
bouthilx: Yes it's because self.algorithms is not instantiated in the view since it is not fully configured. It's not true anymore in commits of the other PR.

src/orion/core/worker/experiment.py
DatCorno: Yes please :)

src/orion/core/worker/experiment.py
DatCorno: Good!

src/orion/core/worker/experiment.py
DatCorno: Makes sense.

src/orion/core/worker/experiment.py
DatCorno: Alright.

bouthilx:
close in favor #96 because of bad codecov report

PR #92, opened 2018-05-20
Feature/evc experiment builder

bouthilx:
Depends on #91

This is the largest of the pull requests for feature/evc on my side. The refactoring of the building process of experiments was required because otherwise the creation of the experiment version control tree would be more complicated and require additional code duplication. 

## (main commit) Refactor experiment build into ExperimentBuilder 

Why:

The instantiation of an `Experiment` is not a trivial process when the
user request an experiment with specific options. One can easily create
a new experiment with `ExperimentView('some_experiment_name')`, but the
configuration of a _writable_ experiment is less straighforward. This is
because there is many sources of configuration and they have a strict
hierarchy. This commit attempts to unify the process of building an
experiment into a single module/class.

How:

The builder can build an experiment only provided the command args. All
fetch of other source of configuration are executed internally. The
implementation is less efficient to favor simplicity. Yet the loss of
efficiency should not impact the global execution as creating an
experiment happens rarely and requires only a few query on the
database. This is also true in the context of experiment version control
trees.

The builder is built as a class such that the methods' efficiency could
be improved latter on by caching configuration fetches to avoid
duplicate fetches.

--------

### Remove metadata/datetime check on init

Why:

This check was (I guess) to make sure the configuration was partly
initialized by the experiment itself which is not yet fully configured.
I don't think such a security measure is necessary and it makes
configuration of experiments more tedious. All information we want is
from default_config, env_vars_config, db_config, cmdconfig and cmdargs.
Attributes set by the experiment at initialization (not configuration)
does not need and should not pollute this configuration. username and
datetime is now set inside ExperimentBuilder.fetch_full_config is not
already available.

Note:

What about adding those metadata as a metadata config inside the
hierarchy? Something like

default <  env_vars < metadata < db_config < cmdconfig < cmdargs

----

### Do no set experiment on pending by default 

Why:

If an experiment is broken, it should not be automatically resumed
simply by fetching the experiment from database and configuring it.
Also, if the experiment is done and the termination criterion does not
change, then it should not be set to pending automatically.

How:

At configuration time, test if status is None or status is not broken
and the termination criterion is False. If yes, then set status to
pending.

----

### Unify resolve_config.merges

Why:

There is many different fetches and merges on the configuration
dictionaries. They more or less all share the same logic, so they are
easily to combine together to avoid code duplication.

How:

A new generic function `merge_configs` is added to merge any kind of
dictionaries with a relation B overwrites A recursively.

----

###  Move and refactor resolve_config 

Why:

The functionalities of resolve_config while strongly related to `cli`
are mostly related to parsing of configuration. All builders are
currently implemented in `orion.core.io`, including
`orion.core.experiment_builder` which will make heavy use of
`resolve_config`. Furthermore, by refactoring the experiment building
process, most of the use of `resolve_config` will be moved from `cli`
to `orion.core.experiment_builder`.

How:

Move file `orion.core.cli.resolve_config` to `orion.core.io.resolve_config`.
Move class `OrionArgParser` with functions `get_basic_args_group` and
`get_user_args_group` to new submodule `orion.core.cli.base`.

All `cli` modules are modified accordingly.

### DIFF
--- a/src/orion/core/cli/__init__.py
+++ b/src/orion/core/cli/__init__.py
@@ -16,12 +16,6 @@ from orion.core.cli.base import OrionArgsParser
 
 log = logging.getLogger(__name__)
 
-CLI_DOC_HEADER = """
-orion:
-  Orion cli script for asynchronous distributed optimization
-
-"""
-
 
 def load_modules_parser(orion_parser):
     """Search through the `cli` folder for any module containing a `add_subparser` function"""
@@ -44,7 +38,7 @@ def main(argv=None):
     # Fetch experiment name, user's script path and command line arguments
     # Use `-h` option to show help
 
-    orion_parser = OrionArgsParser(CLI_DOC_HEADER)
+    orion_parser = OrionArgsParser()
 
     load_modules_parser(orion_parser)
 
--- a/src/orion/core/cli/__init__.py
+++ b/src/orion/core/cli/__init__.py
@@ -12,7 +12,7 @@
 import logging
 import os
 
-from orion.core.cli import resolve_config
+from orion.core.cli.base import OrionArgsParser
 
 log = logging.getLogger(__name__)
 
@@ -21,20 +21,6 @@ orion:
   Orion cli script for asynchronous distributed optimization
 
 """
-
-
-def main(argv=None):
-    """Entry point for `orion.core` functionality."""
-    # Fetch experiment name, user's script path and command line arguments
-    # Use `-h` option to show help
-
-    orion_parser = resolve_config.OrionArgsParser(CLI_DOC_HEADER)
-
-    load_modules_parser(orion_parser)
-
-    orion_parser.execute(argv)
-
-    return 0
 
 
 def load_modules_parser(orion_parser):
@@ -51,6 +37,20 @@ def load_modules_parser(orion_parser):
         if hasattr(module, 'add_subparser'):
             add_subparser = getattr(module, 'add_subparser')
             add_subparser(orion_parser.get_subparsers())
+
+
+def main(argv=None):
+    """Entry point for `orion.core` functionality."""
+    # Fetch experiment name, user's script path and command line arguments
+    # Use `-h` option to show help
+
+    orion_parser = OrionArgsParser(CLI_DOC_HEADER)
+
+    load_modules_parser(orion_parser)
+
+    orion_parser.execute(argv)
+
+    return 0
 
 
 if __name__ == "__main__":
--- a/src/orion/core/io/experiment_builder.py
+++ b/src/orion/core/io/experiment_builder.py
@@ -126,8 +126,8 @@ class ExperimentBuilder(object):
 
         return resolve_config.merge_configs(default_options, env_vars, cmdconfig, cmdargs)
 
-    def fetch_db_config(self, cmdargs):
-        """Get dictionary of options from all local sources (not from db)
+    def fetch_config_from_db(self, cmdargs):
+        """Get dictionary of options from experiment found in the database
 
         Note
         ----
@@ -161,12 +161,12 @@ class ExperimentBuilder(object):
         """
         default_options = self.fetch_default_options()
         env_vars = self.fetch_env_vars()
-        db_config = self.fetch_db_config(cmdargs)
+        config_from_db = self.fetch_config_from_db(cmdargs)
         cmdconfig = self.fetch_file_config(cmdargs)
         metadata = dict(metadata=self.fetch_metadata(cmdargs))
 
         exp_config = resolve_config.merge_configs(
-            default_options, env_vars, db_config, cmdconfig, cmdargs, metadata)
+            default_options, env_vars, config_from_db, cmdconfig, cmdargs, metadata)
 
         return exp_config
 
--- a/src/orion/core/io/experiment_builder.py
+++ b/src/orion/core/io/experiment_builder.py
@@ -1,0 +1,279 @@
+# -*- coding: utf-8 -*-
+# pylint:disable=protected-access
+"""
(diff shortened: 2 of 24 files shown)


DatCorno:
Guess I won't be sleeping tonight!

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,279 @@
...
+            configurations.
+        """
+        default_options = self.fetch_default_options()
+        env_vars = self.fetch_env_vars()
+        cmdconfig = self.fetch_file_config(cmdargs)
+
+        return resolve_config.merge_configs(default_options, env_vars, cmdconfig, cmdargs)
+
+    def fetch_db_config(self, cmdargs):
+        """Get dictionary of options from all local sources (not from db)
```
DatCorno: Either the name of the function is wrong or this comment is misleading, but when we say `fetch db config` we normally expect it to come from the database, which this comment indicates is not what is happening here.

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,279 @@
...
+        ----
+            This method builds an experiment view in the background to fetch the configuration from
+            the database.
+        """
+        default_options = self.fetch_default_options()
+        env_vars = self.fetch_env_vars()
+        db_config = self.fetch_db_config(cmdargs)
+        cmdconfig = self.fetch_file_config(cmdargs)
+
+        exp_config = resolve_config.merge_configs(
```
DatCorno: I really don't like duplicating all these calls to the fetching function. Couldn't we modify `fetch_local_config` with a boolean defaulted to False that indicates if we want to merge the database with the other ones? I feel like this would be cleaner.

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,279 @@
...
+
+            :class:`orion.core.worker.experiment.Experiment` for more information on the experiment
+            object.
+        """
+        full_config = self.fetch_full_config(cmdargs)
+
+        log.info(full_config)
+
+        # Pop out configuration concerning databases and resources
+        full_config.pop('database', None)
```
DatCorno: Useless call since it is already popped inside the call to build_from_config

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,279 @@
...
+        full_config.pop('resources', None)
```
DatCorno: Idem

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,279 @@
...
+        full_config.pop('status', None)
```
DatCorno: Idem

src/orion/core/io/resolve_config.py
```diff
@@ -282,22 +193,66 @@ def merge_orion_config(config, dbconfig, cmdconfig, cmdargs):
...
-    for k, v in cmdargs.items():
-        if v is not None:
-            if k == 'metadata':
-                for vk, vv in v.items():
-                    expconfig[k][vk] = vv
-            else:
-                expconfig[k] = v
-
-    return expconfig
+    return merge_configs(config, dbconfig, cmdconfig, cmdargs)
```
DatCorno: Why not just delete this function and replaced the calls to `merge_orion_config` by `merge_configs`?

src/orion/core/io/resolve_config.py
```diff
@@ -282,22 +193,66 @@ def merge_orion_config(config, dbconfig, cmdconfig, cmdargs):
...
+    value if it was not also a directory, or it will be merged following
+    `merge_configs(old_value, new_value)`.
+
+    .. warning:
+
+        Redefinition of subdictionaries may lead to confusing results because merges do not remove
+        data.
+
+        If for instance, we have {'a': {'b': 1, 'c': 2}} and we would like to update `'a'` such that
+        it only have `{'c': 3}`, it won't work with {'a': {'c': 3}}.
```
DatCorno: Typo : only `has`

src/orion/core/cli/__init__.py
```diff
@@ -53,5 +39,19 @@ def load_modules_parser(orion_parser):
             add_subparser(orion_parser.get_subparsers())
 
 
+def main(argv=None):
+    """Entry point for `orion.core` functionality."""
+    # Fetch experiment name, user's script path and command line arguments
+    # Use `-h` option to show help
+
+    orion_parser = OrionArgsParser(CLI_DOC_HEADER)
```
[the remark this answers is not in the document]
bouthilx: Yes I agree.

src/orion/core/io/experiment_builder.py
bouthilx: Oops, copy-pasted did not edit from method above.

src/orion/core/io/experiment_builder.py
bouthilx: I wanted to make the methods as obvious as possible since the process is confusing. Maybe you are right and I went a bit too far, too verbose. I would agree to remove `fetch_local_config(self, cmdargs)` and turn `fetch_full_config(self, cmdargs)` into `fetch_config(self, cmdargs, use_db=True)`. I would make `True` by default the optional attribute `use_db` since you would expect it to hit the db with a syntax like `ExperimentBuilder().fetch_config(cmdargs)`. And `ExperimentBuilder().fetch_config(cmdargs, use_db=False)` is pretty clear.

src/orion/core/io/resolve_config.py
bouthilx: I was planning to do it in the commit adding ExperimentBuilder. I forgot.

bouthilx:
@DatCorno Thanks for the review! I will finish the last PR for experiment nodes such that everything for EVC is done and then I will fix this PR and the others.

PR #93, opened 2018-05-20
Feature/dynamic/Refactor Consumer to be extendable easier and handle cases

tsirif:


tests/functional/consumer/test_consumer.py
```diff
@@ -150,19 +151,24 @@ def test_consumer_killed_with_sigint(database, monkeypatch, capfd):
...
-    capfd.readouterr()  # Suppress fd level 1 & 2
 
+    time.sleep(3)
     exp = list(database.experiments.find({'name': EXP_NAME}))
     assert len(exp) == 1
     exp = exp[0]
     assert '_id' in exp
     exp_id = exp['_id']
 
+    cc = 0
```
bouthilx: This is why I would prefer to avoid the use of `Popen` in unit-tests. We could mock the methods to automatically trigger signals at the right time rather than fine-tuning sleep intervals or listening to the database to detect an event.

src/orion/core/worker/experiment.py
```diff
@@ -202,7 +202,7 @@ def reserve_trial(self, score_handle=None):
 
         return selected_trial
 
-    def push_completed_trial(self, trial):
+    def push_completed_trial(self, trial, status='completed'):
```
```python
experiment.push_completed_trial(trial, 'reserved')
```
:thinking: 
If the behavior is generic, the name of the method should be generic too.

src/orion/core/worker/experiment.py
```diff
@@ -256,6 +256,11 @@ def is_done(self):
 
         .. note:: To be used as a terminating condition in a ``Worker``.
         """
+        self.status = self._db.read('experiments', {'_id': self._id},
```
bouthilx: That's a weird side-effect in my opinion. If we want to ensure `status` to be up-to-date then maybe the property method should always call the db.

src/orion/core/worker/experiment.py
```diff
@@ -275,6 +280,11 @@ def is_broken(self):
 
         .. note:: To be used as a terminating condition for a failed ``Worker``.
         """
+        self.status = self._db.read('experiments', {'_id': self._id},
```
bouthilx: idem

src/orion/core/worker/consumer.py
```diff
@@ -150,60 +148,66 @@ def consume(self, trial):
...
-            if returncode == 0 or (returncode is None and trial.results):
-                log.debug("### Register successfully evaluated %s.", trial)
-                self.experiment.push_completed_trial(trial)
+        finally:
+            if returncode == 0 or (returncode is None and self.current_trial.results):
+                log.debug("### Update successfully evaluated %s.", self.current_trial)
+                self.experiment.push_completed_trial(self.current_trial)
             elif returncode is not None:
-                update_trial_with_status(trial, 'broken')
+                self.experiment.push_completed_trial(self.current_trial, 'broken')
```
bouthilx: :thinking:

src/orion/core/worker/consumer.py
```diff
@@ -150,60 +148,66 @@ def consume(self, trial):
...
-                                                   delete=False)
-        results_file.close()
-        log.debug("## New temp results file: %s", results_file.name)
-
-        log.debug("## Building command line argument and configuration for trial.")
-        cmd_args = self.template_builder.build_to(config_file.name, trial, self.experiment)
-
-        command = [self.script_path] + cmd_args
-        return self.interact_with_script(command, results_file)
+                self.experiment.push_completed_trial(self.current_trial, new_status)
```
bouthilx: :thinking:

src/orion/core/worker/consumer.py
```diff
@@ -57,27 +65,90 @@ def __init__(self, experiment):
...
+            raise
+
+        finally:
+            trial.results = self.latest_results
+
+            if returncode == 0 or (returncode is None and trial.results):
+                log.debug("### Register successfully evaluated %s.", trial)
+                self.experiment.push_completed_trial(trial)
+            elif returncode is not None:
+                update_trial_with_status(trial, 'broken')
```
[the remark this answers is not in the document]
tsirif: Yes, I agree. We should remove `push_completed_trial` in `Experiment` in favor of `finalize_trial(trial, status)`. The new method would also complete the tuple (start/resume time, finish time) and push it as well, according to what we have discussed.

src/orion/core/worker/experiment.py
tsirif: I agree, but I am not courageous enough to break this kind of API. How about `finalize_trial` that I have suggested above?

src/orion/core/worker/consumer.py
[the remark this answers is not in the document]
bouthilx: `finalize_trial` or `update_trial`? The former sounds like the trial is not expected to change afterwards. This would be confusing: `finalize_trial(trial, 'suspended')`

src/orion/core/worker/experiment.py
bouthilx: See my new comment above. :slightly_smiling_face:

src/orion/core/worker/experiment.py
tsirif: I consider this fixed by #101

bouthilx: Indeed! :)

tsirif:
@bouthilx So what shall we do about this?

1. Rename `push_completed_trials`?
2. Future: config shall be able to change exit codes??
3. Adapt code to reflect #101?

bouthilx:
Yep, and we could discuss about the use of Popen in unit-tests.

bouthilx:
https://github.com/mila-udem/orion/pull/93#discussion_r191035199

tests/functional/consumer/test_consumer.py
tsirif: But then again, who would catch the signals? I believe we need 2 processes for these kinds of tests. How would it be made synchronous, I don't know.

bouthilx: python threading could do the job I guess. The advantage is that we have access to code run inside the thread so that we can mock objects inside of it. Popen totally isolate the executed process and we can't mock anything inside of it.

tsirif:
I will refactor the tests, rework the method, rebase accordingly on top of #101

tsirif:
@bouthilx So `threading` does not support signaling: https://docs.python.org/3/library/threading.html

Perhaps a solution based on the debug output and checking subprocess' stdout with `Popen.communicate` until something desired appears can be better??

something like:
1. use communicate for standard timeout, define a ΔT
2. capture stdout for this ΔT
3. check whether desired substring exists in the captured output
4. perform the kill, or go back to 1

??

bouthilx:
@tsirif It would work with `multiprocessing` then. The advantage over Popen is that you can define mocks before forking the process, whereas Popen would reimport everything in isolation.

bouthilx:
I close this PR due to staleness. I extended further the handling of interruption/suspension in a local branch. I'll try to make a PR soon.

PR #94, opened 2018-05-21
Experiment nodes

bouthilx:
Depends on #84 #91 #92 

Note:

I'm not confident about the EVCBuilder. I consider refactoring the lazy-initialisation of `ExperimentNode` such that `Experiment` can automatically build and instantiate the EVC tree when trying to access it. I wanted to build the tree externally to simplify the process of trimming (when user specify subroot and subleafs), which is not trivial. However I believe the same helper functions used to trim could be used inside the node methods, so that trimming occurs dynamically rather than at build time.

As I said, I'm not sure yet. I'd be happy to discuss it. Obviously there is no time for that because the deadline, so I propose we merge it in current state (all other objections aside) and make such refactoring  next week.

###

## Add EVCBuilder

Why:

Building the EVC tree can by tricky if we support trimming for
optimization on smaller EVC subtree. For example user can define a root
and some leafs that should be the extremums of the tree. Those can be
different than the actual root and leafs of the global EVC tree, making
the trimmed version a small subset of the global version.

How:

Wraps ExperimentBuilder fonctionalities to parse command line.

The process of building the tree is the following:

1. Fetch all documents with a given root_id (root of global tree)
2. Build the global tree based on those shallow documents
   (only minimal info fetched from db)
3. Trim the tree based on root name and leaf names specified by user
4. Turn the shallow nodes into experiment nodes (lazy initialization)

The experiment is then connected to the EVC tree, before being returned.

## Add EVC node to `Experiment`

Why:

For modularity, an `Experiment` does not need to be part of EVC and can
be standalone.

How:

EVC trees can be built using _shallow_ definitions (only pointing to
name of experiments). Once the `item` of the shallow nodes are accessed,
and experiment view is created and is connected to the EVC node. Once
the connection is made, the experiment now have access to trials from
the entire tree. Those trials can be fetched using method
`Experiment.fetch_trials_tree`, a recursive version of
`Experiment.fetch_trials`. Note that both return a list of trials, not a
tree of trials.

Note:

Experiment status and statistics are computed only based on its own
trials, not those of the tree.

##  Update `Experiment` based on new refers definition

Why:

The definition of `refers` is now:

 - `root_id`: Experiment id of the root experiment of the EVC tree
    This helps improve efficiency on db fetches since we can fetch a
    tree simply by specifying the root_id. It is also simple to keep
    coherent compared to a double-linked tree with (parent, children)
    links.
 - `parent_id`: Parent id of the current experiment
 - `adapter`: CompositeAdaptor definition to make trials from parent
    compatible with the current one and vice-versa.

How:

On creation of new experiments, `refers` is updated after the experiment
is added to database such that `refers[root_id]` can point to itself.

Note:

Because `refers` contain sub-dictionaries which are edited after
configuration (adapters instantiation), we need to use deep copies of
the configuration when configuring.

## Add experiment node mechanisms

Why:

The tree of experiments have common use cases like fetching trials and
lazy experiment instantiation. Defining a child `ExperimentNode` is
convenient to group such operations.

How:

ExperimentNode carries an experiment in `item`. If node is created using
only the experiment's name, it will instantiate the experiment only upon
access to `item`. `ExperimentNode.fetch_trials` returns a list of trials
fetched recursively in the EVC tree.

Helper functions to apply operations on the tree, not the node
specifically, are kept outside `ExperimentNode`. Results of helper
functions on the EVC are returned in the form of a tree to keep the
original structure (can be useful when analysing an EVC tree by keeping
experiment results separated in each node).

### DIFF
--- a/src/orion/client/manual.py
+++ b/src/orion/client/manual.py
@@ -9,10 +9,8 @@
       and link them with a particular existing experiment.
 
 """
-from orion.core.cli import resolve_config
-from orion.core.io.database import Database
-from orion.core.utils import (format_trials, SingletonError,)
-from orion.core.worker.experiment import Experiment
+from orion.core.io.experiment_builder import ExperimentBuilder
+from orion.core.utils import format_trials
 
 
 def insert_trials(experiment_name, points, cmdconfig=None, raise_exc=True):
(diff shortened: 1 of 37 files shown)


bouthilx:
Now in #114

PR #96, opened 2018-05-25
Add ExperimentView and ReadOnlyDB wrapper

bouthilx:
Why:

With experiment version control there is the possibility of having many
experiments instantiated at the same time but only one should be
editable, the main one. All the others should not be able to make
modifications in the database. This commit is an attempt at reducing
write errors by wrapping the experiment with a read-only wrapper which
defines which attributes can be accessed and which not.

Note:

It will still be possible to access the write methods of the experiment,
by explicitly fetching it with by accessing the sub attribute of the
wrapper (ex: experiment_view._experiment.register_trials()).
Nonetheless the current wrapper makes it less likely to do so by error.

How:

Create an empty class with a valid_attributes. list which defines which
attribute and method may be accessed. The method .__getattr__ filters out
the request on the wrapper based on valid_attributes.

The property is_done is slightly modified in the view since
experiment.is_done would normally do a write on the database. If
wrapper.is_done returns True, it will not update the database like the
normal experiment would.

### DIFF
--- a/src/orion/core/io/database/__init__.py
+++ b/src/orion/core/io/database/__init__.py
@@ -209,6 +209,36 @@ class AbstractDB(object, metaclass=AbstractSingletonType):
         pass
 
 
+# pylint: disable=too-few-public-methods
+class ReadOnlyDB(object):
+    """Read-only view on a database.
+
+    .. seealso::
+
+        :py:class:`orion.core.io.database.AbstractDB`
+    """
+
+    __slots__ = ('_database', )
+
+    #                     Attributes
+    valid_attributes = (["host", "name", "port", "username", "password"] +
+                        # Properties
+                        ["is_connected"] +
+                        # Methods
+                        ["initiate_connection", "close_connection", "read", "count"])
+
+    def __init__(self, database):
+        """Init method, see attributes of :class:`AbstractDB`."""
+        self._database = database
+
+    def __getattr__(self, attr):
+        """Get attribute only if valid"""
+        if attr not in self.valid_attributes:
+            raise AttributeError("Cannot access attribute %s on view-only experiments." % attr)
+
+        return getattr(self._database, attr)
+
+
 class DatabaseError(RuntimeError):
     """Exception type used to delegate responsibility from any database
     implementation's own Exception types.
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -488,13 +488,12 @@ class ExperimentView(object):
     def __init__(self, name):
         """Initialize viewed experiment object with primary key (:attr:`name`, :attr:`user`).
 
-        Try to find an entry in `Database` with such a key and config this object
-        from it import, if successful. Else, init with default/empty values and
-        insert new entry with this object's attributes in database.
+        Build an experiment from configuration found in `Database` with a key (name, user).
 
         .. note::
 
             A view is fully configured at initialiation. It cannot be reconfigured.
+            If no experiment is found for the key (name, user), a `ValueError` will be raised.
 
         :param name: Describe a configuration with a unique identifier per :attr:`user`.
         :type name: str
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -531,4 +531,4 @@ class ExperimentView(object):
             )
         num_completed_trials = self._experiment._db.count('trials', query)
 
-        return num_completed_trials >= self.max_trials
+        return num_completed_trials >= self.max_trials or self.algorithms.is_done
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -479,9 +479,9 @@ class ExperimentView(object):
     __slots__ = ('_experiment', )
 
     #                     Attributes
-    valid_attributes = (["_id", "name", "refers", "metadata", "pool_size", "max_trials", "status"] +
+    valid_attributes = (["_id", "name", "refers", "metadata", "pool_size", "max_trials"] +
                         # Properties
-                        ["space", "algorithms", "stats", "configuration"] +
+                        ["id", "is_done", "space", "algorithms", "stats", "configuration"] +
                         # Methods
                         ["fetch_completed_trials"])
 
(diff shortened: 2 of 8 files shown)


bouthilx:
@tsirif @DatCorno  Can you please take a final look at the PR and merge if you agree with it. :)

tests/unittests/core/conftest.py
```diff
@@ -114,3 +114,23 @@ def hacked_exp(with_user_dendi, random_dt, clean_db):
     exp = Experiment('supernaedo2')
     exp._id = 'supernaedo2'  # white box hack
     return exp
+
+
+@pytest.fixture()
+def trial_id_substitution(with_user_tsirif, random_dt, clean_db):
```
tsirif: omg yes.

tsirif: hacked_exp is deprecated now

tsirif: or it could be considered the same but from the opposite approach.

tests/unittests/core/test_experiment.py
```diff
@@ -448,7 +448,6 @@ def test_algorithm_config_with_just_a_string(self, exp_config):
         new_config['algorithms'] = 'dumbalgo'
         exp = Experiment('supernaedo3')
         exp.configure(new_config)
-        new_config['status'] = 'pending'
```
tsirif: When should the status change effectively to 'pending'?

src/orion/core/worker/experiment.py
```diff
@@ -313,7 +313,11 @@ def configure(self, config):
         experiment._instantiate_config(self.configuration)
         experiment._instantiate_config(config)
         experiment._init_done = True
-        experiment.status = 'pending'
+
+        # If experiment is new or non-branching parameters have been changed such that the status is
+        # not 'done' anymore
+        if experiment.status is None or (experiment.status != 'broken' and not experiment.is_done):
```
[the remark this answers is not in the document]
tsirif: Is there a specific reason you wanted to have this in `configure`, instead of having it done in `workon`? I find it more explicit this way. Could you explain the status transition table??

src/orion/core/worker/experiment.py
```diff
@@ -471,3 +475,71 @@ def _is_different_from(self, config):
...
+        .. seealso::
+            :attr:`orion.core.worker.experiment.Experiment.is_done`
+        """
+        query = dict(
+            experiment=self._id,
+            status='completed'
+            )
+        num_completed_trials = self._experiment._db.count('trials', query)
+
+        return num_completed_trials >= self.max_trials or self.algorithms.is_done
```
DatCorno: Could we move this to a private function inside Experiment so we don't have to maintain to check every time?

tests/unittests/core/database_test.py
```diff
@@ -46,3 +46,62 @@ def test_instatiation_and_singleton(self):
...
         assert 'singleton' in str(exc_info.value)
+
+
+@pytest.mark.usefixtures("null_db_instances")
+class TestReadOnlyDatabase(object):
+    """Test coherence of read-only database and its wrapped database."""
+
+    def test_valid_attributes(self):
+        """Test attributes are coherent from view and wrapped database."""
+        database = Database(of_type='MongoDB', name='orion_test',
```
DatCorno: Should we fixture this?

src/orion/core/worker/experiment.py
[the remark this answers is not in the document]
bouthilx: Currently the experiment was set to 'pending' by default. I restricted the change of status so that it can only become 'pending' if it is new, or if it is not broken and not completed yet.

I would totally agree to externalise the change of status. I believe this one to be the lesser evil considering the status is changed inside `Experiment.configure`.

tests/unittests/core/conftest.py
bouthilx: I don't know if it's worth removing unit-tests with hacked_exp, but yes this fixture kind of make hacked_exp deprecate. There is a similar fixture but for experiment cross references in #94 .

tests/unittests/core/test_experiment.py
bouthilx: When the maximum number of trials is higher than current number of trials and not algo.is_done. One or the other condition is enough to force 'completed'.

src/orion/core/io/database/__init__.py
```diff
@@ -209,6 +209,36 @@ def remove(self, collection_name, query):
...
+    """Read-only view on a database.
+
+    .. seealso::
+
+        :py:class:`orion.core.io.database.AbstractDB`
+    """
+
+    __slots__ = ('_database', )
+
+    #                     Attributes
```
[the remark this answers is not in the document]
bouthilx: Possibly. It was to align with ```["host", etc]```, but the `#` sign cannot be indented differently. What do you think would be better?

src/orion/core/worker/experiment.py
```diff
@@ -471,3 +475,71 @@ def _is_different_from(self, config):
...
+
+    .. seealso::
+
+        :py:class:`orion.core.worker.experiment.Experiment` for writable experiments.
+
+    """
+
+    __slots__ = ('_experiment', )
+
+    #                     Attributes
```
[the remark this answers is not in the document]
bouthilx: idem :)

src/orion/core/worker/experiment.py
bouthilx: I'm not sure how we could do this _naturally_. I would tend to remove completely `is_done` reimplementation from View and rather catch writing attempts to avoid any changes in the wrapped experiment. This way we could reuse Experiment.is_done and avoid duplicating the methods.

tests/unittests/core/database_test.py
bouthilx: Yes indeed!

tsirif:
@bouthilx This seems to contain the changes from #101, so that `is_done` or `is_broken` should they be read-only, they don't need to be wrapped. I think that this PR is ready right?

bouthilx:
@tsirif Yes indeed, the properties `is_done` and `is_broken` must be specified in the `valid_attributes` of the view though. 

I believe it is ready yes.

@DatCorno ?

bouthilx:
@tsirif Merge?

PR #99, opened 2018-05-28
Experiment branching

DatCorno:
This PR depends en #94

# Summary of the issue
When an experiment is created (be it via `init_only` or `hunt`) under a given name that is already present inside the database, the hyperparameters space is checked to make sure both experiments are the same. This check can fail for multiple reasons : 

- A dimension was removed (or forgotten) from the space
- A dimension was added to the space
- A dimension had changed its current prior and is therefore different

This situation triggers an **Experiment Branch**. This branching process serves as the creation of a new experiment based on the one specified by the user, so that it will become its child (regarding the EVC tree.) Each different dimensions will result in a conflict between the two spaces. These conflicts can be solve using one of the two following methods :

- The branching prompt

These two methods will now be described below.

## The branching prompt
If the user is unfamiliar with the branching process or uses the automatic solver to solve only a subset of the conflicts, the branching prompt will  be launch.
This branching prompt serves as a tool to help the user manually solve the conflicts between the two spaces. By using the commands available to them, they can :

- `add` a given dimension. Only new or changed (with regard to their prior) dimensions may be added. They will now exist in the new experiment space.
- `remove` a given dimension. Only missing dimension may be remove. These dimensions will no longer be inside the experiment space.
- `rename` a dimension to another. If the name of a dimension needs to change for a reason or another, this command will let the user do it. Only a missing dimension may be rename. It may also only be renamed to a new dimension.
- `name` the current experiment. This will define the new name of the experiment inside the database
- `commit` the conflicts they have solved. If their are not completely solved, this will bring the user back to the prompt.
- `abort` and exit orion.

## The automatic solver
By using different markers when declaring their dimensions using the command line or the configuration file, the user can indicate how to automatically solve the conflicts. These markers are as follow, and relate to the commands explained above : 

- `~+`. The tilde and addition sign will add the current dimension. Ex. : `-x~+normal(0,1)`
- `~-`. The tilde and difference operator will remove the current dimension. Ex. : `-x~-'
- `~>`. The tilde and the greater than operator will rename a dimension from the expression to the left to the expression on the right. Ex. : `-x~>y`
- `-b XXX`, `--branch XXX` will rename the current experiment to `XXX`.

# Changes to SpaceBuilder
The implementation of the branching process required some changes to the `SpaceBuilder` class for it to handle the fact that new symbols need to be parsed from the command line or the config file.
To this end, some utilities functions were added. The regex parsing the arguments were adapted to detect markers. A dictionary mapping namespaces from the configuration file to their expressions was added and precautions were taken to only allow the creation of valid dimensions (we do not want to build `-x~>y` for example).

# The Experiment Branch Builder
The **Experiment Branch Builder** (**EBB**) is used to obtain and solve the conflicts. The **EBB** maintains a reference to the original configs object (from the original experiment and the conflicting one) and deep copies of them. The conflicting config reference will be modified to reflect the new name of the experiment and is used inside the `_branch_config` function of the `Experiment` class.
The copies are extended with the expressions from the configuration file and therefore need to leave the original objects intact.
A list of `Conflict` objects is created. These objects store multiple information regarding each conflict: its status, the dimension it is affecting, the expression of that dimension and, finally, whether or not it is solved.
These conflicts are manipulated through the `add_dimension`, `remove_dimension` and `rename_dimension` functions.
Once all the conflicts are solved, the **EBB** generates a `CompositeAdapter` from each of them and returns it to be attributed to the experiment.

### DIFF
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -121,12 +121,14 @@ class Dimension(object):
         return (self.name, self.shape, self.type, self.default_value, self.prior)
 
     def __eq__(self, other):
+        """Return True if other is the same dimension as self"""
         if not isinstance(other, Dimension):
             return False
 
         return self.__hashable_members() == other.__hashable_members()
 
     def __hash__(self):
+        """Return the hash of the hashable members"""
         return hash(self.__hashable_members())
 
     def sample(self, n_samples=1, seed=None):
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -120,6 +120,7 @@ class Dimension(object):
     def __hashable_members(self):
         return (self.name, self.shape, self.type, self.default_value, self.prior)
 
+    # pylint:disable=protected-access
     def __eq__(self, other):
         """Return True if other is the same dimension as self"""
         if not isinstance(other, Dimension):
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -117,6 +117,18 @@ class Dimension(object):
 
         self._default_value = default_value
 
+    def __hashable_members(self):
+        return (self.name, self.shape, self.type, self.default_value, self.prior)
+
+    def __eq__(self, other):
+        if not isinstance(other, Dimension):
+            return False
+
+        return self.__hashable_members() == other.__hashable_members()
+
+    def __hash__(self):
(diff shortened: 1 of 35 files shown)


bouthilx:
Now in #99

PR #100, opened 2018-05-31
Add GenericConverter for parsing unknown config file types

tsirif:
I have tried to minimize changes in `SpaceBuilder` itself, instead I moved the burden of implementation completely to a separate subclass of `BaseConverter`, which is called `GenericConverter`.

1. `GenericConverter` parses for Python regex pattern which allow `<namespace with '/'>~<dimension expression>` for casual building, as well as for the special markers needed for not-interactive experiment branching. Its outputs and expected inputs are exactly in the form that the other two existing converters, which wrap JSON and YAML. This means that the output of the `read` method is a nested dictionary, whose nodes corresponds to namespace levels and leaves to expressions. Correspondingly in the input of `generate` method, the same structure is expected but instead of expression, there are concrete values to be substituted to the template document inside `GenericConverter`.

2. Since I reuse `nesteddict` class, I moved it to `orion.core.utils`.

3. I changed a little bit the interface of `infer_converter_from_file_type` to have 2 optional keyword argument. The first argument is the regex to be used by the `GenericConverter`, which has a default (currently this is used), and the second argument is a prefix (currently expecting this to be `orion~`) to be substituted in the place of the namespace declaration. The second step is needed, in order to comply with the expected I/O of the converter interface. I remind that for JSON or YAML, the declaration is like `orion~<dimension expression>` which is also reflected to how `SpaceBuilder` build from configuration files.


tsirif:
@bouthilx @DatCorno Happy 100 PRs and issues!!!

src/orion/core/io/convert.py
```diff
@@ -122,6 +125,126 @@ def generate(self, filepath, data):
...
+            for name in namespaces:
+                if namespaces.count(name) != 1:
+                    self._raise_conflict(filepath, name)
+
+        # Create template using each namespace as format key,
+        # exactly as provided by the user
+        subst = self.re_module.sub(r'{', r'{{', self.template)
+        subst = self.re_module.sub(r'}', r'}}', subst)
+        substituted, num_subs = self.regex.subn(r'{\1!s}', subst)
+        assert len(ret) == num_subs, "original: {}\n, regex:{}".format(self.template, self.regex)
```
bouthilx: Can we replace with a if; raise? I would limit assertions to unit-tests

src/orion/core/io/convert.py
```diff
@@ -122,6 +125,126 @@ def generate(self, filepath, data):
...
+        # Wrap it in style of what the rest of `Converter`s return
+        ret_nested = nesteddict()
+        for namespace, expression in ret.items():
+            keys = namespace.split('/')
+            if not keys[0]:  # It means that user wrote a namespace starting from '/'
+                keys = keys[1:]  # Safe because of the regex pattern
+                self.has_leading[namespace[1:]] = '/'
+
+            stuff = ret_nested
+            for i, key in enumerate(keys[:-1]):
```
bouthilx: I would be in favor of replacing this with a recursive function, but I won't block this PR for this.

tests/unittests/core/bad_config1.txt
```diff
@@ -0,0 +1,31 @@
+lalala~uniform(1, 3, shape=(100, 3))
```
bouthilx: Personally would rename the files such that their name convey what the problem is.

src/orion/core/io/convert.py
tsirif: I am not sure, that is what Python's `assert` originally was intended to be used for right?

tsirif:
@bouthilx I don't believe we should offer any help to people that try to mess Oríon with a Python script they use to configure their code. It is bad practice, and in any case if they pass with `--config=my_python_config_module` then they should take that string and import it with `importlib` or something. :no_entry_sign:

bouthilx:
I'm not sure if we are talking about the same thing. I'm not talking about config file for Oríon, but only the config file for the experiment, the one which contains only definitions for the hyper-parameter space.

tsirif:
Yes, about the same. A user would write `orion ... ./path_to/their_script.py --config~./my_templated/module_path.py` and orion would call `./path_to/their_script.py --config=./my_templated/module_path.py`, so a user should take this path as a string and import it, which will be super messy. I am against trying to support scripts as configuration files.

src/orion/core/io/convert.py
bouthilx: As far as I understand assert is meant for debugging on the developer's side.

https://docs.python.org/3.5/reference/simple_stmts.html#assert

In this case it would be used by the user to debug it's configuration file definition according to parsing errors.

bouthilx:
I understand your point about "messyness", but it's a common practice and not supporting it could be harmful. Anyway, we already support it somehow, the only issue was about the quotes.

Let's leave it that way and we'll see if we ever get complains about the quote thing.

src/orion/core/io/convert.py
tsirif: Oh yes this assertion was meant to be for the developer. If this doesn't work, something was wrong with parsing code. It is not meant to tell the user that they did anything wrong, but to potentially catch a faulty runtime behavior cause by a case that I, the developer, did not expect and hence did not write a test for.

I prefer to remove it completely, than to substitute it with another way to have a debugging line.

bouthilx: I don't understand your last sentence: 
> I prefer to remove it completely, than to substitute it with another way to have a debugging line.

Then I would be fine with an assertion, but a message should tell the developer what kind of problem is there.

bouthilx:
@tsirif Ping :)

PR #101, opened 2018-06-01
Remove status attribute from Experiment

bouthilx:
Why:

The status is not useful since properties is_done and
is_broken (coming soon) can be used to verify the state
of the experiment. As the status can be inferred, and experiments
are not very likely to be queried based on status (in opposition
to trials which are often queried based on status), there is
no reason to maintain a status attribute.

How:

Replace tests on status (which all had the function of testing if an
experiment is configured or found in DB) with test on _id.

### DIFF
--- a/src/orion/client/manual.py
+++ b/src/orion/client/manual.py
@@ -49,7 +49,7 @@ def insert_trials(experiment_name, points, cmdconfig=None, raise_exc=True):
 
     experiment = Experiment(experiment_name)
     # Configuration is completely taken from the database
-    if experiment._id is None:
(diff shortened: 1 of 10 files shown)


bouthilx:
@tsirif @DatCorno As discussed today.

tsirif:
Ok so let's merge this quickly, so that I rebase on top and change `is_broken` as well.

So now this means that `is_done` and `is_broken` are now read-only and the methods can be removed from the the wrappers right?

I will approve once I grep for `status` to cross check that there is no stuff left. Unless you have already done that.

PR #102, opened 2018-06-04
Feature/evc experiment builder

bouthilx:
Depends on #96

This is the largest of the pull requests for feature/evc on my side. The refactoring of the building process of experiments was required because otherwise the creation of the experiment version control tree would be more complicated and require additional code duplication. 

## (main commit) Refactor experiment build into ExperimentBuilder 

Why:

The instantiation of an `Experiment` is not a trivial process when the
user request an experiment with specific options. One can easily create
a new experiment with `ExperimentView('some_experiment_name')`, but the
configuration of a _writable_ experiment is less straighforward. This is
because there is many sources of configuration and they have a strict
hierarchy. This commit attempts to unify the process of building an
experiment into a single module/class.

How:

The builder can build an experiment only provided the command args. All
fetch of other source of configuration are executed internally. The
implementation is less efficient to favor simplicity. Yet the loss of
efficiency should not impact the global execution as creating an
experiment happens rarely and requires only a few query on the
database. This is also true in the context of experiment version control
trees.

The builder is built as a class such that the methods' efficiency could
be improved latter on by caching configuration fetches to avoid
duplicate fetches.

--------

### Remove metadata/datetime check on init

Why:

This check was (I guess) to make sure the configuration was partly
initialized by the experiment itself which is not yet fully configured.
I don't think such a security measure is necessary and it makes
configuration of experiments more tedious. All information we want is
from default_config, env_vars_config, db_config, cmdconfig and cmdargs.
Attributes set by the experiment at initialization (not configuration)
does not need and should not pollute this configuration. username and
datetime is now set inside ExperimentBuilder.fetch_full_config is not
already available.

Note:

What about adding those metadata as a metadata config inside the
hierarchy? Something like

default <  env_vars < metadata < db_config < cmdconfig < cmdargs

----

### Do no set experiment on pending by default 

Why:

If an experiment is broken, it should not be automatically resumed
simply by fetching the experiment from database and configuring it.
Also, if the experiment is done and the termination criterion does not
change, then it should not be set to pending automatically.

How:

At configuration time, test if status is None or status is not broken
and the termination criterion is False. If yes, then set status to
pending.

----

### Unify resolve_config.merges

Why:

There is many different fetches and merges on the configuration
dictionaries. They more or less all share the same logic, so they are
easily to combine together to avoid code duplication.

How:

A new generic function `merge_configs` is added to merge any kind of
dictionaries with a relation B overwrites A recursively.

----

###  Move and refactor resolve_config 

Why:

The functionalities of resolve_config while strongly related to `cli`
are mostly related to parsing of configuration. All builders are
currently implemented in `orion.core.io`, including
`orion.core.experiment_builder` which will make heavy use of
`resolve_config`. Furthermore, by refactoring the experiment building
process, most of the use of `resolve_config` will be moved from `cli`
to `orion.core.experiment_builder`.

How:

Move file `orion.core.cli.resolve_config` to `orion.core.io.resolve_config`.
Move class `OrionArgParser` with functions `get_basic_args_group` and
`get_user_args_group` to new submodule `orion.core.cli.base`.

All `cli` modules are modified accordingly.

### DIFF
--- a/src/orion/core/io/experiment_builder.py
+++ b/src/orion/core/io/experiment_builder.py
@@ -253,7 +253,6 @@ class ExperimentBuilder(object):
         # Pop out configuration concerning databases and resources
         config.pop('database', None)
         config.pop('resources', None)
-        config.pop('status', None)
 
         experiment = Experiment(config['name'])
 
--- a/src/orion/core/io/experiment_builder.py
+++ b/src/orion/core/io/experiment_builder.py
@@ -111,20 +111,6 @@ class ExperimentBuilder(object):
         """Get dictionary of options from configuration file provided in command-line"""
         return resolve_config.fetch_config(cmdargs)
 
-    def fetch_local_config(self, cmdargs):
-        """Get dictionary of options from all local sources (not from db)
-
-        .. seealso::
-
-            `orion.core.io.experiment_builder` for more information on the hierarchy of
-            configurations.
-        """
-        default_options = self.fetch_default_options()
-        env_vars = self.fetch_env_vars()
-        cmdconfig = self.fetch_file_config(cmdargs)
-
-        return resolve_config.merge_configs(default_options, env_vars, cmdconfig, cmdargs)
-
     def fetch_config_from_db(self, cmdargs):
         """Get dictionary of options from experiment found in the database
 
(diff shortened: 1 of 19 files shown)


bouthilx:
@tsirif @DatCorno If we can merge #96, many of the commits from this one will disappear.

bouthilx:
@DatCorno @tsirif Ready for final review

src/orion/core/io/experiment_builder.py
```diff
@@ -0,0 +1,262 @@
...
+
+        """
+        default_options = self.fetch_default_options()
+        env_vars = self.fetch_env_vars()
+        if use_db:
+            config_from_db = self.fetch_config_from_db(cmdargs)
+        else:
+            config_from_db = {}
+        cmdconfig = self.fetch_file_config(cmdargs)
+        metadata = dict(metadata=self.fetch_metadata(cmdargs))
```
DatCorno: I still maintain that we should use `{metadata=self.fetch_metadata(cmdargs)}` but this is just nitpicking ;)

bouthilx:
Now in #114

PR #103, opened 2018-06-04
Feature/trial hash

DatCorno:
Change trials id to hashes of their params and their experiment id


src/orion/core/worker/experiment.py
```diff
@@ -219,13 +219,16 @@ def register_trials(self, trials):
...
+        try:
+            stamp = datetime.datetime.utcnow()
+            for trial in trials:
+                trial.experiment = self._id
+                trial.status = 'new'
+                trial.submit_time = stamp
+            trials_dicts = list(map(lambda x: x.to_dict(), trials))
+            self._db.write('trials', trials_dicts)
+        except DuplicateKeyError:
+            pass
```
tsirif: Could it be the case that we want this to be reraised? If determinism in the sequence of generated trials was ensured by a future PR, the behaviour caught here would be unacceptable, and it should not have been silenced with a `pass` clause.

src/orion/core/worker/trial.py
```diff
@@ -141,7 +141,7 @@ class Param(Value):
 
         allowed_types = ('integer', 'real', 'categorical')
 
-    __slots__ = ('experiment', '_id', '_status', 'worker',
+    __slots__ = ('experiment', '_status', 'worker', '_id',
```
tsirif: ?

tsirif: Also, it seems that worker is not used at all. But that's a general comment about trials, independent of this PR's intention. Should it be removed, to be re-introduced whenif needed?

src/orion/core/worker/experiment.py
```diff
@@ -385,7 +388,7 @@ def stats(self):
             )
         completed_trials = self._db.read('trials', query,
                                          selection={'_id': 1, 'end_time': 1,
-                                                    'results': 1})
+                                                    'results': 1, 'experiment': 1, 'params': 1})
```
tsirif: Why is that?

src/orion/core/worker/trial.py
```diff
@@ -21,7 +21,7 @@ class Trial(object):
     Attributes
     ----------
     experiment : str
-       Unique identifier for the experiment taht produced this trial.
+       Unique identifier for the experiment that produced this trial.
```
tsirif: nice!

src/orion/core/worker/trial.py
```diff
@@ -155,7 +155,6 @@ def __init__(self, **kwargs):
                 setattr(self, attrname, None)
 
         self.status = 'new'
-        self._id = self.NoID
```
tsirif: Revert this please. Default trial ID is `Trial.NoID`, which in current agreed semantics is Python's `None`, but in the future it could be something else.

src/orion/core/worker/trial.py
```diff
@@ -167,13 +166,20 @@ def __init__(self, **kwargs):
...
+                    setattr(self, attrname, value)
 
     def to_dict(self):
         """Needed to be able to convert `Trial` to `dict` form."""
         trial_dictionary = dict()
 
         for attrname in self.__slots__:
+            if attrname == '_id':
+                continue
+
```
tsirif: Stuff like this are code smells. See total review.

src/orion/core/worker/trial.py
```diff
@@ -233,14 +234,13 @@ def gradient(self):
         """
         return self._fetch_one_result_of_type('gradient')
 
-    @property
-    def is_registered(self):
-        """Check whether `Trial` is registered in database based on `_id` value."""
-        return self._id is not self.NoID
+    def _repr_values(self, values, sep=','):
+        """Represent with a string the given values."""
+        return sep.join(map(lambda value: "{0.name}:{0.value}".format(value), values))
```
tsirif: Isn't pylint nagging about this being a static method? Could be located in `Trial.Value` as one.

src/orion/core/worker/trial.py
```diff
@@ -183,12 +189,7 @@ def to_dict(self):
             trial_dictionary[attrname] = list(map(lambda x: x.to_dict(),
                                                   getattr(self, attrname)))
 
-        # Trial object should not be able to set a value
-        # (whatever value) to its `_id` on its own, but is should just
-        # expect that it could have one
-        trial_dictionary.pop('id')  # Pop invalid name for database id key
-        if self.is_registered:
-            trial_dictionary['_id'] = self.id
+        trial_dictionary['_id'] = self.id
```
tsirif: I am in agreement with these changes since `_id` is to be dictated in reverse now!

src/orion/core/worker/trial.py
```diff
@@ -250,7 +250,11 @@ def hash_name(self):
         """
         if not self.params:
             raise ValueError("Cannot distinguish this trial, as 'params' have not been set.")
-        return hashlib.md5(self.params_repr().encode('utf-8')).hexdigest()
+        return hashlib.md5((self.params_repr() + str(self.experiment)).encode('utf-8')).hexdigest()
```
tsirif: <3 Future will have seed as well. Let's raise an issue to start implementing determinism.

src/orion/core/worker/experiment.py
bouthilx: The plan would be to either let the error climb or catch and return True/False to confirm if the trial was registered properly. Also, if we expect writes on db to raise duplicatekey errors, it may be better to turn this function into `register_trial()` instead of `register_trials()`.

src/orion/core/worker/trial.py
bouthilx: I would leave it there since it's not related to this PR.

src/orion/core/worker/trial.py
DatCorno: No, Pylint is surprisingly fine with it

src/orion/core/worker/experiment.py
bouthilx: Because `trial.id` is built using params and experiment's id. So if we query the db and build the trials using an incomplete definition, we can't recompute `trial.id`.

src/orion/core/worker/trial.py
bouthilx: A trial would now have an id by construction, it cannot be Trial.NoID anymore.

src/orion/core/worker/trial.py
DatCorno: Probably, is it even set somewhere beside the test trials?

src/orion/core/worker/trial.py
bouthilx: Indeed!

src/orion/core/worker/experiment.py
DatCorno: This would raise an error without the `experiment` and `params` when getting the id since we don't have the experience or the params set to a value.

src/orion/core/worker/trial.py
DatCorno: I agree, it hurt a little writing this

src/orion/core/worker/trial.py
DatCorno: Do you agree with the use of MD5?

bouthilx:
@tsirif I'm not sure about the implementation of `id` as `if-else: NoID`. Such behavior would be confusing, and trials are not really meant to change, beside getting more results over training. Something less confusing, in my opinion, could be 

```python
if self.params is None or self.experiment is None:
    raise RuntimeError("Trial id cannot be inferred before params and experiment are set")

return self.__hash__()
```

Note that this check could be located directly in `__hash__` rather than `id`.

DatCorno:
@tsirif Regarding the skipped test, this one is kinda of a problem... Since the DumbAlgo returns the same point twice, we're doomed to get `DuplicateKeyError`. We could adapt the test to check for only one new trial inside the database?

src/orion/core/worker/trial.py
```diff
@@ -167,13 +166,20 @@ def __init__(self, **kwargs):
                 for item in value:
                     attr.append(self.Param(**item))
             else:
-                setattr(self, attrname, value)
+                if attrname == '_id':
```
bouthilx: This modification could be dropped once `self._id` is removed.

bouthilx:
@tsirif @DatCorno For the skipped test, it would be tedious to fix it now and next PR about non-blocking design of trial sampling should help this, this is why I told François to mark the test for skip. It should be re-enabled very soon.

src/orion/core/worker/experiment.py
bouthilx: @tsirif @DatCorno I was thinking about this and the particular issue here is really related to non-blocking algo sampling design. I propose we leave this methods as-is in this PR, maybe with a clear TODO comment like "This method should be refactored to accept one trial at a time only, and return true or false if trial was registered properly or not". The PR fixing this should be done before end of this week.

bouthilx:
@tsirif @DatCorno I'd say the current PR only modification needed is removing _id from Trial.

DatCorno:
Done. Should we remove `worker` now?

bouthilx:
I would leave it there, @tsirif ?

src/orion/core/worker/experiment.py
bouthilx: @DatCorno You can drop the _id in the selection.

src/orion/core/worker/trial.py
```diff
@@ -250,7 +247,11 @@ def hash_name(self):
         """
         if not self.params:
```
bouthilx: Should add self.experiment in the checks here.

bouthilx:
@DatCorno 2 small changes and then good to merge IMO.

src/orion/core/worker/trial.py
DatCorno: What's the second change?

src/orion/core/worker/experiment.py
bouthilx: @DatCorno Second change is here.

src/orion/core/worker/experiment.py
tsirif: I am not sure why this error needs to be caught and get silenced.

I argue that this behaviour, if it ever appears in the future design, it indicates that determinism was not implemented correctly. So this would be our software's error, and we shouldn't let it pass silently because we will never be aware of it. Even if log an error message, if the event is rare enough, this is not a reliable way to have it issued. Having said these, I believe:
1. that we should use the error reporting system of Python here, which is raising exceptions
2. not silence the exception, but at least let the native one break the program, or reraising it from another orion-specific exception to include extra semantical information about how it happened.

I am in favour of making `register_trials` -> `register_trial` and iterating over them in `Producer`.

src/orion/core/worker/trial.py
tsirif: If we guarantee that we can create an id during initialization, then should also `assert params in kwargs and experiment in kwargs`.

src/orion/core/worker/trial.py
tsirif: At least for now, I do :stuck_out_tongue: stress test tries to see if there will be a problem with this.

src/orion/core/worker/experiment.py
tsirif: Could we do this now?

src/orion/core/worker/trial.py
```diff
@@ -141,7 +141,7 @@ class Param(Value):
 
         allowed_types = ('integer', 'real', 'categorical')
 
-    __slots__ = ('experiment', '_id', '_status', 'worker',
+    __slots__ = ('experiment', '_status', 'worker',
                  'submit_time', 'start_time', 'end_time', 'results', 'params')
     allowed_stati = ('new', 'reserved', 'suspended', 'completed', 'interrupted', 'broken')
     NoID = None
```
tsirif: This should also be removed then

src/orion/core/worker/experiment.py
bouthilx: The plan is to silence it only in this PR and fix this in the PR about non-blocking design for algo sampling. When we have this latter PR, this kind of error can be catched in the producer and considered a race condition, meaning that the algo should be updated again before resampling.

src/orion/core/worker/trial.py
```diff
@@ -250,12 +247,16 @@ def hash_name(self):
...
+    def __hash__(self):
+        """Return the hashname for this trial"""
+        return self.hash_name
 
     @property
     def full_name(self):
         """Generate a unique name using the full definition of parameters."""
-        if not self.params:
+        if not self.params or not self.experiment:
             raise ValueError("Cannot distinguish this trial, as 'params' have not been set.")
```
tsirif: idem

src/orion/core/worker/trial.py
```diff
@@ -250,12 +247,16 @@ def hash_name(self):
         """
         if not self.params:
             raise ValueError("Cannot distinguish this trial, as 'params' have not been set.")
```
tsirif: change error message of `ValueError`

src/orion/core/worker/trial.py
tsirif: `if not self.params or not self.experiment`

src/orion/core/worker/trial.py
bouthilx: In my opinion, a non-configured Trial should raise a RuntimeError whenever we try to hash it. Because of this, we don't need to test for params and experiment during initialization.

tsirif: nevermind on that

src/orion/core/worker/experiment.py
bouthilx: Doing it now would mean implementing the non-blocking parallel sampling in this PR.

src/orion/core/worker/trial.py
DatCorno: What should be remove, worker?

DatCorno: Ohhh, `NoID` nvm

DatCorno:
@tsirif we gucci now?

src/orion/core/worker/trial.py
```diff
@@ -245,8 +244,8 @@ def hash_name(self):
 
         .. note:: Two trials that have the same `params` must have the same `hash_name`.
         """
-        if not self.params:
-            raise ValueError("Cannot distinguish this trial, as 'params' have not been set.")
+        if not self.params and not self.experiment:
```
bouthilx: Uh-Oh. It's `and`, but should be `or`.

Issue #81, opened 2018-05-16
Documentation API is not available on readthedocs


bouthilx:
API is not available on orion.readthedocs.io, yet it should based on [index.rst](https://github.com/mila-udem/orion/blob/develop/docs/src/index.rst).

Excerpt:
```
.. toctree::
   :maxdepth: 2
   :caption: Code Reference

   reference/modules
```

This is because sphinx-api is normally called on command line. See tux's commands for documentation compilation:
```
    sphinx-apidoc -f -E -e -M --implicit-namespaces --ext-doctest -o docs/src/reference src/orion
    sphinx-build -E -W --color -c docs/src/ -b html docs/src/ docs/build/html
    sphinx-build -E -W --color -c docs/src/ -b man docs/src/ docs/build/man
```

readthedocs does not have an option for running particular commands, we need to embed sphinx-apidoc inside `docs/src/conf.py`. See [ref](https://github.com/rtfd/readthedocs.org/issues/1139#issuecomment-312626491)

I didn't manage to make it work yet. I get a weird error:
```
Running Sphinx v1.7.2
usage: sphinx-build [OPTIONS] -o <OUTPUT_PATH> <MODULE_PATH> [EXCLUDE_PATTERN, ...]
sphinx-build: error: the following arguments are required: -o/--output-dir
```

bouthilx:
Fixed by #106

PR #106, opened 2018-06-16
Fix code reference in documentation for ReadTheDocs (#81)

bouthilx:
A preview of the new doc is available here: https://orion.readthedocs.io/en/hotfix-doc/

### Why:

We were previously relying on sphinx-apidoc, but it is very difficult to make it work with read-the-docs and it does not give enough control on the format of the documentation. 

### How:

Manually build the API documentation tree, using automodule to fill modules' documentation.

### Note:

I also took the liberty of changing the table of content. I divided "Developer's guide" into "Plugin development's guide" and "Developer's guide". The former is meant as a user guide, for users who would like to develop their own plugin (Algo, database or analysis), while the latter is meant for contributors to Oríon itself. I also moved "Developer's guide" to the end, as this is the less interesting section for users. 

I added a section for examples, for now it only contains one example using pytorch-cifar. There is also a short example at the beginning of the user guide. Both examples are not finished, we will need to add more explanations.


docs/src/conf.py
```diff
@@ -48,6 +48,7 @@
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    # 'sphinx.ext.napolean',
```
DatCorno: If it's commented out, then we don't need it, so we should remove it.

bouthilx:
@tsirif The doc can be viewed here: https://orion.readthedocs.io/en/hotfix-doc/ (note that its not the default page, its the one for branch hotfix-doc)

Issue #87, opened 2018-05-17
Log trials responsible for the creation of another trial


bouthilx:
In the context of experiment version control trials generated by an algorithm may be biased by trials coming from parent or children experiments. To help identify such potential biases we should log in a new trial which trials was part of the algorithm's history at the time it generated it.

We could represent the log as a simple list of trial ids. However if there is thousands of trials then the size of this list will significantly out-weight the rest of the trial. This could use a significant amount of memory and slow down queries when the list of ids is selected.

Another solution is to represent the log as a tree of trial ids. A trial (1) would point to trial (2) which point to trial (3), meaning that trial (1) was created with both trial (2) and (3) in algorithm's history. Worst case scenario is a flat tree, which brings back to the list of ids, but it does not seam like a very likely situation, unless someone sample 1000 trials before starting executing them.

DatCorno:
Isn't this already done?

bouthilx:
Waiting for #152

PR #108, opened 2018-06-18
Non blocking hyperparameters evaluation

DatCorno:
This PR aims to solve the problem when creating new trials that already have been explore. Every time we hit a `DuplicateKeyError` we update the algorithm then sample again until we found a correct trial or we've reached a set limit of maximum tries.

This PR also modifies the DumbAlgo to act in a more deterministic way.

Solves #87


bouthilx:
@DatCorno Don't worry about the failed tests, I'm currently adding new unit-tests for concurrent producers and I will take care of those at the same time. You can focus on the CLI for visualizations. :)

bouthilx:
Now in #137

PR #111, opened 2018-06-20
feature/add_others_to_allowedtypes

dendisuhubdy:
adding `others` to allowed-types in trial 

where I tried to pass a statistic of the run as a dictionary object to the results list in orion.results but not allowed too, based on this error

```Traceback (most recent call last):
  File "/home/keatext/.local/share/virtualenvs/keanet-tbWY-zLv/bin/orion", line 11, in <module>
    load_entry_point('orion.core', 'console_scripts', 'orion')()
  File "/home/keatext/code/orion/src/orion/core/cli/__init__.py", line 35, in main
    orion_parser.execute(argv)
  File "/home/keatext/code/orion/src/orion/core/cli/resolve_config.py", line 137, in execute
    function(args)
  File "/home/keatext/code/orion/src/orion/core/cli/hunt.py", line 53, in main
    _execute(args, config)
  File "/home/keatext/code/orion/src/orion/core/cli/hunt.py", line 81, in _execute
    workon(experiment)
  File "/home/keatext/code/orion/src/orion/core/worker/__init__.py", line 46, in workon
    consumer.consume(trial)
  File "/home/keatext/code/orion/src/orion/core/worker/consumer.py", line 71, in consume
    completed_trial = self._consume(trial, workdirname)
  File "/home/keatext/code/orion/src/orion/core/worker/consumer.py", line 117, in _consume
    value=res['value']) for res in results]
  File "/home/keatext/code/orion/src/orion/core/worker/consumer.py", line 117, in <listcomp>
    value=res['value']) for res in results]
  File "/home/keatext/code/orion/src/orion/core/worker/trial.py", line 98, in __init__
    setattr(self, attrname, value)
  File "/home/keatext/code/orion/src/orion/core/worker/trial.py", line 127, in type
    type_, self.allowed_types))
ValueError: Given type, statistics, not one of: ('objective', 'constraint', 'gradient')```

### DIFF
--- a/src/orion/core/worker/trial.py
+++ b/src/orion/core/worker/trial.py
@@ -132,7 +132,7 @@ class Trial(object):
         function or of an 'constraint' expression.
         """
 
-        allowed_types = ('objective', 'constraint', 'gradient')
+        allowed_types = ('objective', 'constraint', 'gradient', 'other')
 
     class Param(Value):
         """Types for a `Param` can be either an integer (discrete value),
--- a/src/orion/core/worker/trial.py
+++ b/src/orion/core/worker/trial.py
@@ -132,7 +132,7 @@ class Trial(object):
         function or of an 'constraint' expression.
         """
 
-        allowed_types = ('objective', 'constraint', 'gradient', 'other')
+        allowed_types = ('objective', 'constraint', 'gradient', 'statistics')
 
     class Param(Value):
         """Types for a `Param` can be either an integer (discrete value),


bouthilx:
I'm not sure we should add something as non-specific. I would rather add types like 'metric', 'measure', 'statistic', etc. Or I would simply relieve the constraint on types and allow any.

@tsirif What is your though on this?

tsirif:
I agree on this point. A label 'other' means that any constraint can be
bypassed, and deprecates all of them. I disagree on using it.

The other reason is on what information can be passed about a point
evaluating a function to be optimized. All that can be told about a
function at a particular point of its domain, and which are potentially
useful for its optimization, are the derivatives (value, gradient, higher
order stuff) and the constraints. Higher order derivatives are neglected on
purpose because no one will do newton-gauss in hyperparameter search...

Is there a reasoning for this inclusion?

bouthilx:
The reason here is for experiment management. Suppose we do not get Experiment Version Control and related stuff out of Oríon, we should expect users to try using this tool for experiment management too.

tsirif:
I cannot see how EVC is related to an 'other' result of a trial. Can you
give me an example?

dendisuhubdy:
@tsirif I wanted to pass in a dict full of statistics of the run, the F1 score, precision, recall. So when I passed it through either one categories I got a Write error.

bouthilx:
@tsirif It is related in the sense that EVC is more about experiment management than optimization and the use case of Dendi is more about experiment management than optimization. If I may turn my sentence more general: Suppose we do not get experiment management related stuff out of Oríon, we should expect users to try using this tool for experiment management too.

bouthilx:
Hi Dendi,
We discussed the idea of adding new types of results in our meeting today. It was decided that the role of Oríon is only about black-box optimization and not about experiment management. We may provide experiment management in the future in the form of another tool or an optional module, but for now we recommend using tensorboard, sacred or any other similar tool which can be used for experiment management.
Thank you for the PR anyhow!

dendisuhubdy:
@bouthilx no worries

PR #115, opened 2018-07-04
Feature/version control git

ReyhaneAskari:
@bouthilx , `infer_versioning_metadata` and its test are added.


tests/unittests/core/test_resolve_config.py
```diff
@@ -0,0 +1,54 @@
...
+def test_infer_versioning_metadata():
+    """
+    Test how `infer_versioning_metadata` fills its different fields :
+    `is_dirty`, `HEAD_sha`, `active_branch` and `diff_sha`.
+    """
+    os.chdir('../')
+    if not os.path.exists('dummy_orion'):
+        os.makedirs('dummy_orion')
+        os.chdir('dummy_orion')
+        git.Repo.clone_from('https://github.com/ReyhaneAskari/dummy_orion.git', '.')
```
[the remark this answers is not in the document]
ReyhaneAskari: Fixed.

tests/unittests/core/test_resolve_config.py
```diff
@@ -0,0 +1,54 @@
...
+
+    test_repo = git.Repo('.git')
+
+    existing_metadata = {}
+    existing_metadata['user_script'] = '.git'
+    existing_metadata = resolve_config.infer_versioning_metadata(existing_metadata)
+    assert not existing_metadata['VCS']['is_dirty']
+    assert existing_metadata['VCS']['HEAD_sha'] == 'd316303cfcd1394df861ec33658faf42083d3d55'
```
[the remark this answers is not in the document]
ReyhaneAskari: We don't need the `monkeypatch`. I just did a dummy-commit to pass this.

bouthilx:
Thanks a lot! We'll need to make a rebase. It may be complicated if you are not familiar with all the internals of Oríon. I can help.

tests/unittests/core/io/test_resolve_config.py
```diff
@@ -357,3 +357,18 @@ def test_fetch_user_repo_on_non_repo():
...
+    """Test in the case of a detached head."""
+    with open('README.md', 'w+') as f:
+        f.write('dummy contentt')
+    repo.git.add('README.md')
+    repo.index.commit('2nd commit')
+    existing_metadata = {}
+    existing_metadata['user_script'] = '.git'
+    repo.head.reference = repo.commit('HEAD~1')
+    assert repo.head.is_detached
+    existing_metadata = resolve_config.infer_versioning_metadata(existing_metadata)
```
bouthilx: Maybe add assert [VCS][activate_branch] is None?

tests/unittests/core/io/test_resolve_config.py
```diff
@@ -370,5 +362,4 @@ def test_infer_versioning_metadata_on_detached_head(repo):
     repo.head.reference = repo.commit('HEAD~1')
     assert repo.head.is_detached
     existing_metadata = resolve_config.infer_versioning_metadata(existing_metadata)
-    os.chdir('../')
-    shutil.rmtree('dummy_orion')
+    assert existing_metadata['VCS']['active_branch'] is None
```
bouthilx: oh, great! :smile:

tests/unittests/core/experiment.yaml
```diff
@@ -18,9 +18,13 @@
     # from `orion`'s args
     user_args: ["--encoding_layer~choices(['rnn', 'lstm', 'gru'])", "--decoding_layer~choices(['rnn', 'lstm_with_attention', 'gru'])"]
     # from user's vsc, e.g. git
-    user_vcs: git
-    user_version: ~
-    user_commit_hash: as5f7asf5asfa7sf
+    VCS:
+      type: git
+      is_dirty: False
+      HEAD_sha: "d44aecee634f0fadb08c0dbe96cc6c8202b77bc8"
```
bouthilx: Isn't this overwritten by current version of the repo?

bouthilx:
@ReyhaneAskari Unit-tests are fine to me in their current shape, thank you! 

I find it puzzling that the fixed version in the db does not get overwritten in tests however. I'd like to investigate this further before merging.

ReyhaneAskari:
I still have to add one test.

ReyhaneAskari:
We got bunch of new errors. I'll check them tomorrow.

bouthilx:
@ReyhaneAskari I didn't look at every each of them but for the dozen I looked at it is all because of the infer_version_metadata. A conflict is now detected and it should not be the case. If you add the fixture for the mocked infer_version_metadata it should fix those errors.

bouthilx:
It's almost all good now! :tada: :smile:

ReyhaneAskari:
I don't have that error locally :| I changed it to what the sting was expected but I'm not sure why this has happened.

bouthilx:
It's because the dictionary order can vary. I would replace https://github.com/mila-udem/orion/blob/master/src/orion/core/evc/conflicts.py#L1141-L1143

with this
```python
        return "Old hash commit '{0}'  != new hash commit '{1}'".format(
            pprint.pformat(self.old_config['metadata'].get('VCS', None)).replace("\n", ""),
            pprint.pformat(self.new_config['metadata'].get('VCS')).replace("\n", ""))
```

So then the order of the printed dictionary is always the same (`pprint.pformat` sorts the keys) and would print as

```python
"{'HEAD_sha': 'test', 'active_branch': None, 'diff_sha': 'diff', 'is_dirty': False, 'type': 'git'}"
```

src/orion/core/io/resolve_config.py
```diff
@@ -242,9 +249,43 @@ def merge_configs(*configs):
...
-    # User repo's HEAD commit hash
-    return existing_metadata
+def fetch_user_repo(user_script):
+    """Fetch the GIT repo and its root path given user's script."""
+    dir_path = os.path.dirname(os.path.abspath(user_script))
+    try:
+        git_repo = git.Repo(dir_path, search_parent_directories=True)
+    except git.exc.InvalidGitRepositoryError as e:
+        git_repo = None
+        raise Exception('Script should be a git repo') from e
```
bouthilx:
```python
('Script '{}' should be in a git repository.'.format(os.path.abspath(user_script)))
```

src/orion/core/io/resolve_config.py
```diff
@@ -178,14 +180,19 @@ def fetch_metadata(cmdargs):
         if is_exe(abs_user_script):
             user_script = abs_user_script
 
+    if user_script and not os.path.exists(user_script):
+        raise Exception(
+            'The path {} specified for the script does not exist'.format(user_script))
```
bouthilx: SystemExit

tests/unittests/core/io/test_resolve_config.py
```diff
@@ -280,10 +289,81 @@ def test_merge_sub_configs_update_three():
...
+
+
+def test_fetch_user_repo_on_non_repo():
+    """
+    Test if `fetch_user_repo` raises a warning when user's script
+    is not a git repo
     """
-    metadata = {'hello': {'world': 0}}
-    assert resolve_config.infer_versioning_metadata(metadata) == metadata
+    with pytest.raises(Exception):
```
bouthilx:
```python
with pytest.raises(SystemExit) as exc_info:
    resolve_config.fetch_user_repo('.')
assert "Script '{}' should be in a git repository".format(os.getcwd()) in str(exc_info.value)
```

tests/unittests/core/io/test_resolve_config.py
```diff
@@ -60,25 +63,31 @@ def test_fetch_metadata_orion_version():
...
     metadata = resolve_config.fetch_metadata(cmdargs)
-    assert metadata['user_script'] == 'A'
+    assert metadata['user_script'] == 'tests/functional/demo/orion_config.yaml'
+
+
+def test_fetch_metadata_not_existed_path():
+    """Verfiy the raise of error when user_script path does not exist"""
+    cmdargs = {'user_args': ['dummy/path']}
+    with pytest.raises(Exception):
+        resolve_config.fetch_metadata(cmdargs)
```
bouthilx:
```python
with pytest.raises(SystemExit) as exc_info:
    resolve_config.fetch_metadata(cmdargs)
path = os.path.join(os.getcwd(), 'dummy/path') 
assert "The path {} specified for the script does not exist".format(path) in str(exc_info.value)
```

bouthilx:
I believe we should use SystemExit because those errors are because of the user, it is not an error in the system. Also, the unit-tests should test that it's the SystemExit exception it expected, by looking at the error message string.
Otherwise that's good! :)

src/orion/core/io/resolve_config.py
bouthilx: Sorry, `RuntimeError` would be better actually. `SystemExit` should be at the *cli* level. See this [comment](https://github.com/mila-udem/orion/pull/115#issuecomment-414173033) for a bit more info.
```python
RuntimeError('Script '{}' should be in a git repository.'.format(os.path.abspath(user_script)))
```

The advantage of using `RuntimeError` is that a try-catch on it is more specific than `Exception`.

ReyhaneAskari:
Thanks @bouthilx  for the detailed explanations! I guess we are finally done with this PR! 🎉

src/orion/core/io/resolve_config.py
```diff
@@ -181,8 +182,8 @@ def fetch_metadata(cmdargs):
             user_script = abs_user_script
 
     if user_script and not os.path.exists(user_script):
-        raise Exception(
-            'The path {} specified for the script does not exist'.format(user_script))
+        raise OSError(
+            errno.ENOENT, "The path {} specified for the script does not exist".format(user_script))
```
bouthilx: A more standard use of `OSError` would be without the format.
```python
OSError(errno.ENOENT, "The path specified for the script does not exist", user_script)
```

Giving:

```text
FileNotFoundError: [Errno 2] The path specified for the script does not exist: 'some_path'
```

bouthilx:
Thanks you for your patience! :) You will need to rebase this PR before we can merge.

src/orion/core/io/resolve_config.py
ReyhaneAskari: No, it won't print the path. The output will be : `FileNotFoundError(2, 'The path specified for the script does not exist')`

bouthilx: It depends, the `raise` will do, the `__repr__` of the error won't.

```python
>>> raise OSError(errno.ENOENT, "The path specified for the script does not exist", 'some_path')
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] The path specified for the script does not exist: 'some_path'

>>> OSError(errno.ENOENT, "The path specified for the script does not exist", 'some_path')
FileNotFoundError(2, 'The path specified for the script does not exist')
```

ReyhaneAskari: Got it. Thanks!

bouthilx:
Thank you!

PR #117, opened 2018-07-05
Fix setup

dmitriy-serdyuk:


tsirif:
I approve this message

dmitriy-serdyuk:
I just realized that it doesn't work with legacy python. But you guys support Python3 only, right?

bouthilx:
Indeed! 3.4-3.6 for now.

Thanks a lot for the PR by the way! It made it's way into the first release. :)

PR #121, opened 2018-07-10
Update documentation with extra example

Breakend:
This removes the user/pytorch documentation which is a duplicate of the example in the examples/ folder. It also adds an example for RL algorithm use and updates a TODO which should point to database setup docs.

Tested with:

tox -e docs
tox -e serve-docs


docs/src/examples/pytorch_cifar.rst
```diff
@@ -2,7 +2,9 @@
 Example with pytorch-cifar
 **************************
 
-# Link to database setup if not done
+.. note ::
+
+    If database not setup or Oríon not installed, follow setup instructions: :doc:`../installing` and :doc:`../database`
```
bouthilx: :+1: :smile:

bouthilx: We could say in the tutorial how to install orion. It's just ```pip install orion.core```. In the note, we would only point out for the database, and also say something like ```You can test the example without setting up a database by using the option `--debug`, but note that all data gathered during an execution will be lost at the end of it.```

bouthilx:
There is errors in pytorch_a2c_ppo.
```
docs/src/examples/pytorch_a2c_ppo.rst:12: D001 Line too long
docs/src/examples/pytorch_a2c_ppo.rst:44: D001 Line too long
docs/src/examples/pytorch_a2c_ppo.rst:78: D001 Line too long
```

You can test locally with 

```bash
$ tox -e docs
```
(tox is installable with pip if you don't have it yet)

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
+***********************************************
+Example with ikostrikov/pytorch-a2c-ppo-acktr
+***********************************************
```
bouthilx: too long

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
...
+
+.. note ::
+
+    If database not setup or Oríon not installed, follow setup instructions: :doc:`../installing` and :doc:`../database`
+
+Intro
+========
```
bouthilx: too long

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Here, we are looking to update the ikostrikov/pytorch-a2c-ppo-acktr Reinforcement Learning algorithm implementations to use Oríon to find the best hyperparameters while trying to prevent overfitting via a validation set of random evaluation seeds in the environment.
```
bouthilx: Too long

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
...
+
+What to change
+===============
```
bouthilx: idem

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
...
+
+
+To get the original repository of [ikostrikov](https://github.com/ikostrikov/pytorch-a2c-ppo-acktr) to work using Orion, we make a couple of changes.
```
bouthilx: ok, I won't report the other lines that are too long :laughing:

docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,84 @@
...
+
+and then we run
+
+.. code-block:: bash
+
+    chmod +x main.py
+
+To make it executable.
+
+Then, we ensure that we evaluate on a separate set of hold out random seeds for the environment (which should be different than the test set and training seed). For MuJoCo environments where the random seed has an effect, we can simply set the random seed before a rollout. In Atari, we would have to create a new validation set of rollouts perhaps with different human starts.
```
bouthilx: I think we should first show how to use it with current implementation and then say that it is not a good metric to optimize because of overfitting and point to the new file you implemented. Ohterwise is sounds like orion itself requires such a change, while it's something that is lacking in general. The execution with orion does not require it, it works without it, but we do not recommand it. The file `eval.py` is required for the sake of fair benchmarking.

Breakend:
Regarding the removing of the duplicate example in user, I added it back to the tree, but removed the duplicate file. I believe it's the index.rst that really matters, right?

bouthilx:
@Breakend Sorry I didn't realize you made the PR on master. All PRs are on develop, only release and exceptional bug fixes are on master. You will need to rebase your branch on develop, removing the following commits while doing so: 15ce25c, 4e657d0, 327b054 and a021792. 

For the duplicate, I think it is important to have a simple example at the beginning of the user tutorial, to get a grasp of Oríon in general. All the examples would be too specific and too much before a tutorial on how to use the tool in general. This is why I added a simple example at the beginning and others after user section. We could make the one in examples/ point to the one in users/ for now. Later on I propose we add another example for pytorch, something more complicated and more interesting.

bouthilx:
@Breakend Do you think you would have time to make the rebase in the coming week?

Breakend:
Closing this in favour of a new PR to make this easier

PR #123, opened 2018-07-19
Fix verbosity

dmitriy-serdyuk:
Currently `-vvv` corresponds to zero level of verbosity which is not very intuitive. I refactored this part, now verbosity >=2 corresponds to `DEBUG`.


bouthilx:
I do consider this as a bug, thank you very much for the fix! 

Minor comment, we try to follow git-flow workflow for the development. So a PR like this one should come from a branch with a name like `hotfix/verbosity_level`.

Issue #122, opened 2018-07-19
Use `argcomplete` for tab completion in shell


dmitriy-serdyuk:
Have you considered to use http://argcomplete.readthedocs.io/en/latest/ ? This project looks very nice.

bouthilx:
Yes indeed! Autocompletion on the experiment names would be very handy! I looked into it while searching for command line arguments autocompletion a few weeks ago but didn't have time to get into the details. I'll comment further in your PR.

Thomsch:
When we work on this feature, a new issue will be created.

PR #124, opened 2018-07-19
Autocomplete proof of concept

dmitriy-serdyuk:
A very small change, but it seems to be working:
```
$ orion
--debug    --help     --verbose  --version  -V         -d         -h         -v         hunt       init_only  insert
```

#122

### DIFF
--- a/src/orion/core/cli/base.py
+++ b/src/orion/core/cli/base.py
@@ -73,6 +73,14 @@ class OrionArgsParser:
         function(args)
 
 
+def exp_name_completer(prefix, action, parser, parsed_args):
+    from argcomplete import warn
+    warn(prefix)
+    names = ['exp1', 'exp2', 'exp3', 'mnist']
+    names = [name for name in names if name.startswith(prefix)]
+    return names
+
+
 def get_basic_args_group(parser):
     """Return the basic arguments for any command."""
     basic_args_group = parser.add_argument_group(
@@ -83,7 +91,7 @@ def get_basic_args_group(parser):
         '-n', '--name',
         type=str, metavar='stringID',
         help="experiment's unique name; "
-             "(default: None - specified either here or in a config)")
+             "(default: None - specified either here or in a config)").completer = exp_name_completer
 
     basic_args_group.add_argument('-c', '--config', type=argparse.FileType('r'),
                                   metavar='path-to-config', help="user provided "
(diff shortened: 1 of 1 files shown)


bouthilx:
@dmitriy-serdyuk Thank you so much for this proof of concept! 

Do you know what would be the drawback of integrating argcomplete? I had problems in our Mila system to set readline with autocompletion of Cmd.Cmd, so I wonder if using it with the commandline could make it unusable for people with broken compatibility with readline. If we can make it work and be sure broken compatibility does not affect the usability of the command line then I'm all in for this feature. I believe auto-completion is somehow a necessity because of how cumbersome it is to remember and type properly experiment names.

dmitriy-serdyuk:
Argcomplete works in a way that you need to register your completions. The user will need to run
```
eval "$(register-python-argcomplete orion)"
```
or
```
echo "$(register-python-argcomplete orion)" >> ~/.bashrc
```
for permanent completion.

bouthilx:
ok! What if the user does not set this? Is there any error message or it works without autocompletion seamlessly? I think I would make this optional since it requires an additional installation step and is not necessary for minimal functionality.

dmitriy-serdyuk:
This function basically checks some magic env variable set by the shell, otherwise continues script execution seamlessly. I don't think if there any problems here.

I agree, it makes sense to make it optional.

bouthilx:
Great! Since the installation requires some manual intervention of the user, I would only add a autocomplete section in the tutorials where we explain how to install, and in the code I would wrap imports to support autocomplete when argcomplete is available. That means, no modifications to setup.py or any requirements.txt files.

@tsirif Any opinion on this?

dmitriy-serdyuk:
A downside of this is that every time a user hits tab, orion will need to startup and query mongo for experiment names. Maybe some kind of local cache will help performance?

bouthilx:
@dmitriy-serdyuk Would you be interested in finishing it? We discussed about it add we would like to have this feature as an optional install, including autocompletion of arguments using callbacks to the database if possible.

dmitriy-serdyuk:
I can try when I have some time

bouthilx:
OK. Let me know if you need some help for the callbacks on the database. I can give a hand.

bouthilx:
I close this PR due to staleness. The addition of autocomplete for the commandline interface is added to our todo list.

PR #126, opened 2018-07-23
Fix github links

dmitriy-serdyuk:
Now links go to `http://orion.readthedocs.io/en/latest/github.com/mila-udem/orion` for example.


bouthilx:
Thanks!

PR #127, opened 2018-08-03
Add `hunt --worker-trials` argument

dmitriy-serdyuk:
Why:
-----
Sometimes it is desirable to limit number of trials to run on a specific node without changing the total number of trials for the whole experiment. For example, one could run only one trial per node on a cluster with `max_trials` set to some value. This helps to schedule jobs better on a cluster.

How:
-----
`--worker-trials` affects only a specific worker. The worker exits when either `max_trials` for the experiment reached or `worker_trials`.

### DIFF
--- a/src/orion/core/cli/hunt.py
+++ b/src/orion/core/cli/hunt.py
@@ -16,6 +16,7 @@ from orion.core.cli import base as cli
 from orion.core.cli import evc as evc_cli
 from orion.core.io import resolve_config
 from orion.core.io.evc_builder import EVCBuilder
+from orion.core.io.experiment_builder import ExperimentBuilder
 from orion.core.worker import workon
 
 log = logging.getLogger(__name__)
@@ -29,8 +30,17 @@ def add_subparser(parser):
 
     orion_group.add_argument(
         '--max-trials', type=int, metavar='#',
-        help="number of jobs/trials to be completed "
+        help="number of trials to be completed for the experiment. This value "
+             "will be saved within the experiment configuration and reused "
+             "across all workers to determine experiment's completion. "
              "(default: %s)" % resolve_config.DEF_CMD_MAX_TRIALS[1])
+
+    orion_group.add_argument(
+        '--worker-trials', type=int, metavar='#',
+        help="number of trials to be completed for this worker. "
+             "If the experiment is completed, the worker will die even if it "
+             "did not reach its maximum number of trials "
+             "(default: %s)" % resolve_config.DEF_CMD_WORKER_TRIALS[1])
 
     orion_group.add_argument(
         "--pool-size", type=int, metavar='#',
@@ -50,5 +60,7 @@ def main(args):
     """Build experiment and execute hunt command"""
     args['root'] = None
     args['leafs'] = []
+    # TODO: simplify when parameter parsing is refactored
+    worker_trials = ExperimentBuilder().fetch_full_config(args)['worker_trials']
     experiment = EVCBuilder().build_from(args)
-    workon(experiment)
+    workon(experiment, worker_trials)
--- a/src/orion/core/io/resolve_config.py
+++ b/src/orion/core/io/resolve_config.py
@@ -55,6 +55,7 @@ log = logging.getLogger(__name__)
 
 # Default settings for command line arguments (option, description)
 DEF_CMD_MAX_TRIALS = (infinity, 'inf/until preempted')
+DEF_CMD_WORKER_TRIALS = (infinity, 'inf/until preempted')
 DEF_CMD_POOL_SIZE = (10, str(10))
 
 DEF_CONFIG_FILES_PATHS = [
@@ -106,6 +107,7 @@ def fetch_default_options():
     # get some defaults
     default_config['name'] = None
     default_config['max_trials'] = DEF_CMD_MAX_TRIALS[0]
+    default_config['worker_trials'] = DEF_CMD_WORKER_TRIALS[0]
     default_config['pool_size'] = DEF_CMD_POOL_SIZE[0]
     default_config['algorithms'] = 'random'
 
--- a/src/orion/core/worker/__init__.py
+++ b/src/orion/core/worker/__init__.py
@@ -10,6 +10,7 @@
 
 """
 import io
+import itertools
 import logging
 import pprint
 
@@ -20,13 +21,18 @@ from orion.core.worker.producer import Producer
 log = logging.getLogger(__name__)
 
 
-def workon(experiment):
+def workon(experiment, worker_trials=None):
     """Try to find solution to the search problem defined in `experiment`."""
     producer = Producer(experiment)
     consumer = Consumer(experiment)
 
     log.debug("#####  Init Experiment  #####")
-    while True:
+    try:
+        iterator = range(int(worker_trials))
+    except (OverflowError, TypeError):
+        # When worker_trials is inf
+        iterator = itertools.count()
+    for _ in iterator:
         log.debug("#### Try to reserve a new trial to evaluate.")
         trial = experiment.reserve_trial(score_handle=producer.algorithm.score)
 
--- a/tests/functional/parsing/test_parsing_hunt.py
+++ b/tests/functional/parsing/test_parsing_hunt.py
@@ -26,7 +26,7 @@ def test_hunt_command_full_parsing(database, monkeypatch):
     parser, subparsers = _create_parser()
     args_list = ["hunt", "-n", "test",
                  "--config", "./orion_config_random.yaml",
-                 "--max-trials", "400", "--pool-size", "4",
+                 "--max-trials", "400", "--pool-size", "4", "--worker-trials", "5",
                  "./black_box.py", "-x~normal(1,1)"]
 
     hunt.add_subparser(subparsers)
@@ -38,3 +38,4 @@ def test_hunt_command_full_parsing(database, monkeypatch):
     assert args['user_args'] == ['./black_box.py', '-x~normal(1,1)']
     assert args['pool_size'] == 4
     assert args['max_trials'] == 400
+    assert args['worker_trials'] == 5


src/orion/core/cli/hunt.py
```diff
@@ -29,9 +29,14 @@ def add_subparser(parser):
 
     orion_group.add_argument(
         '--max-trials', type=int, metavar='#',
-        help="number of jobs/trials to be completed "
+        help="number of jobs/trials to be completed in total "
              "(default: %s)" % resolve_config.DEF_CMD_MAX_TRIALS[1])
 
+    orion_group.add_argument(
+        '--worker-trials', type=int, metavar='#',
+        help="number of jobs/trials to be completed for this worker "
```
bouthilx: I would add a note: 
```
If the experiment is completed, the worker will die even if it did not reach its maximum number of trials
```

dmitriy-serdyuk: b9f2b09

src/orion/core/worker/__init__.py
```diff
@@ -12,21 +12,24 @@
 import io
 import logging
 import pprint
+from sys import maxsize as infinity
 
 from orion.core.io.database import Database
 from orion.core.worker.consumer import Consumer
 from orion.core.worker.producer import Producer
 
+__all__ = ('workon',)
```
bouthilx: We do not support wildcard imports.

tests/functional/parsing/test_parsing_hunt.py
```diff
@@ -27,7 +27,7 @@ def test_hunt_command_full_parsing(database, monkeypatch):
     args_list = ["hunt", "-n", "test",
                  "--config", "./orion_config_random.yaml",
                  "--max-trials", "400", "--pool-size", "4",
-                 "./black_box.py", "-x~normal(1,1)"]
+                 "./black_box.py", "-x~normal(1,1)", "--worker-trials 5"]
```
bouthilx: As it is, the option ```--worker-trials``` is an argument to the user's script. It should be moved along with ```--max-trials``` and ```--pool-size```

src/orion/core/cli/hunt.py
```diff
@@ -29,9 +29,16 @@ def add_subparser(parser):
 
     orion_group.add_argument(
         '--max-trials', type=int, metavar='#',
-        help="number of jobs/trials to be completed "
+        help="number of jobs/trials to be completed in total for the experiment "
```
bouthilx: Maybe something like
```
number of trials to be completed for the experiment. This value will be saved within the experiment configuration and reused across all workers to determine experiment's completion.
```

src/orion/core/cli/hunt.py
```diff
@@ -29,9 +29,18 @@ def add_subparser(parser):
...
         '--max-trials', type=int, metavar='#',
-        help="number of jobs/trials to be completed "
+        help="number of trials to be completed for the experiment. This value "
+             "will be saved within the experiment configuration and reused "
+             "across all workers to determine experiment's completion. "
              "(default: %s)" % resolve_config.DEF_CMD_MAX_TRIALS[1])
 
+    orion_group.add_argument(
+        '--worker-trials', type=int, metavar='#',
+        help="number of jobs/trials to be completed for this worker. "
```
bouthilx: To be coherent now with ```--max-trials``` help's message. :)
```
number of jobs/trials to be completed [...]
```

bouthilx:
Wow, you push fixes so fast that it feels like my comments are automatically turned into fixes. :laughing:

For further changes, it would be appreciated if you could follow a similar template as https://github.com/mila-udem/orion/pull/112 for PR description and commit messages. Basically just What, How sections. It helps a lot when backtracking the history for bugs or design decisions. Thanks!

src/orion/core/cli/hunt.py
```diff
@@ -50,5 +59,6 @@ def main(args):
     """Build experiment and execute hunt command"""
     args['root'] = None
     args['leafs'] = []
+    worker_trials = args.pop('worker_trials')
```
bouthilx: Current [build](https://travis-ci.org/mila-udem/orion/builds/411819932) fails because of this line. The default value of ```worker_trials``` that is set within ```resolve_config``` actually takes place inside ```EVCBuilder().build_from(args)```. I know this is confusing, I'm currently preparing a PR for refactoring the configuration process. I'll adapt it when I make this PR.

For now you can fix with
```python
from orion.core.io.experiment_builder import ExperimentBuilder

# ... 

worker_trials = ExperimentBuilder().fetch_full_config(args)['worker_trials']
```

dmitriy-serdyuk:
Why don't you use argparse's default values?

bouthilx:
Because there is a complex hierarchy of configuration and commandline has the highest level in the hierarchy, meaning that any value passed by commandline would overwrite default value, global configuration file, local configuration file or even experiment configuration saved in db.  I tried to explain this in the docstring of ```experiment_builder.py``` [here](https://github.com/mila-udem/orion/blob/master/src/orion/core/io/experiment_builder.py).

After the refactoring I'm doing it will be simpler. Experiment configuration and Orion configuration will be decoupled and something like ```worker_trials``` would end up in the singleton object ```orion.config.worker``` rather then in the experiment's configuration dictionary.

tests/functional/parsing/test_parsing_hunt.py
```diff
@@ -26,7 +26,7 @@ def test_hunt_command_full_parsing(database, monkeypatch):
     parser, subparsers = _create_parser()
     args_list = ["hunt", "-n", "test",
                  "--config", "./orion_config_random.yaml",
-                 "--max-trials", "400", "--pool-size", "4",
+                 "--max-trials", "400", "--pool-size", "4", "--worker-trials 5",
```
bouthilx: ```"--worker-trials 5"``` -> ```"--worker-trials",  "5",```

bouthilx:
@dmitriy-serdyuk There is two unit-tests to fix for the change of ```infinity``` definition: https://travis-ci.org/mila-udem/orion/jobs/411836628#L571

After that I think this is all good.

dmitriy-serdyuk:
Should I squash?

bouthilx:
Yes but before that, I think you missed one of the infinity failure, in test_resolve_config.py. Those builds will probably fail.

src/orion/core/cli/hunt.py
```diff
@@ -29,9 +30,18 @@ def add_subparser(parser):
...
+             "will be saved within the experiment configuration and reused "
+             "across all workers to determine experiment's completion. "
              "(default: %s)" % resolve_config.DEF_CMD_MAX_TRIALS[1])
 
+    orion_group.add_argument(
+        '--worker-trials', type=int, metavar='#',
+        help="number of trials to be completed for this worker. "
+             "If the experiment is completed, the worker will die even if it "
+             "did not reach its maximum number of trials "
+             "(default: %s)" % resolve_config.DEF_CMD_WORKER_TRIALS[1])
```
bouthilx: Thinking about this, this will get a terrible rendering; ```(default: 9223372036854775807)```. I would rather stick to ```numpy.inf``` and handle it at the for-loop in ```worker```. Sorry for changing my mind late. :sweat_smile:

dmitriy-serdyuk: It renders `default: inf/until preempted`

dmitriy-serdyuk: `numpy.inf` is float, and even when you do `int(numpy.inf)` it crashes. It is very inconvenient value.

bouthilx: Oh, that's right. :thinking: I'm still not very comfortable with that fact that it is a precise value that cannot be said to be inf. For instance 1000000000000 would have the same practical effect but we could not say it is inf. But whatever, I guess that may not be very important.

bouthilx:
Alright! Waiting for @tsirif 's review before merging.

@dmitriy-serdyuk Thanks a lot for this PR and very quick responses!

src/orion/core/cli/hunt.py
dmitriy-serdyuk: Sometimes people use -1 for infinity, but this is also not very nice, because you have to do more coding with if/else everywhere.

bouthilx: Yeah, so would 'inf'. But both would not generate situations where the user load back its configuration and see ```{'max_trials': 9223372036854775807}``` :astonished:. I don't mind a couple of if-else if it makes user experience better. :wink: Also, -1 or 'inf' are not likely to change from one platform to another.

dmitriy-serdyuk: What about using `None` for infinity?

bouthilx: Could also be. What would be the advantage against `numpy.inf`?

dmitriy-serdyuk: Advantage is that it is a special value. Some integer functions work with `numpy.inf`, some don't. In the contrast, most of things will crash when `None` is passed. So, it will be easier to debug in the future. ("Let it crash" philosophy I guess)

dmitriy-serdyuk: And I really don't like the fact that variables defined with `int` type are assigned `float` type.

bouthilx: Those are good points indeed. I'm not sure however what could someone try to do with `worker_trials` other than printing it or iterating on it. So printing would work either way, and iterating would crash either way. `numpy.inf` however has the nice attribute that it is comparable with the proper behavior we are looking for.

dmitriy-serdyuk: What about `max_trials`? Where is it used?

bouthilx: There for instance:
https://github.com/mila-udem/orion/blob/master/src/orion/core/worker/experiment.py#L327

dmitriy-serdyuk:
Wow, you have a harsh linter.

src/orion/core/worker/__init__.py
```diff
@@ -20,13 +20,18 @@
...
     producer = Producer(experiment)
     consumer = Consumer(experiment)
 
     log.debug("#####  Init Experiment  #####")
-    while True:
+    try:
+        worker_trials = int(worker_trials)
+    except OverflowError:
+        # When worker_trials is inf
+        worker_trials = None
```
bouthilx: ```sys.maxsize```?

bouthilx: You could also use a while-loop rather than a for-loop. In that case it would reuse the fact that we can compare `inf` with integers. 
```python
nth_trial = 0
while nth_trial < worker_trials:
    nth_trial += 1
    # ...
```

tests/unittests/core/io/test_experiment_builder.py
```diff
@@ -38,7 +40,7 @@ def test_fetch_local_config_from_incomplete_config(incomplete_config_file):
     assert local_config['database']['host'] == 'mongodb://user:pass@localhost'
     assert local_config['database']['name'] == 'orion'
     assert local_config['database']['type'] == 'incomplete'
-    assert local_config['max_trials'] == float('inf')
```
bouthilx: We could keep it as-is.
```python
>>> float('inf') == numpy.inf
True
```

src/orion/core/io/resolve_config.py
```diff
@@ -37,6 +37,7 @@
 import socket
 
 from numpy import inf as infinity
+
```
bouthilx: There should be no space here, flake8 complains.

dmitriy-serdyuk: I don't get it, your linter doesn't like any order. With spaces or without.

bouthilx:
Third tier packages should be grouped together and in alphabetical order. I believe this should work:
```python
import logging
import os
import socket

from numpy import inf as infinity
import yaml

import orion
```

Note that this doesn't
```python
# ...

import yaml
from numpy import inf as infinity

# ...
```

bouthilx:
By the way, you can test it locally with tox.

```bash
$ pip install -r tests/requirements.txt
$ pip install tox
$ tox -e flake8
$ tox -e pylint
```

dmitriy-serdyuk:
I see. I was pretty sure that pep8 recommends to place `from ...` imports after `import ...`.

dmitriy-serdyuk:
Done, it works now.

bouthilx:
The current non-branching attributes such as `max-trials` would stay within the experiment configuration. `worker-trials` is currently only a local attribute in the sense that it is not logged in any database, it's only available at run time for a specific worker. I guess at some point it could be logged in database as a worker configuration, as you just said. Then the non-branching attributes of the experiment would stay within the experiment, and the local attributes of the worker would be saved in the worker's configuration.

bouthilx:
@dmitriy-serdyuk Can you please rebase this branch. We will merge it.

bouthilx:
Thanks again!

PR #128, opened 2018-08-03
[WIP] Feature/strategy

mnoukhov:
need to 
- figure out where to add `parallel_strategy` definition (in `experiment` or maybe in `algo`?)
- set `NoParallelStrategy` as default
- write tests for parallel strategy updates

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -25,7 +25,8 @@ from orion.core.io.interactive_commands.branching_prompt import BranchingPrompt
 from orion.core.io.space_builder import SpaceBuilder
 from orion.core.utils.format_trials import trial_to_tuple
 from orion.core.worker.primary_algo import PrimaryAlgo
-from orion.core.worker.strategies import Strategy
+from orion.core.worker.strategy import (BaseParallelStrategy,
+                                        Strategy)
 from orion.core.worker.trial import Trial
 
 log = logging.getLogger(__name__)
@@ -379,6 +380,10 @@ class Experiment(object):
                 config[attrname] = copy.deepcopy(config[attrname])
                 config[attrname]['adapter'] = config[attrname]['adapter'].configuration
 
+            if self._init_done and attrname == "producer" and attribute.get("strategy"):
+                config[attrname] = copy.deepcopy(config[attrname])
+                config[attrname]['strategy'] = config[attrname]['strategy'].configuration
+
         # Reason for deepcopy is that some attributes are dictionaries
         # themselves, we don't want to accidentally change the state of this
         # object from a getter.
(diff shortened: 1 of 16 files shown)


bouthilx:
If we follow the same structure, producer would be set by default inside `resolve_config` and overwritten by configuration file of user if defined or by experiment's configuration in database, again, if defined (this occurs in `ExperimentBuilder`. This means the attribute `experiment.producer` could be `None` at initialization of the object experiment, but it would always be set at the execution of `experiment.configure(config)` inside `ExperimentBuilder`, because `config` would always contain it.

src/orion/core/worker/experiment.py
```diff
@@ -297,6 +297,18 @@ def fetch_completed_trials(self):
 
         return completed_trials
 
+    def fetch_active_trials(self):
+        """Fetch currently incomplete trials for this `Experiment`
+
+        :return: list of incomplete `Trial` objects
+        """
+        # TODO(mnoukhov): add reserved to query?
```
bouthilx: Yes indeed!

bouthilx: It should actually be incompleted trials in general. That means any status that is not `completed`. We want to lie for new trials and broken trials as well.

src/orion/core/worker/strategies.py
```diff
@@ -0,0 +1,123 @@
...
+        objective = None
+    elif len(objectives) == 1:
+        objective = objectives[0]
+    elif len(objectives) > 1:
+        raise RuntimeError("Trial %d has %d objectives", trial.id, len(objectives))
+
+    return objective
+
+
+class BaseParallelStrategy(object, metaclass=ABCMeta):
```
bouthilx: There must be a factory class which inherits from BaseParallelStrategy and has metaclass=Factory in order to support creation of parallel strategies using the meta class You can look at OptimizationAlgorithm for a inspiration. Also, I would define parallel strategies outside the core, just like algos, so that users can customize them as well. Parallel strategies are heuristics and there is world of possibilities to explore.

src/orion/core/worker/strategies.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def lie(self, trial):
+        """See BaseParallelStrategy.lie"""
+        if get_objective(trial):
+            raise RuntimeError("Trial %d is completed but should not be.", trial.id)
+
+        return Trial.Result(name='lie', type='lie', value=self.mean_result)
+
+
+class Strategy(BaseParallelStrategy, metaclass=Factory):
```
bouthilx: Oh, there it is. `ParallelStrategy` maybe?

src/orion/core/worker/strategies.py
```diff
@@ -0,0 +1,123 @@
...
+    """Class used to build a parallel strategy given name and params
+
+    .. seealso:: `orion.core.utils.Factory` metaclass and `BaseParallelStrategy` interface.
+    """
+
+    @classmethod
+    def build(cls, strategy_dict):
```
bouthilx: This is useless. One should be able to create parallel strategies with `Strategy(of_type='SomeParallelStrategy', some=arguments)` The `Factory` will build it automatically. `BaseAlgorithm` has a complex machinery in `__init__` because it supports recursive definitions of algorithms, but if you look at `AbstractDB` (an equivalent of `BaseAlgorithm` but for databases) you can see that there is nothing. It's the same for parallel strategies.

PR #135, opened 2018-08-17
Introduction of the Concept base class

DatCorno:
This PR introduces a new class called `Concept` inside the utils module. It also refactors the way the Factory metaclass works and changes to code to handle this new way of doing things.

### `Factory`
The `Factory` metaclass was working great until now. However, it had one major drawback: it could only instantiate objects from `.py` files that reside inside the same folder as the base class. This means that the architecture of things is very limited. The new `Factory` will now recursively look for `.py` files inside the base class folder. Classes are now stored inside the `typenames` list through their qualified name, and not just their file name to prevent any conflicts from submodules that would share the same name without the same path (`orion.viz.matplot.bar` and `orion.viz.panda.bar` for example).

Another change that ties in with the introduction of `Concept` is that the factories are not explicitly defined inside the code. This required some changes to the entry points definition, which will now be listed through the base class name and not the factory name.

### `Concept`
This new class comes from the desire to extract the creation process of an object like `BaseAlgorithm` so that this runtime dynamic instantiation can be use towards new type without copy-pasting the code. Classes that inherit from `Concept` need to define one class-wide constant called `name` which will be use for logging purposes. It is a human-readable string representation of the concept (`Algorithm` for `BaseAlgorithm` for example).
They can also define other class-wide constants. The only one that currently has an impact is the `implementation_module` one which overrides the module name use when getting the qualified name of subtypes. This is useful in the case where the base class exists in a non-`__init__.py` file. This use case comes up when creating concepts that can be extended through other packages via `pip install`. Since having a `__init__.py` file inside the folder would prevent the new packages from being injected, the base concept cannot be declared there.
The process of instantiating a subtype has changed a little bit. Now, if the dictionary contains multiple elements, it tries to instantiate every single one of them. If it fails, it considers that item as a class attribute which calls `setattr` on it.

### Other changes
Here's a bullet list of all the others less important changes coming with this PR
- The adapters have been changed to follow the new `Factory` call convention
- `PrimaryAlgo` has been moved to `orion.algo.base` for consistency reasons
- The `Database` required special handling for their situation, since they are not an actual `Concept` child class.
- `DumbAlgo` has been moved to its own egg so that it can be instantiated through the `BaseAlgorithm` and not require factory hacks.
- The `capture=no` statement inside the `tox:devel` environment was causing trouble. It has been removed.


src/orion/core/io/database/__init__.py
```diff
@@ -255,11 +252,169 @@ class DuplicateKeyError(DatabaseError):
     pass
 
 
-# pylint: disable=too-few-public-methods,abstract-method
-class Database(AbstractDB, metaclass=SingletonFactory):
-    """Class used to inject dependency on a database framework.
+# pylint: disable=too-few-public-methods,abstract-method,too-many-public-methods
+class Database(Wrapper, metaclass=SingletonType):
```
bouthilx: I would drop the abstract and make a wrapper like we do for views; automatically fetch wrapped concept's attributes if not overriden by the wrapper.

src/orion/core/io/experiment_builder.py
```diff
@@ -206,8 +206,9 @@ def build_view_from(self, cmdargs):
 
         # Information should be enough to infer experiment's name.
         log.debug("Creating %s database client with args: %s", dbtype, db_opts)
+
         try:
-            Database(of_type=dbtype, **db_opts)
```
bouthilx: Why remove `of_type`? It should still work based on `Dataset.__init__`, right?

src/orion/core/utils/__init__.py
```diff
@@ -27,6 +27,15 @@ def nesteddict():
     return defaultdict(nesteddict)
 
 
+def get_qualified_name(package, name):
```
bouthilx: `_get_qualified_name`

src/orion/core/utils/__init__.py
```diff
@@ -131,12 +148,17 @@ def __call__(cls, of_type, *args, **kwargs):
 
         :return: The object which was created on the first call.
         """
+        module, name = of_type
+        qualified_name = get_qualified_name(module, name).lower()
+
         for inherited_class in cls.types:
-            if inherited_class.__name__.lower() == of_type.lower():
+            inh_qualified_name = get_qualified_name(inherited_class.__module__,
```
bouthilx: `inh_qualified_name` :scream: 
`inherited_class_name` maybe?

src/orion/core/io/database/__init__.py
DatCorno: You mean, drop the `ABCMeta` metaclass from `AbstractDB`?

src/orion/core/io/experiment_builder.py
DatCorno: Because the `Database` wrapper takes care of constructing the dictionary, it just removes unnecessary fluff from the call and reduces the complexity of the API, since you don't have to remember to call through explicit arguments

src/orion/core/utils/__init__.py
DatCorno: I'm tempted to leave it public just in case it is needed elsewhere

src/orion/core/utils/__init__.py
DatCorno: Yes :grimacing:

bouthilx:
[Review Concept bass class](https://trello.com/c/ysNcrOZM/2-review-concept-bass-class)

bouthilx:
@DatCorno I made a PR for small changes: https://github.com/DatCorno/orion/pull/3

Sorry for the horribly long delay! Once this is ready for merge, I'll fix the dependency of `orion.algo.skopt` to `orion.core <= v0.1.2` and adjust it to `Concept`. I'll make simultaneous release candidates for both `orion.core` and `orion.algo.skopt`.

bouthilx:
@DatCorno Can you rebase it? It should trigger the tests at the same time.

DatCorno:
Glad to see this finally being closed! What a great time we had on that airplane :')

PR #136, opened 2018-08-20
Add RL example

Breakend:
I think this is the rebased version of the previous PR, please double check that this doesn't mess anything up.

### DIFF
--- a/docs/src/examples/pytorch_a2c_ppo.rst
+++ b/docs/src/examples/pytorch_a2c_ppo.rst
@@ -1,0 +1,76 @@
+***********************************************
+Example with ikostrikov/pytorch-a2c-ppo-acktr
+***********************************************
+
+## Intro
+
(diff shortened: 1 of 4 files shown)


docs/src/examples/pytorch_a2c_ppo.rst
```diff
@@ -0,0 +1,116 @@
...
+
+.. code-block:: bash
+
+    eval.py
+
+.. note ::
+
+  The execution with Oríon does not require the added evaluation methods for
+  a validation set and could use the final training performance. However, for
+  sake if adhering to best practices, we create a validation set method in
```
bouthilx: sake if -> sake of

Breakend: fixed

docs/src/index.rst
```diff
@@ -13,7 +13,6 @@
    :caption: User's Guide
    :maxdepth: 1
 
-   user/pytorch
```
bouthilx: This is causing an error.

https://travis-ci.org/mila-udem/orion/jobs/418260617#L571

 I would keep this section there even though it is somehow a duplicate.

bouthilx:
Thanks!

PR #137, opened 2018-08-20
Feature/non blocking

bouthilx:
This is a rebase of #108 and #109 plus a few small fixes and additional unit-tests.

### Non-Blocking design 

Algorithms may generate duplicate trials over the course of optimisation, especially if there is many copy of an algorithm generating trials concurrently. To avoid waste of computations that would occur because of such duplication, we need to ensure it is not possible to register two identical trials. 

### Trials history

To ensure reproducibility, the history of all trials generated should be logged so that we can trace back what was the training data of the algorithm when it generated a given trial.

### Non-Blocking design 
- Experiments can now only register one trial at a time
- Producer will never generate trials already existing. Note: there is a default value of 100 attempts. If an algorithm generates duplicate trials a 100 times, a `RuntimeError` will be raised.

### Trials history
- Experiment now have a trials_history attribute, used to mark new trials with the history used to train the algorithm that generated them. Note: Algorithms that do not train such as RandomSearch still benefits from this history logging, since recapitulating the trial history would also increment the randomstate similarly and thus reproduce the same new trial.
- Trials now have a parents attribute, which refers to the trials used to train the algorithm that generated them.


src/orion/core/worker/producer.py
```diff
@@ -24,32 +25,58 @@ class Producer(object):
...
-        log.debug("### Convert them to `Trial` objects.")
-        new_trials = list(map(lambda data: format_trials.tuple_to_trial(data, self.space),
-                              new_points))
-
-        log.debug("### Register to database: %s", new_trials)
-        self.experiment.register_trials(new_trials)
+        sampled_points = 0
+        n_attemps = 0
+
+        while sampled_points < self.pool_size and n_attemps < self.max_attemps:
```
DatCorno: There's a typo here, it's spelled `attempts`.

src/orion/core/worker/producer.py
```diff
@@ -24,32 +25,58 @@ class Producer(object):
...
+            n_attemps += 1
+            log.debug("### Suggest new ones.")
```
DatCorno: Shouldn't this be changed to `Suggest new trial`?

DatCorno: I feel like this could be changed to something more explicit like `Suggesting new trials` since it meant for debug purposes.

DatCorno: And here's another reason why : I thought this was suggesting trials, but it points.

src/orion/core/worker/producer.py
bouthilx: Oh indeed! Thanks

src/orion/core/worker/producer.py
bouthilx: Agreed!

src/orion/core/worker/trial.py
```diff
@@ -203,6 +204,26 @@ def __str__(self):
 
     __repr__ = __str__
 
+    def _fetch_all_ancestors(self, current_ancestors):
```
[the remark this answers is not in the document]
bouthilx: I agree it's confusing. I'll see how I can rework it.

src/orion/core/worker/trials_history.py
```diff
@@ -11,18 +11,23 @@
 
 
 # pylint:disable=protected-access,too-few-public-methods
-class TrialsHistory:
+class TrialsHistory(object):
```
DatCorno: @bouthilx why do we need to explicitly inherit from `object` here?

PR #138, opened 2018-08-24
Add conda recipe

abergeron:
This doesn't build the packages on any CI, but once this PR is in the actual building is very easy.  We just need to figure out where they will be uploaded.

I've had to fix up the versioneer config a little bit, but this change will mean that all future release tags will have to start by `v`.

If that is unacceptable, there are other options.


bouthilx:
Oh, thanks! I'll try it out.

You will need to rebase on develop. We don't merge anything but releases into master.

abergeron:
Ah ok, I'll do that.

PR #139, opened 2018-08-25
Change URLs with mila-udem to epistimio

bouthilx:
Why:

The repository of Oríon has been moved to Epistímio organization.


src/orion/core/__init__.py
```diff
@@ -30,7 +30,7 @@
     'tsirif': ('Christos Tsirigotis', 'tsirif@gmail.com'),
     'bouthilx': ('Xavier Bouthillier', 'xavier.bouthillier@gmail.com'),
 }
-__url__ = 'https://github.com/mila-udem/orion'
```
tsirif: (I am not sure why I cannot comment on something else apart from the changes, but I think what I am about to write is relevant here in this file and PR)

We should probably apply corresponding changes to the private-magic variables above this one, with respect to the `MILA` -> `Epistimio` change.

bouthilx: Indeed. I did it in Kleio but forgot here. I'll add this and rebase.

bouthilx: Done

bouthilx: @tsirif Ping :)

nouiz: @tsirif Ping :)

bouthilx:
PR rebased, waiting for approval.

src/orion/core/__init__.py
```diff
@@ -23,6 +23,7 @@
 __version__ = VERSIONS['version']
 __license__ = 'BSD-3-Clause'
 __author__ = u'Epistímio Team'
+__author_short__ = u'Epistímio Team'
```
tsirif: @bouthilx what is the need for this? Shall we name it without the ` Team` in front?

bouthilx: I named it `Epistímio Team` because it was `Oríon Team`, but I do prefer `Epistímio` indeed.

DatCorno: Same here, I preter `Epistímio`

PR #140, opened 2018-08-25
Drop support for py34 and add py37

bouthilx:
Why:

Python 3.4 is getting old and now entered "security fixes only mode".

https://www.python.org/dev/peps/pep-0429/#id4
3.4.0 final: March 16, 2014

https://www.python.org/downloads/release/python-349/
> Python 3.4.9
> Release Date: 2018-08-02
>
> Python 3.4.9 was released on August 2nd, 2018.
>
> Python 3.4 has now entered "security fixes only" mode, and as such the
> only changes since Python 3.4.7 are security fixes. Also, Python 3.4.9
> has only been released in source code form; no more official binary
> installers will be produced.

> There are no specific plans for the next release of Python 3.4.
> However, the expectation is that Python 3.4.10 will be released in March
> of 2019, and this will be the final release of Python 3.4.

How:

Python 3.7 is supported as-is. Only modification needed is the version
for CI tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,22 +8,36 @@ matrix:
   include:
     - env: TOXENV=flake8
       python: 3.7
+      dist: xenial
+      sudo: true
     - env: TOXENV=pylint
       python: 3.7
+      dist: xenial
+      sudo: true
     - env: TOXENV=doc8
       python: 3.7
+      dist: xenial
+      sudo: true
     - env: TOXENV=packaging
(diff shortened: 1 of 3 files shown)


bouthilx:
As was deciding in meeting, we will drop pythoh3.4 and add support for 3.7. I would merge it if @ReyhaneAskari or @tsirif can give their approval for the last modifications. (Merge request is currently blocked)

PR #141, opened 2018-08-27
Add recipe for conda packages.

abergeron:
I might add on to this PR the config required to automatically build packages or do that in a follow-up PR.

This modifies the dependencies to move pytest-runner to test dependencies and change the versioneer config to use `v` as the tag prefix.


bouthilx:
Looks like travis is now running only the packaging test. Could it be that `jobs` isn't playing nicely with `matrix` in `.travis.cfg`? That would explain the drop of coverage.

abergeron:
The loss of coverage is from before this so it still doesn't explain it.  However no running the tests is not a good option, so I'll try to fix that.

Do you mind if I convert everything to the jobs thing and dropped the matrix?

bouthilx:
I don't mind if it's not significantly more verbose.

abergeron:
It shouldn't be.  I'll do that tomorrow.

.travis.yml
```diff
@@ -3,35 +3,33 @@ language: python
...
     - env: TOXENV=pylint
       python: 3.6
     - env: TOXENV=doc8
       python: 3.6
-    - env: TOXENV=packaging
-      python: 3.6
     - env: TOXENV=docs
       python: 3.6
-    - env: TOXENV=py34
+    - stage: test
```
bouthilx: Oh! :heart:

abergeron: This also makes it run each group completely before moving on to the next.  If a part of the group fails, it skips the following groups.

abergeron:
I've made a new commit, but it doesn't show up here.  This is strange.

bouthilx:
Wow the organization of tests on Travis is so nice, I love it!

abergeron:
This looks ok now.  I'll do a clean up of the history.

To get the package upload working there is a bit of config remaining:

 - [ ] Create a conda channel
 - [ ] Create an API token for uploading:
Something like this:
```
$ anaconda login
$ anaconda auth -c -n travis --max-age 307584000 --url https://anaconda.org/USERNAME/PACKAGENAME --scopes "api:write api:read"
```
- [ ] Set up the token as a Travis Secure Variable with the name ANACONDA_TOKEN: 
https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

bouthilx:
Alright, I'll take care of this today and check the boxes when done. Thank you very much!

bouthilx:
I think I've done all that was necessary. $ANACONDA_TOKEN is set in travis with the token I created for organisation epistimio channel orion on anaconda.

abergeron:
The only thing left to do would be to merge this and make a tag to test the process.

The test tag doesn't have to be a release tag (vX.XX), you can use anything.

bouthilx:
@tsirif Ping :)

abergeron:
I've rebased for python 3.7 update.

setup.cfg
```diff
@@ -3,7 +3,7 @@ VCS=git
 style=pep440-post
 versionfile_source=src/orion/core/_version.py
 versionfile_build=orion/core/_version.py
-tag_prefix=
+tag_prefix=v
```
tsirif: What is the difference in the convention here?

.travis.yml
```diff
@@ -1,42 +1,51 @@
...
+      env: TOXENV=py37
       python: 3.7
       dist: xenial
       sudo: true
-    - env: TOXENV=demo_random
+    - <<: *test
+      env: TOXENV=demo_random
+      python: 3.6
+    - stage: packaging
+      env: TOXENV=packaging
```
tsirif: The naming of toxenv is probably bad here, `packaging` is supposed to be a pretest.

bouthilx:
Which CI issue? All checks have passed.

setup.cfg
bouthilx: I don't remember exactly the source of the problem, but from what I remember it's about conda which is looking for only X.Y.Z format and fails if we provide vX.Y.Z format without explicitly telling it in the tag_prefix. 
@abergeron Am I wrong?

.travis.yml
bouthilx: The following test is a deployment to conda, I guess he wanted to bundle packaging and conda upload together in post-test stage. @abergeron How is the deployment working? Shouldn't it only upload when tested with the master branch?

setup.cfg
abergeron: The difference is that the previous config is wrong since you actually use the `v` prefix for tags.  It has not been a problem before because pip "fixes" the version and strips the v.  My packaging pipeline is less flexible and includes the v.

.travis.yml
abergeron: I always thought that packaging was to be done last, but if you want it as a pretest, we can move it.

As for the conda packages, since packaging and (conditional) upload are bundled it might be better to put it last.

bouthilx: What is the condition for the upload? If we are to keep conda packaging and upload bundled I would keep the other packaging test at this stage for coherence.

abergeron: It will upload the package if it is building a tag.

bouthilx: Oh ok, just checked the docs about [TRAVIS_TAG](https://github.com/Epistimio/orion/pull/141/files#diff-00041ed7660060aca02d68ff8d10fbd7R17)? I didn't know this was how it gets set. I remember you telling me about testing the build with a tag.

abergeron: This builds the package on every run, but only uploads for tags.

So what I meant is that it would be good once this is merge to add some tag in order to test that the upload process works correctly.

bouthilx: We don't need to wait for a merge. I made a new branch in the repo and tagged it to test the upload. 

Branch:
https://github.com/Epistimio/orion/tree/feature/conda

Tag:
https://github.com/Epistimio/orion/releases/tag/test-conda-upload

Travis test:
https://travis-ci.org/Epistimio/orion/builds/442688743

bouthilx: @abergeron It seems anaconda is not installed properly.

> ./conda/conda_build.sh: line 20: anaconda: command not found

https://travis-ci.org/Epistimio/orion/jobs/442688753#L2681

abergeron: Yep, anaconda-client was missing, I fixed that.

abergeron: You can force push a new commit id for the test-conda-upload tag to make sure it works, or create a new one.

bouthilx: I made a new tag with commit cfbe4ee:
https://github.com/Epistimio/orion/releases/tag/test-conda-upload-1.1

Travis test:
https://travis-ci.org/Epistimio/orion/builds/442800952

abergeron: It worked and the packages are on ananconda.org: 

https://anaconda.org/epistimio/orion

Although you might want to delete those since they essentially from a random commit and not a real release.

bouthilx: Yes thanks. I was waiting to install conda and test that the available packages are fine. On my Ubuntu system at least.

bouthilx: There seems to be a problem with the packages on conda. I can't find it.

```bash
$ conda install -c epistimio orion
Fetching package metadata .............

PackageNotFoundError: Packages missing in current channels:
            
  - orion
```

```bash
$ conda search -c epistimio --override-channels
Loading channels: done

PackagesNotFoundError: The following packages are not available from current channels:

  - *
```

Could it be because package name is specified as `orion` rather than `orion.core` in the conda [config](https://github.com/Epistimio/orion/pull/141/files#diff-dbb380fc11ea56e32b4687905343af66R4)?

abergeron: It works for me.

abergeron: Oh I get it somehow, there aren't any packages for linux-64.  This is only config missing, which is probably because the build packages are for that and it skips the conversion, which makes us skip the upload.

I'll have to fix that.

bouthilx: I updated tag test-conda-upload

Travis test:
https://travis-ci.org/Epistimio/orion/builds/444765160

conda/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  host:
+    - python {{ python }}
+    - setuptools
+    - pytest-runner
+  run:
+    - python
+    - numpy
+    - scipy
+    - pyyaml
+    - pymongo >=3
```
bouthilx: I think `gitpython` is missing here. When I install `orion` with conda I get 

```bash
pkg_resources.DistributionNotFound: The 'gitpython' distribution was not found and is required by orion.core
```
There's no way to automatically reuse the list from setup.py I guess? 

Also, can we add a test before the deployment to make sure such missing dependencies does not happen again in the future?

One final thing, can we list the package as orion.core to stay consistent with PiPy?

abergeron: No there is no way to sync with setup.py. 

I am testing that the produced package imports fine, but we might do some more tests to make sure it actually works.  I'm not sure which though.

bouthilx: Just running `orion --help` after installation with conda would at least verify the dependencies.

bouthilx: @abergeron Any updates on this? I guess you were very busy with the IVADO classes last week. Would you have time to fix the requirements issue and add a test this week?

abergeron: Maybe, but I have the git tutorial to give, so probably not until Friday.  If I find some time before that, I'll do it.

Otherwise, I don't mind if you want to do it to get it done faster.

bouthilx: Oh, sure, I forgot about the tutorial. I'll reserve my time for #137 and #128 for now. Thanks for you help!

abergeron:
I've fixed the issues with missing packages and also rebased on the current develop branch.

This should now be ready for merging.

bouthilx:
Thanks! Running last tests and then merging.

Tag:
https://github.com/Epistimio/orion/releases/tag/test-conda-upload-1.3

Travis test:
https://travis-ci.org/Epistimio/orion/builds/451109669

abergeron:
The one thing I'm ambivalent about is that it seems to create packages with version 0, which would seem to indicate that it doesn't pickup any tags.

However, it seems that the develop branch does indeed not have any tags, so maybe it's normal.  I don't know.  All the previous experience I have with this was from projects that don't have separate master and develop branches.

PR #145, opened 2018-09-04
Feature/visualization interface

DatCorno:
This PR requires merging of #135 

### Visualization
There is two parts to the visualization module: the analysers and the plotters. They come as a pair, the first one gathering and organizing the data for the latter to process. As the names suggest, the analysers will use the trials of an experiment to create an analysis of different types (as simple as a single value or more complex typing). On the other hand, the plotters are consumers of analysis. They recover the outputs of these analysers and create visual representation of that data.

### Analysers
The analysers follow a simple interface given by the `BaseAnalyser` class inside `orion.viz.analysers.base`. This base class provides the following abstract method `analyse` which performs the analysis. This method can take an optional argument `of_type` which specifies which type of analysis to return, in cases where the analyser can return multiple type of analysis.
The type of analysis provided by the analyser are obtain through the `available_analysis` property that can be overridden by child classes.
The analysers have access to the current experiment to analyse, the completed trials for this experiment and the space of hyper-parameters to work on.
There are two analysers currently available : a basic one called `commandline` which only returns the id of the best trial for the experiment. The second one is the LPI analysis.

### Plotters
The plotters also follow a simple interface: `BasePlotter` inside `orion.viz.plotters.base`. This base class requires the override of the `plot` method. This method takes the object of `Analysis` type stored inside the plotter and uses it to create some visual representation (bar plot, line plot, pie charts, etc.)
The type(s) of analysis each plot can operate on is given through the `required_analysis` property.
Each plotter has access to the analysis, as well as the file formats to use to save the representation.

There is no constraints to what kind of library the plotters can use for their representation. They simply need to make sure the correct wrapper is available for them to use. These wrappers serve two purposes: they make the calls to library specific function (like `title` for `pyplot`) and they also help prevent conflicts for the same kind of plots under different libraries (like a bar plot for `panda` andn for `matplotlib`.)


bouthilx:
Merge of #140 will drop failing py34 tests.

bouthilx:
@DatCorno You can rebase to fix some of the tests.

bouthilx:
[Review Vizualisation Interface](https://trello.com/c/dZCVnXPD/3-review-vizualisation-interface)

bouthilx:
I moved the branch to the main repo to archive it ([feature/visualization_interface](https://github.com/Epistimio/orion/tree/feature/visualization_interface)). After discussion we agreed that we do not have enough time to support the visualizations yet so we archive the branch rather than merging and having a dead-limb like submodule in the framework.

PR #134, opened 2018-08-10
Adapt documentation of the core installation

bouthilx:
Why:

Based on issue #132 it looks like the title of the section is
misleading. Also the PyPI link was not added now that we have one.

How:

Change title to `installation` and explain how to install the `core`
alone, plus write additional doc on how to install plugins.

PR #150, opened 2018-10-12
Tutorial

ReyhaneAskari:
WIP

### DIFF
--- a/docs/src/user/pytorch.rst
+++ b/docs/src/user/pytorch.rst
@@ -5,8 +5,11 @@ Simple example
 Installation and setup
 ======================
 
+In this tutorial you will run a very simple MNIST example in pytorch using Oríon.
+
 First, install Oríon follwing :doc:`/installing` and configure the database (:doc:`/database`).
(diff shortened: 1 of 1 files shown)


bouthilx:
@ReyhaneAskari You can make a branch on Epistimio:orion so that readthedocs automatically build the docs. It makes it very convenient for reviews. By the way, you should probably take PR #134 as a starting point, I should merge it soon actually.

bouthilx:
Sorry, I don't understand what is the problem. #134 is going to be merged in develop, and so is this one, #150. I don't see where master comes in?

ReyhaneAskari:
Oh I'm sorry. So in your previous comment you meant instead of having a branch on my own repo, I create a branch on Epistimio:orion?

bouthilx:
Oh ok I see the source of confusion now. You can push this branch to epistimio:orion yes. 

So, two things:
1. You should rebase on top of #134
2. Then push your rebased branch to epistimio:orion as something like `doc/tutorial`.

PR #151, opened 2018-10-24
Fix mongodb count deprecation

bouthilx:
Why:

The method count() is now deprecate in pymongo.

How:

Use count_documents() instead.

### DIFF
--- a/src/orion/core/io/database/mongodb.py
+++ b/src/orion/core/io/database/mongodb.py
@@ -229,7 +229,10 @@ class MongoDB(AbstractDB):
 
         """
         dbcollection = self._db[collection_name]
-        return dbcollection.count_documents(filter=query if query else {})
+        if hasattr(dbcollection, 'count_documents'):
+            return dbcollection.count_documents(filter=query if query else {})
+
+        return dbcollection.count(filter=query)
 
     def remove(self, collection_name, query):
         """Delete from a collection document[s] which match the `query`.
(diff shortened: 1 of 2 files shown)


tsirif:
There should probably be a version requirement in`setup.py` too in this.

DatCorno:
Agreed, we should only really stop supporting an older version if some features are not available.

src/orion/core/io/database/mongodb.py
```diff
@@ -229,6 +229,9 @@ def count(self, collection_name, query=None):
 
         """
         dbcollection = self._db[collection_name]
+        if hasattr(dbcollection, 'count_documents'):
+            return dbcollection.count_documents(filter=query if query else {})
+
         return dbcollection.count(filter=query)
```
mnoukhov: this could be in an `else` statement, but that's a personal choice

bouthilx: Actually pylint would complain because the else is useless. An alternative would be

```
if something:
    count = 
else:
    count = 
return count
```

bouthilx:
@tsirif Any comments on the support for old versions of pymongo?

bouthilx:
As was agreed during last [meeting](https://discourse-epistimio.mila.quebec/t/2018-10-31-plan/61) that we support previous versions of pymongo, I will now merge this PR.

PR #152, opened 2018-10-31
Add Strategies with Trials history

bouthilx:
This is a rebase of #137 (itself rebased on #108 and #109) and #128 plus a few small fixes and additional unit-tests.

### Non-Blocking design 

Algorithms may generate duplicate trials over the course of optimisation, especially if there is many copy of an algorithm generating trials concurrently. To avoid waste of computations that would occur because of such duplication, we need to ensure it is not possible to register two identical trials. 

### Trials history

To ensure reproducibility, the history of all trials generated should be logged so that we can trace back what was the training data of the algorithm when it generated a given trial.

### Strategies

To improve parallel optimization, algorithms should take into account non-completed trials to ensure they do not suggest again similar trials.

### Non-Blocking design 
- Experiments can now only register one trial at a time
- Producer will never generate trials already existing. Note: there is a default value of 100 attempts. If an algorithm generates duplicate trials a 100 times, a `RuntimeError` will be raised.

### Trials history
- Experiment now have a trials_history attribute, used to mark new trials with the history used to train the algorithm that generated them. Note: Algorithms that do not train such as RandomSearch still benefits from this history logging, since recapitulating the trial history would also increment the randomstate similarly and thus reproduce the same new trial.
- Trials now have a parents attribute, which refers to the trials used to train the algorithm that generated them.

### Strategies
- New ParallelStrategy object in `orion.core.worker.strategy`. Note that this should be taken out of the core eventually, when we refactor Algorithm along with Strategy and Scheduler.
- Producer now contains a copy of the algorithm, called the `naive_algorithm`. The latter is copied from the original algorithm at each update and then updated on fake trials generated by the ParallelStrategy object.
- A new collection, `lying_trials` is used for the fake trials. Those are registered through `experiment.register_lie()`.


tests/unittests/core/worker/test_trials_history.py
```diff
@@ -0,0 +1,75 @@
...
+def test_added_children_with_ancestors():
+    """Verify that children with ancestors are added to history"""
+    trials_history = TrialsHistory()
+    trials = [DummyTrial(i, []) for i in range(3)]
+    trials_history.update(trials)
+    assert trials_history.children == [0, 1, 2]
+
+    trials = [DummyTrial(i, [trials[i % 3].id]) for i in range(3, 6)]
+    trials_history.update(trials)
+    assert len(set(trials_history.children) & set([3, 4, 5])) == 3
```
mnoukhov: shouldn't this just be `set(trials_history.children) == set([3, 4, 5])`

bouthilx: Not exactly. I wanted to avoid testing that the duplicates are properly removed, which is the role of `test_discarded_children`. That's why I only test that (3, 4, 5) are present in `trials_history.children`, without testing if other trials are still there or discarded properly.

mnoukhov: Ahh, makes sense!

bouthilx:
[Review Parallel Strategies](https://trello.com/c/uzIDiniM/1-review-parallel-strategies)

bouthilx:
@tsirif or @DatCorno would you mind taking a look before we merge?

PR #155, opened 2019-02-07
Feature/generic cmdline

DatCorno:
This pull request adds the class `CmdlineParser` partially taken from the Kleio project and adapted to suit the needs of Oríon. It is a simple commandline parser.

### DIFF
--- a/src/orion/core/io/cmdline_parser.py
+++ b/src/orion/core/io/cmdline_parser.py
@@ -120,7 +120,6 @@ class CmdlineParser(object):
 
         Examples
         --------
-
         Positional and named arguments:
 
         >>> parser = CmdlineParser()
@@ -180,7 +179,8 @@ class CmdlineParser(object):
 
         return self.arguments
 
-    def _key_to_arg(self, arg):
+    @staticmethod
+    def _key_to_arg(arg):
         if len(arg) > 1:
             return "--" + arg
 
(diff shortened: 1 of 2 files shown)


src/orion/core/io/cmdline_parser.py
```diff
@@ -0,0 +1,247 @@
...
+        >>> parser.parse('python --args value1 value2'.split(' '))
+        OrderedDict([('_pos_0', 'python'), ('args', ['value1', 'value2'])])
+
+        Named argument defined with `=`:
+
+        >>> parser.parse('python --arg=value'.split(' '))
+        OrderedDict([('_pos_0', 'python'), ('arg', 'value')])
+
+        """
+        if not commandline or self._already_parsed:
```
bouthilx: That's likely to be confusing. I remember having a use-case for that in Kleio because of the branching but I don't see any use case for Oríon. I'd say if already_parse raise RuntimeError('Already parsed').

tests/unittests/core/io/test_cmdline_parser.py
```diff
@@ -0,0 +1,126 @@
...
+    command = "python script.py some pos args " \
+              "--with args --and multiple args --plus --booleans "
+
+    configuration = cmdline_parser.parse(command.split(" "))
+    assert " ".join(cmdline_parser.format(configuration)) == command.strip(" ")
+
+    cmdline_parser = CmdlineParser()
+    command2 = "python script.py some pos args " \
+              "--with something --and multiple args --plus --booleans --to update"
+    branch_configuration = cmdline_parser.parse(command2.split(" "))
```
bouthilx: This would now raise RuntimeError, that would be the good unit test to catch this.

PR #153, opened 2018-10-31
Tutorial

ReyhaneAskari:
To do list:
- [x] Installation
- [ ] Experiment configuration
- [x]  EVC (Search Space, ..)
- [ ] Results Retrieval
- [ ] Debugging
- [ ] Algorithms 

Links need to be fixed. I have to review it too.

PR #146, opened 2018-09-04
Add node property

christopher-beckham:
In reference to: https://github.com/Epistimio/orion/issues/143

(I made a PR earlier but closed it since I forked off the wrong branch. And then git was being problematic on the MILA servers! It should be good now.)

PR #133, opened 2018-08-10
Improve docs

dmitriy-serdyuk:
Fixes #132

PR #131, opened 2018-08-08
Fix/database.rst

ReyhaneAskari:
Same PR as https://github.com/mila-udem/orion/pull/129, except for changing the base to develop.

PR #156, opened 2019-02-08
Release candidate v0.1.1rc1

bouthilx:
## Breaking Changes
- Add git commit detection for version control #115
- Add Strategies with Trials history #152 
## Additional New Features
- Add `hunt --worker-trials` argument #127 
- Add node property to experiment for the EVC system. #146 
- Add a generic commandline parser #155
## Bug Fixes
- Set verbosity >= 2 to DEBUG #123
- Fix github links #126
## Deprecation
- Drop support for py34 and add py37 #140 
- Fix mongodb deprecation #151
## Other Improvements
- Add recipe for conda packages #141 
## Documentation Improvements
- Updated instructions for Atlas MongoDB. Added instructions for non-root user #131 
- Clarify doc section name #133 
- Add RL example #136 
- Fix README utf-8 encoding #117
- Changed URLs from mila-udem to epistimio #139 
- Add simple tutorial #153


tsirif:
I have a question regarding #155: Is this going to substitute the existing parsing/formatting logic for a particular trial consumption? If yes, my concern is that it implies that a certain way for the user's program arguments to be parsed, e.g. that all positional arguments preceed keyword ones. If no, what is the motivation apart from it being a more clear way to do things?

tsirif:
#126 refers to mila-udem github user, not Epistimio

bouthilx:
Concerning #155. Yes it assumes there is no positional arguments after the named arguments. Note that it is not possible to have positional arguments after named arguments unless the positional arguments are subcommands. One thing to note however, is that the parser would not fail to reproduce a commandline when the assumption is not satisfied. It will be able to parse it and reformat it as-is. The issue is that the configuration parsed (`parser.arguments`) will be incoherent with the semantics.

Beside this possible inconsistency, I believe this parsing is the simplest way to support change detection at the level of semantics (`"--arg1 one --arg2 two" == "--arg2 two --arg1 one"`). But mostly, it also makes it simpler to provide a unified architecture that supports both commandline calls and a python API. That's because `parser.arguments` will be treated basically the same as `arguments` passed to a function with the python API.

bouthilx:
Yes #139 changes to `epistimio` the links fixed in #126 .

PR #157, opened 2019-02-14
Relax constraint for git repository

bouthilx:
Why:

We decided that Oríon would crash if the script being executed is not part of a code version control (only support git for now). This was to ensure reproducibility, but this constraint is troublesome for users using a different VCS system or none at all. It's not Oríon's role to enforce a VCS on users. They could at least use Oríon without `git`, they will only loose the automatic code change detection.

How:

Log a warning that code change cannot be detected instead of raising an error.

### DIFF
--- a/src/orion/core/io/resolve_config.py
+++ b/src/orion/core/io/resolve_config.py
@@ -254,10 +254,10 @@ def fetch_user_repo(user_script):
     dir_path = os.path.dirname(os.path.abspath(user_script))
     try:
         git_repo = git.Repo(dir_path, search_parent_directories=True)
-    except git.exc.InvalidGitRepositoryError as e:
+    except git.exc.InvalidGitRepositoryError:
         git_repo = None
-        raise RuntimeError('Script {} should be in a git repository.'.format(
-            os.path.abspath(user_script))) from e
+        logging.warning('Script %s is not in a git repository. Code modification '
+                        'won\'t be detected.', os.path.abspath(user_script))
     return git_repo
 
 
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -583,12 +583,12 @@ class Experiment(object):
         # Just overwrite everything else given
         for section, value in config.items():
             if section not in self.__slots__:
-                log.warning("Found section '%s' in configuration. Experiments "
-                            "do not support this option. Ignoring.", section)
+                log.info("Found section '%s' in configuration. Experiments "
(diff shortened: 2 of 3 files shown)


bouthilx:
@ReyhaneAskari @tsirif @mnoukhov @DatCorno I believe the code is fine, but the change is worth at least a short discussion. Do you have any objections to this _relaxation_ of the code version constraint?

ReyhaneAskari:
I also think we shouldn't force VCS. But is this going to be logged each time the user runs Orion for that specific project? If so, isn't it going to be troublesome too?

bouthilx:
That could be annoying indeed. We could add a variable `ignore-vcs-warning` in global config so that user can turn it off. It would be obviously `False` by default.

ReyhaneAskari:
Can't we detect if we have already showed this to the user?

bouthilx:
For a single execution, using `warnings.warn` instead of `logger.warning` would do the trick. Across multiple executions, we can restrict the warning message to new experiment registration. That means re-executing an existing experiment would silence the warning.

I'm concerned that silencing it automatically would raise the risk that the user misses the warning on the first time and ignores that the code change is not detected, possibly leading to incoherent experiment history (using different code version). I feel forcing the user to manually set the `ignore-vcs-warning` would be a way of making sure that the user is aware of the risk when the warning is silenced. This feature is unique to Oríon after all, as far as I know, so users may not be aware of it or would not expect it.

What do you think would be best?

ReyhaneAskari:
How about setting it as a flag?

bouthilx:
What do you mean as a flag? I feel we may talk about the same thing. What I meant is something that could be set by env var or config, just like database config.

env var
```bash
ORION_IGNORE_VCS_WARNING=True orion hunt ...
```

global config
```yaml
ignore_vcs_warning: True
database:
    ...
```

ReyhaneAskari:
oh I see. Then that looks good to me too!

bouthilx:
Looking at `resolve_config.py`, I realize I would need to push the config as an argument into all the calls from `fetch_metadata` down to `fetch_user_repo`. I don't think it is worth the trouble, considering that we should have shortly a refactored configuration which will make it possible to access the config from `fetch_metadata` without passing it as an argument.

I would suggest we create an issue about the flag and implement it when the refactored config is ready. Meanwhile some users may have warnings, but I guess that is no very problematic for now.

bouthilx:
@ReyhaneAskari @tsirif @mnoukhov @DatCorno Ping :)

PR #158, opened 2019-02-18
Realign db config

bouthilx:


bouthilx:
@ReyhaneAskari Same here :)

PR #159, opened 2019-02-20
Raise error if branching in non interactive shell

bouthilx:
Why:

The comand prompt raises OSError when the shell is non-interactive, but
for some reason it sometimes infinitely raises the OSError and
reactivate the command prompt.

How:

Raise error before starting the prompt if we detect that we are in a non
interactive shell.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -634,8 +634,17 @@ class Experiment(object):
         """
         experiment_brancher = ExperimentBranchBuilder(conflicts, branching_configuration)
 
-        if not experiment_brancher.is_resolved or experiment_brancher.auto_resolution:
+        needs_manual_resolution = (not experiment_brancher.is_resolved or
+                                   experiment_brancher.auto_resolution)
+
+        if needs_manual_resolution:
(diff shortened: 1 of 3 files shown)


bouthilx:
@DatCorno Would you have time to review this?

src/orion/core/worker/experiment.py
```diff
@@ -633,8 +633,17 @@ def _branch_config(self, conflicts, branching_configuration):
         """
         experiment_brancher = ExperimentBranchBuilder(conflicts, branching_configuration)
 
-        if not experiment_brancher.is_resolved or experiment_brancher.auto_resolution:
+        needs_manual_resolution = (not experiment_brancher.is_resolved or
+                                   experiment_brancher.auto_resolution)
```
DatCorno: If auto_resolution is True, wouldn't that mean that the user does not want the manual resolution?

bouthilx: This is confusing indeed. If using auto-resolution, then the conflict resolver prompts so that the user confirms that the changes are fine.

DatCorno: I see

PR #164, opened 2019-02-26
 Add Linearize transform and linear requirement

bouthilx:
Why:

Some algorithms works better if the space is linearized. If the algo is
sampling from another internal prior for instance, the algorithm would
need to lookup on the space to see if the dimension has a logarithmic
distribution in order to adapt the internal prior.

How:

Turn all dimensions to `real`, and those with logarithmic distribution
are then further linearized by taking the log.

Note:

Also modifies `unpack_point` and `pack_point` to make them idempotent.

### DIFF
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -949,7 +949,7 @@ def pack_point(point, space):
 
 
 def unpack_point(point, space):
-    """Flatten `point` in `space` and convert it to a 1D `numpy.ndarray`.
+    """Flatten `point` in `space` and convert it to a 1D list.
 
     This function is deprecated and will be removed in v0.2.0. Use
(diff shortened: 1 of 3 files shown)


bouthilx:
@tsirif Would have some time to take a look at this?

bouthilx:
Little update. It seems the bayesian optimizer built on top of scikit-optimizer is not sampling from log-uniform but rather uniform. So it turns out this linearization would be very helpful to improve the bayesian optimizer as well.

bouthilx:
[Add Linearisation](https://trello.com/c/Qh7JkloX/75-add-linearisation)

Thomsch:
Forcing Travis to migrate to .com

PR #167, opened 2019-03-06
Add seeding to algorithm and fix parallel strategies

bouthilx:
# Add seeding to algorithm

### Why:

Some (if not many) algorithms are random and should be _seedable_.

### How:

Add the method `seed_rng()` which may seed the algorithm's interval
random number generator. For flexibility, the implementation of the
seeding mechanism is speficic to each algorithm and there is no standard
`RNG` object made public.

# Fix parallel strategies

### Why:

The methods `Strategy.observe` were expecting results as `Result` type,
but were being passed results as dictionaries.

### How:

Since Algorithm.observe gets results as dictionaries, we keep the same
for `Strategy` for coherence.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -54,8 +54,6 @@ class DumbAlgo(BaseAlgorithm):
         rval = []
         while len(rval) < num:
             value = self.possible_values[min(self._index, len(self.possible_values) - 1)]
-            print(self.possible_values)
-            print(self._index, value)
             self._index += 1
             rval.append(value)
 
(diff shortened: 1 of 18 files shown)


tests/conftest.py
```diff
@@ -35,19 +36,27 @@ def __init__(self, space, value=5,
...
         rval = []
         while len(rval) < num:
-            if len(self.possible_values) > 1:
-                value = self.possible_values.pop(0)
-            else:
-                value = self.possible_values[0]
-
+            value = self.possible_values[min(self._index, len(self.possible_values) - 1)]
+            print(self.possible_values)
+            print(self._index, value)
```
mnoukhov: are these print statements here on purpose?

bouthilx: No indeed, thanks!

PR #170, opened 2019-03-07
Set pool-size to 1 and strategy to MaxParallelStrategy by default

bouthilx:
### Why:

Users are often testing sequentially, and even if running in parallel we
now have parallel strategies which makes pool-size > 1 useless unless
the algorithm have strategies different then those implemented in Oríon.
This is such an unlikely scenario that it is better to set the default
value to 1 to avoid confusion.

## Set default value of strategy to MaxParallelStrategy

Why:

Since pool-size is now 1 by default, we want a default strategy that
will handle parallel optimization properly. MaxParallelStrategy is the
most pessimistic one and the one which will mostly favor exploration.
Until we have benchmarkings showing benefits of the different parallel
strategies this ones seems to be the most reasonable one.


bouthilx:
Someone may want to use NoParallelStrategy for benchmarking purposes.

mnoukhov:
Fair! all good then!

PR #171, opened 2019-03-20
Fix --worker-trials

bouthilx:
## Why

The loop in the worker was either producer or consumer, which made
worker count each production of trials as 1 trial execution. In other
words, --worker-trials 1 would execute no trial if it produced trials.

### DIFF
--- a/src/orion/core/worker/__init__.py
+++ b/src/orion/core/worker/__init__.py
@@ -21,6 +21,24 @@ from orion.core.worker.producer import Producer
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


bouthilx:
@DatCorno Ready for review

src/orion/core/worker/__init__.py
```diff
@@ -21,6 +21,24 @@
 log = logging.getLogger(__name__)
 
 
+def reserve_trial(experiment, producer):
+    """Reserve a new trial, or produce and reserve a trial if non are available."""
```
DatCorno: Typo here.

PR #174, opened 2019-04-22
Add port option for db backend

bouthilx:
Why:

SSH tunnels require a way to set ports dynamically from outside the yaml
config of orion, using env vars for instance.

How:

Add the env var ORION_DB_PORT and add tests to make sure it properly
overwrites the db host value in MongoDB.

### DIFF
--- a/docs/src/examples/cluster.rst
+++ b/docs/src/examples/cluster.rst
@@ -1,0 +1,134 @@
+**************
+Running on HPC
+**************
+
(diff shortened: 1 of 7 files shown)


bouthilx:
Doc available here: https://orion.readthedocs.io/en/feature-dbport/examples/cluster.html

bouthilx:
@DatCorno Can you please have a look at this PR. That'll be useful for the new cluster which does not have access to internet on compute nodes.

PR #177, opened 2019-05-29
Feature/fidelity

DatCorno:
Add a Fidelity class.

### DIFF
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -667,6 +667,11 @@ class Fidelity(Dimension):
         """Do not do anything."""
         raise NotImplementedError
 
+    def __repr__(self):
+        """Represent the object as a string."""
+        return "{0}(name={1})".format(self.__class__.__name__, self.name)
+
+
 
 class Space(OrderedDict):
     """Represents the search space.
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -630,6 +630,11 @@ class Categorical(Dimension):
 class Fidelity(Dimension):
     """Fidelity `Dimension` for representing multi-fidelity.
 
+    Fidelity dimensions are not optimized by the algorithms. If it supports multi-fidelity, the
+    algorithm will select a fidelity level for which it will sample hyper-parameter values to
+    explore a low fidelity space. This class is used as a place-holder so that algorithms can
+    discern fidelity dimensions from hyper-parameter dimensions.
+
     Attributes
     ----------
     name : str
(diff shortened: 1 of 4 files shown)


tests/unittests/algo/test_space.py
```diff
@@ -117,6 +117,13 @@ def test_no_default_value(self):
         dim = Dimension('yolo', 'uniform', -3, 4)
         assert dim.default_value is None
 
+    def test_no_prior(self):
+        """Test the giving a null prior defaults prior_name to `None`."""
```
bouthilx:
```suggestion
        """Test that giving a null prior defaults prior_name to `None`."""
```

tests/unittests/algo/test_space.py
```diff
@@ -454,6 +461,38 @@ def test_cast_bad_category(self):
...
 
 
+class TestFidelity(object):
+    """Test methods of a Fidelity object."""
+
+    def test_simple_instance(self, seed):
+        """Test Fidelity.__init__."""
+        dim = Fidelity('epoch')
+
+        assert str(dim) == "Fidelity(name=epoch, prior={None: (), {}}, " + \
```
bouthilx:
```suggestion
        assert str(dim) == "Fidelity(name=epoch)"
```

src/orion/algo/space.py
```diff
@@ -621,6 +627,46 @@ def get_category(value):
...
+
+    Attributes
+    ----------
+    name : str
+    type : str
+
+    """
+
+    def __init__(self, name):
+        """Search space dimension that can represent a fidelity value.
```
bouthilx:
```suggestion
        """Fidelity dimension that can represent a fidelity level.
```

src/orion/algo/space.py
```diff
@@ -621,6 +627,46 @@ def get_category(value):
         return casted_point
 
 
+class Fidelity(Dimension):
+    """Subclass of `Dimension` for representing real parameters.
```
bouthilx:
```suggestion
    """Fidelity `Dimension` for multi-fidelity.
    
    Fidelity dimensions are not optimized by the algorithms. If it supports multi-fidelity, the algorithm will select a fidelity
    level for which it will sample hyper-parameter values to explore a low fidelity space. This class is used as a 
    place-holder so that algorithms can discern fidelity dimensions from hyper-parameter dimensions.
```

src/orion/algo/space.py
```diff
@@ -621,6 +627,46 @@ def get_category(value):
...
+
+    Attributes
+    ----------
+    name : str
+    type : str
+
+    """
+
+    def __init__(self, name):
```
bouthilx: You'll need ```pylint:disable=super-init-not-called```

src/orion/algo/space.py
bouthilx: Part of the doc is missing from your commit

PR #178, opened 2019-06-03
Add error message for missing configuration

DatCorno:


bouthilx:
Unit-tests? :)

src/orion/core/cli/base.py
```diff
@@ -67,8 +68,11 @@ def parse(self, argv):
 
     def execute(self, argv):
         """Execute main function of the subparser"""
-        args, function = self.parse(argv)
-        function(args)
+        try:
+            args, function = self.parse(argv)
+            function(args)
+        except NoConfigurationError as ex:
+            print("No commandline configuration found for new experiment.")
```
bouthilx:
```suggestion
            print("Error: No commandline configuration found for new experiment.")
```

bouthilx:
Do we have a test for building without the commandline when exp-name is found in db?

DatCorno:
I don't think so. I'll add it to the PR.

bouthilx:
@DatCorno Missing test for experiment found in DB and rebase. Then ready for merge.

tests/unittests/core/io/test_experiment_builder.py
```diff
@@ -287,3 +297,32 @@ def test_build_from_config_hit(old_config_file, create_db_instance, exp_config,
     assert exp.pool_size == exp_config[0][0]['pool_size']
     assert exp.max_trials == exp_config[0][0]['max_trials']
     assert exp.algorithms.configuration == exp_config[0][0]['algorithms']
+
+
+@pytest.mark.usefixtures("clean_db", "null_db_instances", "with_user_tsirif")
+def test_build_without_config_hit(old_config_file, create_db_instance, exp_config, script_path):
```
bouthilx: I meant testing if the experiment is properly build if `cmdargs` is like the one at l.268, but the experiment name is found in the db.

DatCorno: This is what this does, doesn't it? It first makes sure we have a correctly built experiment inside the database, so that we have a hit when looking at it by name. Then, we try to build an experiment through the command line with no argument.

bouthilx: but 'user_args' should be empty.

bouthilx: My bad, you were indeed building it again only using the exp name at second stage.

PR #181, opened 2019-06-07
Handle bad results in trials

bouthilx:
Why:

If the user pass bad results (missing objective, None objective, etc),
Oríon will later crash in mysterious ways for the users. We should
detect this early and raise an insightfull error for the user.

How:

Make the result attribute private and test in results.setter if the
provided results are valid.

### DIFF
--- a/src/orion/core/worker/trial.py
+++ b/src/orion/core/worker/trial.py
@@ -148,13 +148,13 @@ class Trial(object):
         allowed_types = ('integer', 'real', 'categorical')
 
     __slots__ = ('experiment', '_id', '_status', 'worker',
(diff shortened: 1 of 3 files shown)


src/orion/core/worker/trial.py
```diff
@@ -202,6 +202,26 @@ def __str__(self):
...
+    def results(self):
+        """List of results of the trial"""
+        return self._results
+
+    @results.setter
+    def results(self, results):
+        """Verify results before setting the property"""
+        objective = self._fetch_one_result_of_type('objective', results)
+
+        print(type(objective))
```
DatCorno: Is the print suppose to be there?

bouthilx: Nop!

PR #182, opened 2019-06-07
Fix null comparison in EphemeralDB

bouthilx:
Why:

Sometimes a field is None which causes comparisons in queries to crash.
Comparisons with a None field should return False instead of crashing.


bouthilx:
Closed and reopened because the link between travis and github was somehow broken.

PR #183, opened 2019-06-12
Add timeout for connection to Database

DatCorno:
Why: we were not failing fast when dealing with a non-existing
database. Now there is a default value for a timeout when connecting.

### DIFF
--- a/src/orion/core/io/database/mongodb.py
+++ b/src/orion/core/io/database/mongodb.py
@@ -78,9 +78,8 @@ class MongoDB(AbstractDB):
     """
 
     def __init__(self, host='localhost', name=None,
-                 port=None, username=None, password=None):
+                 port=None, username=None, password=None, server_timeout=5000):
         """Init method, see attributes of :class:`AbstractDB`."""
-        self.options['authSource'] = name
         self.uri = None
 
         if port is not None:
@@ -89,6 +88,9 @@ class MongoDB(AbstractDB):
             port = pymongo.MongoClient.PORT
 
         super(MongoDB, self).__init__(host, name, port, username, password)
+
+        self.options['serverSelectionTimeoutMS'] = server_timeout
+        self.options['authSource'] = name
 
     @mongodb_exception_wrapper
     def initiate_connection(self):
--- a/src/orion/core/io/database/mongodb.py
+++ b/src/orion/core/io/database/mongodb.py
@@ -78,7 +78,7 @@ class MongoDB(AbstractDB):
     """
 
     def __init__(self, host='localhost', name=None,
-                 port=None, username=None, password=None, server_timeout=5000):
+                 port=None, username=None, password=None, serverSelectionTimeoutMS=5000):
         """Init method, see attributes of :class:`AbstractDB`."""
         self.uri = None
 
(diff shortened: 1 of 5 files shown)


src/orion/core/io/database/mongodb.py
```diff
@@ -80,7 +80,7 @@ class MongoDB(AbstractDB):
     def __init__(self, host='localhost', name=None,
                  port=None, username=None, password=None):
         """Init method, see attributes of :class:`AbstractDB`."""
-        self.options = {'authSource': name}
+        self.options['authSource'] = name
```
bouthilx: You need to init the Parent class before, otherwise `self.options` does not exist yet.

src/orion/core/io/resolve_config.py
```diff
@@ -127,6 +127,9 @@ def fetch_default_options():
         for _, key, default_value in env_vars:
             default_config[signifier][key] = default_value
 
+    if 'serverSelectionTimeoutMS' not in default_config['database']:
+        default_config['database']['serverSelectionTimeoutMS'] = 5000
```
bouthilx: I would put this instead in the default values of the init of MongoDB database since this option is specific to this backend.

tests/unittests/core/mongodb_test.py
```diff
@@ -107,12 +107,19 @@ def test_overwrite_partial_uri(self, monkeypatch):
...
+                           password='pass')
         # reinit connection does not change anything
         orion_db.initiate_connection()
         orion_db.close_connection()
         assert MongoDB() is orion_db
 
+    def test_change_server_timeout(self):
+        """Test that the server timeout is correctly changed."""
+        orion_db = MongoDB(username='user', password='pass', name='orion_test', server_timeout=6000)
+        assert 6000 in str(orion_db)
```
bouthilx: What we would like to test here, is that server_timeout actually affects the timeout. You can test with 1000 and verify that `time.time() - start` is smaller than 2000 and try again with 5000 to make sure it gets adapted.

src/orion/core/io/database/mongodb.py
```diff
@@ -90,6 +89,9 @@ def __init__(self, host='localhost', name=None,
 
         super(MongoDB, self).__init__(host, name, port, username, password)
```
bouthilx: The authentification errors are because of `initiate_connection()` in the Parent class. You can solve this by passing `authSource` and `serverSelectionTimeoutMS` directly to `__init__` to get them set in `options`. By the way, I would use `serverSelectionTimeoutMS` in the interface instead of `server_timeout` to keep it coherent with `pymongo` and because it is specific to the MongoDB backend.

tests/unittests/core/mongodb_test.py
DatCorno: Woopsie

tests/unittests/core/mongodb_test.py
```diff
@@ -107,12 +107,20 @@ def test_overwrite_partial_uri(self, monkeypatch):
 
     def test_singleton(self):
         """Test that MongoDB class is a singleton."""
-        orion_db = MongoDB(username='user', password='pass', name='orion_test')
+        orion_db = MongoDB('mongodb://localhost',
+                           port=27017, name='orion', username='lala',
+                           password='pass')
```
bouthilx: Same error here. username should be ```user```.

tests/unittests/core/mongodb_test.py
bouthilx:
```suggestion
        assert '6000' in str(orion_db)
```

PR #184, opened 2019-06-14
Add `working-dir` argument and its support

DatCorno:
The `working-dir` argument aims to offer the possibility to run the execution of the script inside a permanent directory instead of a temporary one inside the `Consumer`.


src/orion/core/worker/consumer.py
```diff
@@ -52,9 +53,15 @@ def __init__(self, experiment):
...
         self.script_path = experiment.metadata['user_script']
-        self.tmp_dir = os.path.join(tempfile.gettempdir(), 'orion')
-        os.makedirs(self.tmp_dir, exist_ok=True)
+
+        if experiment.working_dir:
+            self.dir = os.path.abspath(experiment.working_dir)
+        else:
+            self.dir = os.path.join(tempfile.gettempdir(), 'orion')
+
+        os.makedirs(self.dir, exist_ok=True)
```
bouthilx: You can push this inside `WorkingDir`.

tests/functional/demo/test_demo.py
```diff
@@ -241,6 +241,36 @@ def test_stress_unique_folder_creation(database, monkeypatch, tmpdir, capfd):
...
+@pytest.mark.usefixtures("null_db_instances")
+def test_working_dir_argument_cmdline(database, monkeypatch, tmp_path):
+    """Check that a permanent directory is used instead of tmpdir"""
+    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
+    orion.core.cli.main(["hunt", "-n", "allo", "--working-dir", str(tmp_path / "test"),
+                         "--max-trials", "2", "--config", "./database_config.yaml",
+                         "./black_box.py", "-x~uniform(-50,50)"])
+
+    exp = list(database.experiments.find({'name': 'allo'}))[0]
+    assert exp['working_dir'] == str(tmp_path / "test")
```
bouthilx: You should test that the folder does not exist before the call to `main` and delete it at the end of the unit-test.

tests/functional/demo/test_demo.py
```diff
@@ -241,6 +241,36 @@ def test_stress_unique_folder_creation(database, monkeypatch, tmpdir, capfd):
...
+@pytest.mark.usefixtures("null_db_instances")
+def test_working_dir_argument_config(database, monkeypatch):
+    """Check that a permanent directory is used instead of tmpdir"""
+    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
+    orion.core.cli.main(["hunt", "-n", "allo", "--max-trials", "2",
+                         "--config", "./working_dir_config.yaml", "./black_box.py",
+                         "-x~uniform(-50,50)"])
+
+    exp = list(database.experiments.find({'name': 'allo'}))[0]
+    assert exp['working_dir'] == "/tmp/orion/test"
```
bouthilx: Same here

tests/unittests/core/experiment.yaml
```diff
@@ -469,6 +480,7 @@
   # orion's configuration
   pool_size: 2
   max_trials: 1000
+  working_dir:
```
bouthilx: Does it crash if `working_dir` is not in yaml?

tests/unittests/core/test_experiment.py
```diff
@@ -316,8 +321,25 @@ def test_good_set_before_init_no_hit(self, random_dt, database, new_config):
...
+        """Check if working_dir is correctly changed."""
+        exp = Experiment(new_config['name'])
+        exp.configure(new_config)
+        assert exp._init_done is True
+        database.experiments.update_one({'name': 'supernaekei', 'metadata.user': 'tsirif'},
+                                        {'$set': {'working_dir': './'}})
+        found_config = list(database.experiments.find({'name': 'supernaekei',
+                                                       'metadata.user': 'tsirif'}))
+
+        print(found_config)
```
bouthilx: leftover print :)

tests/unittests/core/utils/test_working_dir.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Collection of tests for :mod:`orion.core.utils.working_dir`."""
+import os
+
+from orion.core.utils.working_dir import WorkingDir
+
+
```
bouthilx: You should add a test for when folder exists and temp is True to make sure it does not crash.

src/orion/core/worker/consumer.py
```diff
@@ -65,10 +72,12 @@ def consume(self, trial):
         :type trial: `orion.core.worker.trial.Trial`
 
         """
-        log.debug("### Create new temporary directory at '%s':", self.tmp_dir)
-        with tempfile.TemporaryDirectory(prefix=self.experiment.name + '_',
-                                         dir=self.tmp_dir) as workdirname:
-            log.debug("## New temp consumer context: %s", workdirname)
+        log.debug("### Create new directory at '%s':", self.dir)
+        temp_dir = self.experiment.working_dir is None
+        prefix = self.experiment.name + "_"
```
bouthilx: It should be ```'{experiment.name}_{trial.id}'``` otherwise there will be clash of dirname for different trials when `working_dir` is set. Note that the trailing `_` is only required when `working_dir` is not set since it is appended with a random string.

src/orion/core/worker/consumer.py
DatCorno: It'll get call once instead of being call every time the `WorkingDir` is created, which is every time `consume` is called. So, while it'll be call multiple times for nothing, because this doesn't change throughout execution.

bouthilx: Agreed, but os.makedirs is called every time for the subfolder anyway. Also, this is optimizing milliseconds in favor of code simplicity, while there is bottlenecks that are significantly greater here (most particularly the HP sampling).

tests/unittests/core/experiment.yaml
DatCorno: Well, `orion_config.yaml` doesn't have it and the tests pass fine ;)

bouthilx: I mean in the DB most specifically ;)

src/orion/core/utils/working_dir.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        self.working_dir = working_dir
+        self._temp = temp
+        self._suffix = suffix
+        self._prefix = prefix
+        self._tmpdir = None
+
+    def __enter__(self):
+        """Create the a permanent directory or a temporary one."""
+        if not self._temp:
```
bouthilx: `suffix` is not used.

src/orion/core/worker/consumer.py
DatCorno: Alright, I'll move it.

tests/unittests/core/utils/test_working_dir.py
```diff
@@ -0,0 +1,35 @@
...
+    shutil.rmtree(tmp_path / "hi")
+
+
+def test_create_temp_dir(tmp_path):
+    """Check if a temporary directory is created."""
+    with WorkingDir(tmp_path, prefix="hi") as w:
+        assert os.path.exists(w)
+
+    assert not os.path.exists(w)
```
bouthilx: You should add a test for suffix now. :)

src/orion/core/worker/consumer.py
bouthilx: Actually, you can simply **re**move it. :upside_down_face:

src/orion/core/utils/working_dir.py
```diff
@@ -24,16 +24,22 @@ def __init__(self, working_dir, temp=True, suffix=None, prefix=None):
...
-        self.working_dir = working_dir
         self._temp = temp
         self._suffix = suffix
         self._prefix = prefix
         self._tmpdir = None
 
+        if working_dir:
+            self.working_dir = os.path.abspath(working_dir)
+        else:
+            self.working_dir = os.path.join(tempfile.gettempdir(), 'orion')
```
bouthilx: :thinking: Maybe leave the if-else in the consumer. The `os.makedirs` can be remove entirely as the dir will be created in `__enter__` anyway.

tests/unittests/core/test_checks.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Collection of tests for :mod:`orion.core.cli.checks`."""
+from orion.core.cli.checks.presence import PresenceStage
```
bouthilx: Leak from another branch? :)

src/orion/core/utils/working_dir.py
DatCorno: I didn't need to move it then ;)

tests/unittests/core/test_checks.py
DatCorno: Definitely!

src/orion/core/utils/working_dir.py
bouthilx: Yes, my comment was only about `os.makedirs` and it was about **re**moving it. :) (https://github.com/Epistimio/orion/pull/184#discussion_r295933318)

DatCorno: Semantics :roll_eyes:

src/orion/core/worker/consumer.py
```diff
@@ -65,12 +69,12 @@ def consume(self, trial):
         :type trial: `orion.core.worker.trial.Trial`
 
         """
-        log.debug("### Create new directory at '%s':", self.dir)
-        temp_dir = self.experiment.working_dir is None
+        log.debug("### Create new directory at '%s':", self.working_dir)
+        temp_dir = self.working_dir is None
         prefix = self.experiment.name + "_"
-        suffix = "{}".format(trial.id) if self.dir else ""
+        suffix = "{}".format(trial.id) if self.working_dir else ""
```
bouthilx: This will always be true. We can simply set it as `suffix = trial.id`, it's ok if the trial id is included in the tmp dir.

src/orion/core/worker/consumer.py
```diff
@@ -65,12 +69,12 @@ def consume(self, trial):
         :type trial: `orion.core.worker.trial.Trial`
 
         """
-        log.debug("### Create new directory at '%s':", self.dir)
-        temp_dir = self.experiment.working_dir is None
+        log.debug("### Create new directory at '%s':", self.working_dir)
+        temp_dir = self.working_dir is None
```
bouthilx: This will always be True as well.

DatCorno: How will this always be true?

bouthilx: Always be False, sorry, because `self.working_dir` is always set in the init. Before fixing this, you should add a unit test that detects if not passing --working-dir has the intended effect of using a tmp dir.

src/orion/core/worker/consumer.py
```diff
@@ -65,10 +69,14 @@ def consume(self, trial):
...
 
         """
-        log.debug("### Create new temporary directory at '%s':", self.tmp_dir)
-        with tempfile.TemporaryDirectory(prefix=self.experiment.name + '_',
-                                         dir=self.tmp_dir) as workdirname:
-            log.debug("## New temp consumer context: %s", workdirname)
+        log.debug("### Create new directory at '%s':", self.working_dir)
+        temp_dir = self.working_dir is None
+        prefix = self.experiment.name + "_"
+        suffix = "{}".format(trial.id)
```
bouthilx:
```suggestion
        suffix = trial.id
```

bouthilx:
Just a little rebase and good to merge. :)

PR #188, opened 2019-06-26
Create command line utility to check if the database is correctly set up.

DatCorno:


src/orion/core/cli/checks/operations.py
```diff
@@ -0,0 +1,94 @@
...
+            raise CheckError(str(ex))
+
+        if not len(result):
+            raise CheckError("Expected 'value', received nothing.")
+
+        return "Success", ""
+
+    @register_check(_Checks.checks, "check if database supports count operation... ")
+    def check_count(self):
+        """Check if database supports read operation."""
```
bouthilx: count

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+        if 'database' not in config:
+            return "Skipping", "No default configuration found for database."
+
+        self.db_config = config['database']
+
+        return "Success", ""
+
+    @register_check(_Checks.checks, "check if oríon's environment variables have been set... ")
+    def check_environment_vars(self):
+        """Check if database supports delete operation."""
```
bouthilx: idem

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+        config = self.builder.fetch_env_vars()
+
+        config = config['database']
+        names = ['type', 'name', 'host']
+
+        for name in names:
+            if name not in config or config[name] is None:
+                return "Skipping", "Missing {} environment variable.".format(name)
```
bouthilx: Not all of them may be set. The set should support partial definition using env vars.

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+                raise CheckError("No database found in configuration file.")
+
+            config = config['database']
+            names = ['type', 'name', 'host']
+
+            for name in names:
+                if name not in config or config[name] is None:
+                    raise CheckError("Missing {} inside configuration.".format(name))
+        except CheckError as ex:
+            if self.db_config is not None:
```
bouthilx: We should fallback to default values and test using them instead of failing the tests. Maybe the user set his db according to the default values and do not need to provide any config.

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+        self.db_config = None
+
+    @staticmethod
+    def checks():
+        """Return the registered checks."""
+        return _Checks.checks
+
+    @register_check(_Checks.checks, "check if oríon's default options are present... ")
+    def check_default_config(self):
+        """Check if database supports delete operation."""
```
bouthilx: docstring

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+
+            if 'database' not in config:
+                raise CheckError("No database found in configuration file.")
+
+            config = config['database']
+            names = ['type', 'name', 'host']
+
+            for name in names:
+                if name not in config or config[name] is None:
+                    raise CheckError("Missing {} inside configuration.".format(name))
```
bouthilx: Same here, a config with partial definition can be valid.

src/orion/core/cli/checks/presence.py
```diff
@@ -0,0 +1,100 @@
...
+
+from orion.core.utils.decorators import register_check
+from orion.core.utils.exceptions import CheckError
+
+
+class _Checks:
+    checks = []
+
+
+class PresenceStage:
```
bouthilx: There isn't much to test so far in the configuration. What would be useful is to print the current config after each step so that the user can see how global, env vars and local config file have precedence over the others.

src/orion/core/cli/checks/creation.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        self.p_stage = presence_stage
+        self.instance = None
+
+    @staticmethod
+    def checks():
+        """Return checklist."""
+        return _Checks.checks
+
+    @register_check(_Checks.checks, "Check if database of specified type can be created... ")
```
bouthilx: Why do you use a decorator for this? Why not just print it at the beginning of the method call?

src/orion/core/cli/test_db.py
```diff
@@ -0,0 +1,59 @@
...
+                print(status)
+
+                if status == "Skipping":
+                    print(msg)
+
+            stage.post_stage()
+
+    except CheckError as ex:
+        print("Failure")
+        print(ex)
```
bouthilx: There should be a cleanup method for each stage to make sure interrupting the checks in the middle would not lead to broken tests. The count test will fail if write test is executed twice before remove test is.

tests/unittests/core/cli/test_checks.py
```diff
@@ -91,14 +91,11 @@ def test_envvar_config_pass(monkeypatch, presence):
 
 def test_envvar_config_skip(monkeypatch, presence):
     """Check if test skips when missing an environment variable."""
```
bouthilx: when **no** env var

src/orion/core/cli/test_db.py
bouthilx: @DatCorno There is only this issue remaining. The tests should be more isolated, so that running the tests after interrupting between write and remove test does not break the count test.

PR #198, opened 2019-07-08
Add pydocstyle version to tox-flake8 deps

bouthilx:
Why:

The version of pydocstyle is not upper bounded in flake8. The new
version 4.0.0 of pydocstyle is not yet supported by flake8 and won't
ever be by flake8 3.5.0.

How:

Add pydocstyle == 3.0.0 to force lower version number than what is
automatically installed when installing flake8.


DatCorno:
I think you need to rebase.

PR #201, opened 2019-07-09
added log level for filelock

mirkobronzi:


bouthilx:
What about setting the logging level for filelock inside `orion.core.io.database.pickleddb` instead?

mirkobronzi:
> What about setting the logging level for filelock inside `orion.core.io.database.pickleddb` instead?

Yes, I think it makes sense. Just modified the code.

src/orion/core/io/database/__init__.py
```diff
@@ -13,6 +13,7 @@
    - :class:`orion.core.io.database.mongodb.MongoDB`
 
 """
+import logging
 from abc import abstractmethod, abstractproperty
```
bouthilx: You should switch those lines to make flake8 happy.

src/orion/core/io/database/__init__.py
```diff
@@ -264,3 +265,7 @@ class Database(AbstractDB, metaclass=SingletonFactory):
     """
 
     pass
+
+
+# set per-module log level
+logging.getLogger('filelock').setLevel('ERROR')
```
bouthilx: Wait, ERROR? Isn't that always going to log? Maybe we should add a unit-test for this logging. :)

mirkobronzi: it's the opposite actually..

given we cannot change the level in the module `filelock`, which is info, we change the required level for logging to error.

so, now, when the code runs on the line `logger.info`, info < error, and it does not print.

bouthilx: Oh yes, my bad! I got confused. :sweat_smile:

tests/unittests/core/test_pickleddb.py
```diff
@@ -263,6 +263,26 @@ def test_read_and_write_no_match(self, orion_db):
 
         assert loaded_config is None
 
+    def test_logging_when_getting_file_lock(self, caplog, orion_db):
+        """When logging.level is ERROR, there should be no logging."""
+        caplog.set_level(logging.INFO)
+        caplog.clear()
+        """Any operation will trigger the lock."""
```
bouthilx: This should be a comment instead of a string.

tests/unittests/core/test_pickleddb.py
```diff
@@ -263,6 +263,26 @@ def test_read_and_write_no_match(self, orion_db):
...
+        orion_db.read(
+            'experiments',
+            {'name': 'supernaedo2', 'metadata.user': 'dendi'})
+
+        assert 'acquired on orion_db.pkl.lock' in caplog.text
+
+        caplog.set_level(logging.ERROR)
+        caplog.clear()
+        """Any operation will trigger the lock."""
```
bouthilx: Idem

tests/unittests/core/test_pickleddb.py
mirkobronzi: done :)

tests/unittests/core/test_pickleddb.py
mirkobronzi: done :)

bouthilx:
Now needs rebase

PR #202, opened 2019-07-09
Add heartbeat to Trial

DatCorno:
When a trial is reserved, a heartbeat should be updated every N seconds so that we can detect if the trial is still reserved/running or lost. Currently, the wait time is configurable inside the monitoring thread, but we are waiting for the Configuration overhaul to implement it inside the configuration options.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -28,6 +28,7 @@ from orion.core.worker.primary_algo import PrimaryAlgo
 from orion.core.worker.strategy import (BaseParallelStrategy,
                                         Strategy)
 from orion.core.worker.trial import Trial
+from orion.core.worker.trial_monitor import TrialMonitor
 
 log = logging.getLogger(__name__)
 
@@ -245,7 +246,7 @@ class Experiment(object):
         # status meanwhile, read_and_write will fail, because query will fail.
         query = {'_id': selected_trial.id, 'status': selected_trial.status}
 
-        update = dict(status='reserved')
+        update = dict(status='reserved', heartbeat=datetime.datetime.utcnow())
 
         if selected_trial.status == 'new':
             update["start_time"] = datetime.datetime.utcnow()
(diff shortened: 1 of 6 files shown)


src/orion/core/worker/experiment.py
```diff
@@ -257,11 +258,12 @@ def reserve_trial(self, score_handle=None):
             selected_trial = self.reserve_trial(score_handle=score_handle)
         else:
             selected_trial = Trial(**selected_trial_dict)
+            TrialMonitor(self, selected_trial.id).start()
 
         return selected_trial
 
     def push_completed_trial(self, trial):
-        """Inform database about an evaluated `trial` with results.
+        """Inform database about an evaluated `trial` with resultlts.
```
bouthilx: Not sure this is an improvement :rofl:

tests/unittests/core/worker/test_trial_monitor.py
```diff
@@ -0,0 +1,112 @@
...
+    trials = exp.fetch_trials({'_id': trial.id, 'status': 'reserved'})
+
+    assert trial.heartbeat != trials[0].heartbeat
+
+    data = {'status': 'interrupted'}
+    Database().write('trials', data, query=dict(_id=trial.id))
+
+    time.sleep(2)
+
+    trial_monitor.join()
```
bouthilx: Add a comment here to clarify that if it leaves it means the monitor stopped as expected.

src/orion/core/worker/experiment.py
DatCorno: I fixed it (here)[https://github.com/Epistimio/orion/pull/203/files#diff-6d901bae3fe0c9e74e137277f6fdf9b4L264] ahaha

bouthilx:
@DatCorno Needs rebase

bouthilx:
@DatCorno Needs rebase again :sweat_smile:

bouthilx:
@DatCorno Seems like the rebase was on another branch than current develop

PR #203, opened 2019-07-09
Indiana Jones and the Raiders of the Lost Trial

DatCorno:
This PR depends on #202.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -28,6 +28,7 @@ from orion.core.worker.primary_algo import PrimaryAlgo
 from orion.core.worker.strategy import (BaseParallelStrategy,
                                         Strategy)
 from orion.core.worker.trial import Trial
+from orion.core.worker.trial_monitor import TrialMonitor
 
 log = logging.getLogger(__name__)
 
@@ -245,7 +246,7 @@ class Experiment(object):
         # status meanwhile, read_and_write will fail, because query will fail.
         query = {'_id': selected_trial.id, 'status': selected_trial.status}
 
-        update = dict(status='reserved')
+        update = dict(status='reserved', heartbeat=datetime.datetime.utcnow())
 
         if selected_trial.status == 'new':
             update["start_time"] = datetime.datetime.utcnow()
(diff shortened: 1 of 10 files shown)


src/orion/core/worker/experiment.py
```diff
@@ -257,9 +260,30 @@ def reserve_trial(self, score_handle=None):
...
+    def fix_lost_trials(self):
+        """Find lost trials and set them to interrupted.
+
+        A lost trial is defined as a trial whose heartbeat as not been updated since two times
+        the wait time for monitoring. This usually means that the trial is stalling or has been
+        interrupted in some way without its status being changed. This functions finds such
+        trials and set them as interrupted so they can be launched again.
+
+        """
+        query = {'experiment': self._id, 'status': {'$in': ['reserved']}}
```
```suggestion
        query = {'experiment': self._id, 'status': 'reserved'}
```

src/orion/core/worker/experiment.py
```diff
@@ -257,9 +260,30 @@ def reserve_trial(self, score_handle=None):
...
+
+        trials = self.fetch_trials(query)
+
+        for trial in trials:
+            difference = (datetime.datetime.utcnow() - trial.heartbeat).total_seconds()
+            if abs(difference) >= 120:
+                query['_id'] = trial.id
+                print(query)
```
bouthilx: print

src/orion/core/worker/experiment.py
bouthilx: I would add `'heartbeat': {'$lte': datetime.datetime.utcnow() - datetime.timedelta(seconds=60 * 2)}` to limit the number of trials loaded and then you can get rid of the `if` in the `for` loop.

Please add a `TODO` comment to make sure we don't forget to modify this part when we make the heartbeat configurable.

Also, you'll need to add the `$lte` command here: https://github.com/Epistimio/orion/blob/master/src/orion/core/io/database/ephemeraldb.py#L281

tests/unittests/core/test_experiment.py
```diff
@@ -626,14 +630,39 @@ def fake_handle(self, xxx):
...
+        hacked_exp.fix_lost_trials()
+
+        assert not hacked_exp.fetch_trials(exp_query)
+
+        exp_query.pop('status')
+
+        trial = hacked_exp.fetch_trials(exp_query)[0]
+
+        assert trial.status == 'interrupted'
+
```
bouthilx: I know other tests could fail if it was not the case anyway, but we should add one test to explicitly test that the non-lost trials are not set to interrupted.

bouthilx:
Ready for merge, you can rebase

PR #208, opened 2019-07-09
Add new cli command to pretty-print a tree of the experiments

DatCorno:
Depends on #207


bouthilx:
unit-tests :smile:

bouthilx:
Test is failing because you need to add the `pptree` dependency here as well: https://github.com/Epistimio/orion/blob/master/conda/meta.yaml#L25

src/orion/core/cli/list.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=eval-used,protected-access
```
bouthilx: You can remove this I believe

src/orion/core/utils/pptree.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
+# flake8: noqa
```
bouthilx: You need to disable pylint as well

PR #212, opened 2019-07-10
Add a functional test for user script config

bouthilx:
The use of a config file for a user was not tested in the functional tests. I added one to make sure we cover this important use-case.

### DIFF
--- a/tests/functional/demo/test_demo.py
+++ b/tests/functional/demo/test_demo.py
@@ -92,6 +92,51 @@ def test_demo(database, monkeypatch):
 
 
 @pytest.mark.usefixtures("clean_db")
+@pytest.mark.usefixtures("null_db_instances")
+def test_demo_with_script_config(database, monkeypatch):
+    """Test a simple usage scenario."""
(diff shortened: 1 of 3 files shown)


tests/functional/demo/test_demo.py
```diff
@@ -91,6 +91,51 @@ def test_demo(database, monkeypatch):
...
+                                                      'dx_tolerance': 1e-7}}
+    assert 'user' in exp['metadata']
+    assert 'datetime' in exp['metadata']
+    assert 'orion_version' in exp['metadata']
+    assert 'user_script' in exp['metadata']
+    assert os.path.isabs(exp['metadata']['user_script'])
+    assert exp['metadata']['user_args'] == ['--config', 'script_config.yaml']
+
+    trials = list(database.trials.find({'experiment': exp_id}))
+    assert len(trials) <= 15
```
DatCorno: `max-trials` is 100 but the script only has 15 trials?

bouthilx: Gradient descent is quite efficient. :)

PR #216, opened 2019-07-12
Handle `orion list --name` properly

bouthilx:
### Handle --name argument properly 

Why:

The argument --name was currently ignored. It should filter the name of the experiments

### Get rid of config file in `list` tests 

Why:

There is no need for a config file, what we need is to fix the DB instance in the fixture so that this one is used in all tests and not the global config or the current user running the tests.

### DIFF
--- a/tests/functional/commands/test_list_command.py
+++ b/tests/functional/commands/test_list_command.py
@@ -90,3 +90,43 @@ def test_three_exp(capsys, three_experiments):
 
     assert captured == " test_list_double┐\n                 └test_list_double_child\n \
 test_list_single\n"
+
+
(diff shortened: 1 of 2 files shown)


tests/functional/commands/test_list_command.py
```diff
@@ -6,29 +6,42 @@
...
 
 @pytest.fixture
 def no_experiment(database):
-    """Make sure there is no experiment."""
+    """Create and save a singleton for an empty database instance."""
     database.experiments.drop()
+    database.lying_trials.drop()
+    database.trials.drop()
+    database.workers.drop()
+    database.resources.drop()
```
DatCorno: Why drop everything else?

PR #217, opened 2019-07-14
Fix race conditions during algo updates

bouthilx:
### Fetch all completed trials instead of subset
Why:

When there is a high number of workers and communication gets slower,
it happens that fetches based on last trial's end time will miss some of
them. Since is no simple way of defining a backward internal to catch
them all, and since the actual cost of fetching all trials is relatively
cheap (which may most likely be less than 1000), it may be better to
simple fetch them all every time and filter locally.

How:

Add `__contains__` to TrialsHistory.
Move filtering in `Producer` which contains `TrialsHistory`.

### Bundle update calls in Producer for consistency
Why:

There can be modifications in trials status between the call to
`fetch_completed_trials` and `fetch_noncompleted_trials`, leading
to inconsistent states, where very recently completed trials are neither
in completed trials, nor noncompleted trials. This is problematic for
ASHA, because a promoted trial may be observed by the naive algorithm
while the base trial (non-promoted one) is lost between the 2 calls.
When this happens the naive algo crashes because the rungs are
inconsistent.

How:

Fetch all trials once and then pass the completed ones to
`update_algorithm` and pass the non completed ones to
`update_naive_algorithm`. This way the fetch is a single snapshot and
garantees consistency.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -87,7 +87,7 @@ class Experiment(object):
 
     __slots__ = ('name', 'refers', 'metadata', 'pool_size', 'max_trials',
                  'algorithms', 'producer', 'working_dir', '_init_done', '_id',
-                 '_node', '_last_fetched', '_storage')
+                 '_node', '_storage')
     non_branching_attrs = ('pool_size', 'max_trials')
 
     def __init__(self, name, user=None):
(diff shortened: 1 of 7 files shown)


bouthilx:
[Fetch all completed trials instead of subset](https://trello.com/c/RxwFiZa3/190-fetch-all-completed-trials-instead-of-subset)

PR #221, opened 2019-07-15
Legacy proto

Delaunay:


bouthilx:
[Create LegacyProtocol](https://trello.com/c/sOk2Sqti/129-create-legacyprotocol)

src/orion/core/worker/__init__.py
```diff
@@ -72,14 +71,15 @@ def workon(experiment, worker_trials=None):
         log.info("No trials completed.")
         return
 
-    best = Database().read('trials', {'_id': stats['best_trials_id']})[0]
+    # Database().read('trials', {'_id': stats['best_trials_id']})[0]
```
bouthilx: You can remove the commented line

src/orion/core/worker/__init__.py
```diff
@@ -72,14 +71,15 @@ def workon(experiment, worker_trials=None):
...
+    best = experiment.fetch_trials({'_id': stats['best_trials_id']})[0]
 
     stats_stream = io.StringIO()
     pprint.pprint(stats, stream=stats_stream)
     stats_string = stats_stream.getvalue()
 
     best_stream = io.StringIO()
-    pprint.pprint(best['params'], stream=best_stream)
+    pprint.pprint(best.params, stream=best_stream)
```
bouthilx: To keep the same output I would use `best.to_dict()['params']`.

src/orion/core/worker/consumer.py
```diff
@@ -86,17 +82,20 @@ def consume(self, trial):
...
                 self._consume(trial, workdirname)
+
         except KeyboardInterrupt:
             log.debug("### Save %s as interrupted.", trial)
             trial.status = 'interrupted'
-            Database().write('trials', trial.to_dict(),
-                             query={'_id': trial.id})
+
+            # FIXME: do we need to push all the attributes, or only pushing the status is enough ?
+            self.experiment.update_trial(trial, **trial.to_dict())
```
bouthilx: Only the status is enough.

src/orion/core/worker/consumer.py
```diff
@@ -86,17 +82,20 @@ def consume(self, trial):
...
             raise
         except RuntimeError:
             log.debug("### Save %s as broken.", trial)
             trial.status = 'broken'
-            Database().write('trials', trial.to_dict(),
-                             query={'_id': trial.id})
+
+            # FIXME: do we need to push all the attributes, or only pushing the status is enough ?
+            self.experiment.update_trial(trial, **trial.to_dict())
```
bouthilx: Only the status is enough.

src/orion/core/worker/experiment.py
```diff
@@ -106,8 +106,9 @@ def __init__(self, name, user=None):
         """
         log.debug("Creating Experiment object with name: %s", name)
         self._init_done = False
-        self._db = Database()  # fetch database instance
-        self._setup_db()  # build indexes for collections
+        self._protocol = StorageProtocol('legacy')
+        # self._db = Database()  # fetch database instance
```
bouthilx: You can remove the commented lines

src/orion/core/worker/experiment.py
```diff
@@ -122,8 +123,9 @@ def __init__(self, name, user=None):
         self.working_dir = None
         self.producer = {'strategy': None}
 
-        config = self._db.read('experiments',
-                               {'name': name, 'metadata.user': user})
+        # config = self._db.read('experiments',  {'name': name, 'metadata.user': user})
```
bouthilx: You can remove the commented line

src/orion/core/worker/experiment.py
```diff
@@ -170,7 +159,8 @@ def fetch_trials(self, query, selection=None):
         """
         query["experiment"] = self._id
 
-        trials = Trial.build(self._db.read('trials', query, selection))
+        # trials = Trial.build(self._db.read('trials', query, selection))
```
bouthilx: You can remove the commented line

src/orion/core/worker/experiment.py
```diff
@@ -209,15 +199,25 @@ def connect_to_version_control_tree(self, node):
...
         """Find *new* trials that exist currently in database and select one of
         them based on the highest score return from `score_handle` callable.
 
         :param score_handle: A way to decide which trial out of the *new* ones to
            to pick as *reserved*, defaults to a random choice.
         :type score_handle: callable
+        :param _depth: recursion depth only used for logging purposes can be ignored
         :return: selected `Trial` object, None if could not find any.
         """
+        # log.debug(f'{">" * _depth} Reserving trial with (score: {score_handle})')
```
bouthilx: You can bring back the logging

src/orion/core/worker/experiment.py
```diff
@@ -243,24 +246,31 @@ def reserve_trial(self, score_handle=None):
...
         update = dict(status='reserved', heartbeat=datetime.datetime.utcnow())
 
         if selected_trial.status == 'new':
             update["start_time"] = datetime.datetime.utcnow()
 
-        reserved = self._db.write('trials', query=query, data=update)
+        # Query on status to ensure atomicity. If another process change the
+        # status meanwhile, update will fail, because query will fail.
+        # This relies on the atomicity of document updates.
+        # /!\ This is not atomic if you are using monogo db replica set !
```
bouthilx: Following our discussion I believe you can remove this comment. Since we only rely on writes to handle race-conditions, and writes are atomic on the primary set, then this cannot fail. The only issue is about reads on replica sets which may be incoherent, but that is fine because either the status is marked as reserved and then the trial simply isn't reserved at this moment, or it is marked as available (new, interrupted, ...) and the reservation will fail. Either case, we are fine.

src/orion/core/worker/experiment.py
```diff
@@ -243,24 +246,31 @@ def reserve_trial(self, score_handle=None):
...
+
+        # reserved = self._db.write('trials', query=query, data=update)
```
bouthilx: You can remove the commented out code

src/orion/core/worker/experiment.py
```diff
@@ -282,7 +292,10 @@ def fix_lost_trials(self):
         for trial in trials:
             query['_id'] = trial.id
             log.debug('Setting lost trial %s status to interrupted...', trial.id)
-            updated = self._db.write('trials', {'status': 'interrupted'}, query)
+
+            updated = self._protocol.update_trial(trial, status='interrupted', where=query)
+            # updated = self._db.write('trials', {'status': 'interrupted'}, query)
```
bouthilx: You can remove the commented out code

src/orion/core/worker/experiment.py
```diff
@@ -325,7 +338,8 @@ def register_lie(self, lying_trial):
         lying_trial.status = 'completed'
         lying_trial.end_time = datetime.datetime.utcnow()
 
-        self._db.write('lying_trials', lying_trial.to_dict())
+        # FIXME
```
bouthilx: You should document what needs to be fixed here.

src/orion/core/worker/experiment.py
```diff
@@ -448,7 +462,8 @@ def is_broken(self):
 
         """
         query = {'experiment': self._id, 'status': 'broken'}
-        num_broken_trials = self._db.count('trials', query)
+        # num_broken_trials = self._db.count('trials', query)
```
bouthilx: You can remove the commented code

src/orion/core/worker/experiment.py
```diff
@@ -551,17 +566,17 @@ def configure(self, config, enable_branching=True, enable_update=True):
             # This will raise DuplicateKeyError if a concurrent experiment with
             # identical (name, metadata.user) is written first in the database.
 
-            self._db.write('experiments', final_config)
+            self._protocol.create_experiment(final_config)
+            # self._db.write('experiments', final_config)
```
bouthilx: Idem

src/orion/core/worker/experiment.py
```diff
@@ -570,7 +585,8 @@ def configure(self, config, enable_branching=True, enable_update=True):
             # `db.write()`, thus seamingly breaking  the compound index
             # `(name, metadata.user)`
             final_config.pop("name")
-            self._db.write('experiments', final_config, {'_id': self._id})
+            self._protocol.update_experiment(self, **final_config)
+            # self._db.write('experiments', final_config, {'_id': self._id})
```
bouthilx: idem

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.storage.base -- Generic Storage Protocol
+=============================================================================
```
```suggestion
:mod:`orion.storage.base -- Generic Storage Protocol
====================================================
```

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+class BaseStorageProtocol:
+    """Implement a generic protocol to allow Orion to communicate using
+    different storage backend
+    """
+
+    def create_experiment(self, config):
+        """Insert a new experiment inside the database"""
+        raise NotImplementedError()
+
+    def update_experiment(self, experiment: 'Experiment', fields: Optional[dict] = None,
```
bouthilx: Can we drop typing for consistency?

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+                          where: Optional[dict] = None, **kwargs):
+        """Update a the fields of a given trials
+
+        Parameters
+        ----------
+        experiment: Experiment object to update
```
bouthilx: Numpy docstring style is
```
name: type
     description
```

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+        fields: Optional[dict] a dictionary of fields to update
+        where: constraint experiment must respect
+        kwargs: a dictionary of fields to update
+
+        Example
+        -------
+        .. code-block:: python
```
bouthilx: Numpy docstring style is
```
Examples
--------
>>> query = {'status': 'RUNNING'}
>>> BaseStorageProtocol.update_experiment(experiment, fields=query)
some output ...
>>> BaseStorageProtocol.update_experiment(experiment, status='RUNNING')
```
By the way, experiments do not have status so the example is a bit misleading. ;)

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+    def fetch_experiments(self, query):
+        """Fetch all experiments that match the query"""
+        raise NotImplementedError()
+
+    def register_trial(self, trial):
+        """Create a new trial to be executed"""
+        raise NotImplementedError()
+
+    def reserve_trial(self, *args, **kwargs):
+        """Select a pending trial and book it for the executor"""
```
bouthilx: and *reserve* it?

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+        raise NotImplementedError()
+
+    def fetch_trials(self, query, *args, **kwargs):
+        """Feetch all the trials that match the query"""
```
```suggestion
        """Fetch all the trials that match the query"""
```

src/orion/storage/base.py
Delaunay: >>> make pylint fail

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+        raise NotImplementedError()
+
+    def update_trial(self, trial: 'Trial', fields: Optional[dict] = None,
+                     where: Optional[dict] = None, **kwargs) -> 'Trial':
+        """Update a the fields of a given trials
```
```suggestion
        """Update the fields of a given trials
```

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Parameters
+        ----------
+        trial: Trial object to update
```
bouthilx: docstring format

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+        where: constraint trial must respect
+        kwargs: a dictionary of fields to update
+
+        returns the updated trial
+
+        Example
+        -------
+        .. code-block:: python
+            query = {'status': 'RUNNING'}
+            BaseStorageProtocol.update_trial(trial, fields=query)
```
bouthilx: query? :)

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+            BaseStorageProtocol.update_trial(trial, status='RUNNING')
+
+        """
+        raise NotImplementedError()
+
+    def retrieve_result(self, trial, results_file=None, **kwargs):
+        """Read the results from the trial and append it to the trial object"""
```
bouthilx: That's confusing. The results are already in the trial and we append it to the trial again? Do you mean update the local trial object with results from DB or File and update DB accordingly?

bouthilx: The arguments should be documented here

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+        raise NotImplementedError()
+
+
+# pylint: disable=too-few-public-methods,abstract-method
+class StorageProtocol(BaseStorageProtocol, metaclass=Factory):
+    """Storage protocol is a generic way of allowing Orion to interface with different storage.
+    MongoDB, track, cometML, MLFLow, etc...
+
+    Protocol('track', uri='file://orion_test.json')
```
bouthilx: Examples?

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+    Protocol('legacy', experiment=experiment)
+    """
+
+    pass
+
+
+# pylint: disable=too-few-public-methods
+class ReadOnlyStorageProtocol(object):
+    """Read-only view on a database.
```
bouthilx: The documentation should be adapted to StorageProtocol

src/orion/storage/base.py
```diff
@@ -0,0 +1,126 @@
...
+    valid_attributes = {"fetch_trials", "fetch_experiments"}
+
+    def __init__(self, protocol):
+        """Init method, see attributes of :class:`AbstractDB`."""
+        self._protocol = protocol
+
+    def __getattr__(self, attr):
+        """Get attribute only if valid"""
+        if attr not in self.valid_attributes:
+            raise AttributeError("Cannot access attribute %s on view-only experiments." % attr)
```
```suggestion
            raise AttributeError("Cannot access attribute %s on ReadOnlyStorageProtocol." % attr)
```

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,113 @@
...
+from orion.core.io.database import Database
+from orion.core.worker.trial import Trial
+from orion.storage.base import BaseStorageProtocol
+
+
+class Legacy(BaseStorageProtocol):
+    """Legacy protocol, forward most request to experiment"""
+
+    def __init__(self, uri=None):
+        """INIT METHOD"""
```
bouthilx: No docstring for `__init__`

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,113 @@
...
+        self._db.ensure_index('trials', 'status')
+        self._db.ensure_index('trials', 'results')
+        self._db.ensure_index('trials', 'start_time')
+        self._db.ensure_index('trials', [('end_time', Database.DESCENDING)])
+
+    def create_experiment(self, config):
+        """See :func:`~orion.storage.BaseStorageProtocol.create_experiment`"""
+        return self._db.write('experiments', config)
+
+    def update_experiment(self, experiment: 'Experiment', fields: Optional[dict] = None,
```
bouthilx: Why should we support both `fields` and `kwargs`? `fields` can easily be avoided with `foo(**data)`.

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,113 @@
...
+    def fetch_trials(self, query, selection=None):
+        """See :func:`~orion.storage.BaseStorageProtocol.fetch_trials`"""
+        return [Trial(**t) for t in self._db.read('trials', query=query, selection=selection)]
+
+    def register_trial(self, trial):
+        """See :func:`~orion.storage.BaseStorageProtocol.register_trial`"""
+        self._db.write('trials', trial.to_dict())
+        return trial
+
+    def retrieve_result(self, trial, results_file=None, **kwargs):
```
bouthilx: The behavior of `LegacyProtocol.retrieve_result` is different, it should be documented.

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,113 @@
...
+        """See :func:`~orion.storage.BaseStorageProtocol.retrieve_result`"""
+        results = self.converter.parse(results_file.name)
```
bouthilx: Do we need to hold a converter object in `self` or could we simply instantiate it locally each time?

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,113 @@
...
+            Trial.Result(
+                name=res['name'],
+                type=res['type'],
+                value=res['value']) for res in results
+        ]
+        return trial
+
+    def update_trial(self, trial: Trial, fields: Optional[dict] = None,
+                     where: Optional[dict] = None, **kwargs) -> Trial:
+        """See :func:`~orion.storage.BaseStorageProtocol.update_trial`"""
```
bouthilx: Same comment for fields kwargs.

tests/unittests/core/test_experiment.py
```diff
@@ -150,6 +155,23 @@ def new_config(random_dt):
     return new_config
 
 
+def assert_protocol(exp):
+    """Transitional method to move away from mongodb"""
+    # assert exp._protocol._db is create_db_instance
```
bouthilx: Why comment out?

tests/unittests/core/test_experiment.py
```diff
@@ -150,6 +155,23 @@ def new_config(random_dt):
...
+    pass
+
+
+def count_experiment(exp):
+    """Transitional method to move away from mongodb"""
+    # return count_experiment(exp)
```
bouthilx: You can remove comment

tests/unittests/core/test_experiment.py
```diff
@@ -150,6 +155,23 @@ def new_config(random_dt):
...
+    return exp._protocol._db.count("experiments")
+
+
+def get_underlying_db(exp):
```
bouthilx: get_db_from_view?

tests/unittests/core/test_experiment.py
```diff
@@ -996,15 +1018,27 @@ def test_experiment_view_stats(hacked_exp, exp_config, random_dt):
     assert len(stats) == 6
 
 
+#
+# @pytest.mark.usefixtures("with_user_tsirif")
```
bouthilx: You can remove this commented out test.

src/orion/storage/base.py
bouthilx: What is the error message?

src/orion/storage/legacy.py
Delaunay: I believe that's how it was done before

bouthilx: I think we can make it better. :)

src/orion/storage/legacy.py
Delaunay: what if the attribute name is odd ? `space deal_with_it`

src/orion/storage/base.py
Delaunay: experiment undefined

src/orion/storage/legacy.py
bouthilx: How can that be an attribute of an object? :thinking:

Delaunay: could have pushed with something different, what if MLFlow, comet ml or else have weird stuff ?

src/orion/storage/base.py
bouthilx: Oh
```
Examples
--------
>>> experiment = Experiment('some experiment')
>>> query = {'status': 'RUNNING'}
>>> BaseStorageProtocol.update_experiment(experiment, fields=query)
some output ...
>>> BaseStorageProtocol.update_experiment(experiment, status='RUNNING')
```

src/orion/storage/legacy.py
bouthilx: But the fields we pass through this are the attributes of the trial object. If some weird mapping is required for some other backend then they will handle this internally.

src/orion/core/worker/experiment.py
```diff
@@ -325,7 +330,8 @@ def register_lie(self, lying_trial):
         lying_trial.status = 'completed'
         lying_trial.end_time = datetime.datetime.utcnow()
 
-        self._db.write('lying_trials', lying_trial.to_dict())
+        # FIXME: find a way to implement lying_trials differently
```
bouthilx: Why? ;)

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,119 @@
...
+        return [Trial(**t) for t in self._db.read('trials', query=query, selection=selection)]
+
+    def register_trial(self, trial):
+        """See :func:`~orion.storage.BaseStorageProtocol.register_trial`"""
+        self._db.write('trials', trial.to_dict())
+        return trial
+
+    def retrieve_result(self, trial, results_file=None, **kwargs):
+        """Parse the result file that was generated by the trial process and
+        append the result to the trial object
```
bouthilx: Note that the result is not updated in DB, whereas the doc of BaseStorageProtocol states the opposite. I would lean on making this function only update the local trial, and `update_trial` is the task of `Experiment.push_completed_trial`.

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,119 @@
...
+                type=res['type'],
+                value=res['value']) for res in results
+        ]
+        return trial
+
+    def update_trial(self, trial: Trial, where=None, **kwargs) -> Trial:
+        """See :func:`~orion.storage.BaseStorageProtocol.update_trial`"""
+
+        if not kwargs:
+            return trial
```
bouthilx: return False?

src/orion/storage/legacy.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def create_experiment(self, config):
+        """See :func:`~orion.storage.BaseStorageProtocol.create_experiment`"""
+        return self._db.write('experiments', config)
+
+    def update_experiment(self, experiment, where=None, **kwargs):
+        """See :func:`~orion.storage.BaseStorageProtocol.update_experiment`"""
+
+        if not kwargs:
+            return True
```
bouthilx: return False?

src/orion/storage/base.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Parameters
+        ----------
+        experiment: Experiment
+            Experiment object to update
+
+        where: Optional[dict]
+            constraint experiment must respect
+
+        kwargs: dict
```
```suggestion
        **kwargs: dict
```

src/orion/storage/base.py
```diff
@@ -0,0 +1,121 @@
...
+            a dictionary of fields to update
+
+        """
```
bouthilx: Maybe add a `Returns` section. This will help make the behavior coherent for all backends.

src/orion/storage/legacy.py
Delaunay: nothing to update is it really failing ?

src/orion/storage/base.py
```diff
@@ -0,0 +1,121 @@
...
+        trial: Trial
+            Trial object to update
+
+        where: Optional[dict]
+            constraint trial must respect
+
+        kwargs: dict
+            a dictionary of fields to update
+
+        returns the updated trial
```
bouthilx: returns section

src/orion/storage/legacy.py
bouthilx: Oh, sometimes it is None, and if write succeeds but read doesn't for some obscure reason then the output is the trial document from the write. :thinking: What about just updating and leaving it up to the external code to update the trial object if necessary?

src/orion/storage/legacy.py
bouthilx: Well, if the question is *was the storage updated*, then the answer is no.

tests/unittests/core/test_experiment.py
```diff
@@ -734,6 +736,37 @@ def fetch_lost_trials(self, query):
...
+    hacked_exp.update_completed_trial(trial, results_file=results_file)
+
+    yo = database.trials.find_one({'_id': trial.id})
+
+    assert len(yo['results']) == len(trial.results)
+    assert yo['results'][0] == trial.results[0].to_dict()
+    assert yo['status'] == 'completed'
+    assert yo['end_time'] == random_dt
+
+    results_file.delete
```
bouthilx: delete()?

Delaunay: all good now

PR #223, opened 2019-07-17
Don't configure experiment views

bouthilx:
Why:

When configuring the experiment, missing local config files can make the
configuration crash. There is however little value for now to configure
experiment views as we are never using their space or algorithm.
Therefore the simplest solution to this problem is to stop configuring
the experiment views. This may be problematic however for users who may
use the experiment view to fetch and analyze results. For this reason,
we should seek to bring back the full instantiation of experiment views
once the refactoring of Oríon's global configuration is done.


bouthilx:
[Dont configure experiment views](https://trello.com/c/O6i0Hxxf/198-dont-configure-experiment-views)

src/orion/core/worker/experiment.py
```diff
@@ -781,18 +781,22 @@ def __init__(self, name, user=None):
...
-
-            raise
-
+        # TODO: Views are not fully configured until configuration is refactored
+        #       This snippet is to instantiate adapters anyhow, because it is required for
+        #       experiment views in EVC.
+        if self.refers and not isinstance(self.refers.get('adapter'), BaseAdapter):
+            self._experiment.refers['adapter'] = Adapter.build(self.refers['adapter'])
+
+        # try:
```
DatCorno: We should remove this commented code. It's on Github, no need to keep it in.

bouthilx: I think in this case it is better to keep it there as a strong visual remainder of the work required here once the refactoring of the configuration is done.

PR #224, opened 2019-07-17
Make sure higher shape points are not `ndarray` objects

DatCorno:


src/orion/core/utils/points.py
```diff
@@ -0,0 +1,73 @@
...
+
+.. module:: points
+   :platform: Unix
+   :synopsis: Conversion functions between higher shape points and lists.
+
+"""
+
+
+def regroup_dims(point, space):
+    """Take a list of points and regroup them appropriately as a point from `space`.
```
bouthilx: list of items representing a point

src/orion/core/utils/points.py
```diff
@@ -0,0 +1,73 @@
...
+    Parameters
+    ----------
+    point: array
+        Points to be regrouped.
+    space: `orion.algo.space.Space`
+        The optimization space.
+
+    Returns
+    -------
+    list or tuple
```
bouthilx: It should always be the same. I would go for list.

bouthilx:
[Transform `numpy.ndarray` to list inside Trial](https://trello.com/c/T3FoKPex/199-transform-numpyndarray-to-list-inside-trial)

PR #231, opened 2019-07-20
[WIP] Track integration

Delaunay:


src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,51 @@ def consume(self, trial):
...
+
+        Parameters
+        ----------
+        results_file: str
+           file used to store results, this is only used by the legacy protocol
+
+        trial: Trial
+           reference to the trial object that is going to be run
+
+        Note
```
bouthilx: Notes

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,51 @@ def consume(self, trial):
...
+        ----
+        This functions define the environment variables described below
+
+        ORION_PROJECT: str
+           name of the project the user is currently working on.
+           it defaults to `orion` is no project name was defined
```
bouthilx:
```suggestion
           it defaults to `orion` if no project name was defined
```

bouthilx: Maybe we should add a note that this is specific to Track StorageProtocol

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,51 @@ def consume(self, trial):
...
+
+        ORION_EXPERIMENT: str
+           current experiment id that this trial belong to
```
bouthilx:
```suggestion
           current experiment id that this trial belongs to
```

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,51 @@ def consume(self, trial):
...
+
+        ORION_TRIAL_ID: str
+           current trial id that is currently being ran in this process
```
bouthilx:
```suggestion
           current trial id that is currently being executed in this process
```

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,51 @@ def consume(self, trial):
...
+
+        ORION_RESULTS_PATH: str
+           orion results file that is read by the legacy protocol to get the results of the trial
+           after a successful run
+
+        """
+        env = dict(os.environ)
+
+        env['ORION_PROJECT'] = 'orion'
+        env['ORION_EXPERIMENT'] = self.experiment.name
```
bouthilx: This is not the experiment's id. The id is based on (name, metadata.user), and will now be (name, version, metadata.user).

src/orion/storage/base.py
```diff
@@ -104,6 +115,10 @@ def retrieve_result(self, trial, *args, **kwargs):
         """
         raise NotImplementedError()
 
+    def fetch_pending_trials(self, experiment):
+        """Fetch trials that have not run yet"""
```
bouthilx: This should be all trials that can be executed. It includes interrupted and suspended trials, which mean they were executed yet but didn't completed.

src/orion/storage/legacy.py
```diff
@@ -142,3 +151,11 @@ def update_trial(self, trial: Trial, where=None, **kwargs) -> Trial:
 
         where['_id'] = trial.id
         return self._db.write('trials', data=kwargs, query=where)
+
+    def fetch_pending_trials(self, experiment):
+        """Fetch trials that have not run yet"""
```
bouthilx: Same comment:
> This should be all trials that can be executed. It includes interrupted and suspended trials, which mean they were executed yet but didn't completed.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+log = logging.getLogger(__name__)
+
+
+if HAS_TRACK:
+    _status = [
+        CustomStatus('new', TrackStatus.CreatedGroup.value + 1),
+        CustomStatus('reserved', TrackStatus.CreatedGroup.value + 2),
+
+        CustomStatus('suspended', TrackStatus.FinishedGroup.value + 1),
+        # CustomStatus('completed', TrackStatus.FinishedGroup.value + 2),
```
bouthilx: Why not completed?

bouthilx: Oh, just saw below. There is no official status for any of these in Track beside Completed?

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+    """Convert track status to orion status"""
+    return status.name.lower()
+
+
+def remove_leading_slash(name):
+    """Remove leading slash"""
+    # if name[0] == '/':
+    #     return name[1:]
+    # return name
+    return name
```
bouthilx: Not doing it anymore?

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+def add_leading_slash(name):
+    """Add leading slash"""
+    # if name[0] == '/':
+    #     return name
+    # return '/' + name
+    return name
```
bouthilx: Not doing it anymore?

src/orion/storage/track.py
Delaunay: already exists in Track, we reused track's completed instead of defining another.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        self.memory = orion_trial
+        self.session_group = None
+        self._params = None
+        self.objective_key = objective
+        self.objectives_values = None
+        self._results = []
+
+    def _repr_values(self, values, sep=','):
+        """Represent with a string the given values."""
+        return
```
bouthilx: Why do you need this if it does nothing?

src/orion/storage/track.py
Delaunay: still planing too but orion complains if I strip them and put them back

src/orion/storage/track.py
Delaunay: don't recall in particular, but it is probably a function elsewhere that calls this methods when getting a Trial, needed for duck typing

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def __str__(self):
+        """Represent partially with a string."""
+        param_rep = ','.join(map(lambda value: "{0.name}:{0.value}".format(value), self.params))
+        ret = "TrialAdapter(uid={3}, experiment={0}, status={1}, params={2})".format(
+            repr(self.experiment[:10]), repr(self.status), param_rep, self.storage.uid)
+        return ret
```
bouthilx: Is this string supposed to be different than `str(Trial())`?

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        return convert_track_status(self.storage.status)
+
+    @status.setter
+    def status(self, value):
+        """See `~orion.core.worker.trial.Trial`"""
+        pass
+
+    def to_dict(self):
+        """See `~orion.core.worker.trial.Trial`"""
+        import copy
```
bouthilx: Why not import at top?

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        return None
+
+    @property
+    def objective(self):
+        """See `~orion.core.worker.trial.Trial`"""
+        def result(val):
+            return OrionTrial.Result(name=self.objective_key, value=val, type='objective')
+
+        if self.objective_key is None:
+            raise RuntimeError('not objective was defined!')
```
bouthilx:
```suggestion
            raise RuntimeError('no objective key was defined!')
```

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+                project_id=self.project.uid,
+                metadata=to_json(config)
+            )
+        )
+
+        config['_id'] = self.group.uid
+        return config
+
+    def update_experiment(self, experiment, where=None, **kwargs):
+        """Update a the fields of a given trials
```
bouthilx:
```suggestion
        """Update the fields of a given trials
```

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+
+        Parameters
+        ----------
+        experiment: Experiment
+            Experiment object to update
+
+        where: Optional[dict]
+            constraint experiment must respect
```
bouthilx:
```suggestion
            constraint experiment must respect for the update to take place
```

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+
+        **kwargs: dict
+            a dictionary of fields to update
+
+        Returns
+        -------
+        returns true if the underlying storage was updated
+
+        """
+        pass
```
bouthilx: experiments cannot be updated for now?

src/orion/storage/track.py
Delaunay: yes, more compact & easier to read

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        return self.backend.fetch_groups(query)
+
+    def register_trial(self, trial):
+        """Create a new trial to be executed"""
+        stamp = datetime.datetime.utcnow()
+        trial.status = 'new'
+        trial.submit_time = stamp
+
+        metadata = dict()
+        metadata['params_types'] = {remove_leading_slash(p.name): p.type for p in trial. params}
```
bouthilx:
```suggestion
        metadata['params_types'] = {remove_leading_slash(p.name): p.type for p in trial.params}
```

src/orion/storage/track.py
Delaunay: tracks create a new experiment every time, the experiments old are never updated

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        metadata['submit_time'] = to_json(trial.submit_time)
+        metadata['end_time'] = to_json(trial.end_time)
+        metadata['worker'] = trial.worker
+        metadata['metric_types'] = {remove_leading_slash(p.name): p.type for p in trial.results}
+        metadata['metric_types'][self.objective] = 'objective'
+        metadata['heartbeat'] = stamp
+
+        metrics = defaultdict(list)
+        for p in trial.results:
+            metrics[p.name] = p.value
```
bouthilx: A result value is a float or an integer, not a list. The defaultdict(list) won't have lists as values.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        """Fetch all the trials that match the query"""
+        def sort_key(item):
+            submit_time = item.submit_time
+            if submit_time is None:
+                return 0
+            return submit_time
+
+        query = to_json(query)
+
+        new_query = {}
```
bouthilx: The query conversion process should be extracted and tested separately.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+                    self.backend.set_trial_status(trial, get_track_status(value))
+                elif key in self._ignore_updates_for:
+                    continue
+                else:
+                    pair = {key: to_json(value)}
+                    self.backend.log_trial_metadata(trial, **pair)
+
+            return True
+
+        except RuntimeError:
```
bouthilx: Is this RuntimeError raises within Track? Would it make sense to define a custom error? I'm worried that we catch errors that we didn't intend to.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def update_trial(self, trial, where=None, **kwargs):
+        """Update the fields of a given trials
+
+        Parameters
+        ----------
+        trial: Trial
+            Trial object to update
+
+        where: Optional[dict]
```
bouthilx: The query constraint is not used.

src/orion/storage/track.py
```diff
@@ -0,0 +1,464 @@
...
+        refreshed_trial = self.backend.get_trial(trial)[0]
+        new_trial = TrialAdapter(refreshed_trial, objective=self.objective)
+
+        assert new_trial.objective is not None, 'Trial should have returned an objective value!'
+
+        log.info("trial objective is (%s: %s)", self.objective, new_trial.objective.value)
+        return new_trial
+
+    def fetch_pending_trials(self, experiment):
+        """Fetch trials that have not run yet"""
```
bouthilx: Same comment for the docstring

src/orion/storage/track.py
bouthilx: And there is no official status for any of these in Track beside Completed?

src/orion/storage/track.py
bouthilx: I would suggest we harmonize str(Trial) and str(TrialAdapter) for coherence. Do you think it is worth changing str(Trial)?

src/orion/storage/track.py
bouthilx: How can we support updates then? There is a few fields that can be updated in Oríon, such as `max_trials`.

bouthilx:
[Adapt track to latest Orion](https://trello.com/c/T5gROJ6I/200-adapt-track-to-latest-orion)

src/orion/storage/track.py
Delaunay: We should remove it entirely and move away for dict queries

bouthilx: So what should we do now?

src/orion/storage/track.py
Delaunay: actualyl yes I, updated the status, only 'reserved' and 'new' are custom status

src/orion/storage/track.py
Delaunay: So it actually does not need the constraint because all the information needed to perform the update is stored in `trial` and track cannot do batch update

src/orion/storage/track.py
bouthilx: Did you push the modifications? I don't see any.

src/orion/storage/track.py
Delaunay: Should merge & it will be refactored shortly

src/orion/storage/track.py
Delaunay: not yet

src/orion/storage/track.py
bouthilx: That means it cannot work with Orion. We need the query to ensure we do not reserve trials twice. Isn't there a global lock in Track that we could use for the time of the updates?

src/orion/storage/track.py
bouthilx: Sounds risky as we don't know exactly when and how we will refactor. Meanwhile any change in the core that breaks this will go undetected.

src/orion/storage/track.py
Delaunay: track adds a version id for each records, they should match else it means the record was modified by something else

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
             trial.status = 'broken'
             self.experiment.update_trial(trial, status=trial.status)
 
+    def get_execution_environment(self, trial, results_file='results.log'):
+        """Set a few environment variables to allow users and
+        underlying processes to know if they are running under orion
```
DatCorno: *Oríon

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
...
+
+        Parameters
+        ----------
+        results_file: str
+           file used to store results, this is only used by the legacy protocol
```
DatCorno: only use*

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
...
+
+        trial: Trial
+           reference to the trial object that is going to be run
+
+        Notes
+        ----
+        This functions define the environment variables described below
```
DatCorno: function* defines*

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
...
+
+        ORION_PROJECT: str
+           name of the experiment the user is currently working on.
+           A project is a set of experiments
+
+        ORION_EXPERIMENT: str
+           current experiment triplet that fully the experiment (i.e unique)
```
DatCorno: Can you clarify this? I don't really understand what this is supposed to say.

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
...
+
+        trial: Trial
+           reference to the trial object that is going to be run
+
+        Notes
+        ----
+        This functions define the environment variables described below
+
+        ORION_PROJECT: str
+           name of the experiment the user is currently working on.
```
DatCorno: Wouldn't the name of the Experiment the user is currently working on be stored inside `ORION_EXPERIMENT` below?

src/orion/core/worker/consumer.py
```diff
@@ -100,6 +100,52 @@ def consume(self, trial):
...
+        ORION_RESULTS_PATH: str
+           orion results file that is read by the legacy protocol to get the results of the trial
+           after a successful run
+
+        """
+        env = dict(os.environ)
+
+        env['ORION_PROJECT'] = self.experiment.name
+        env['ORION_EXPERIMENT'] = \
+            f'{self.experiment.name}-{self.experiment.version}-{self.experiment.metadata["user"]}'
```
DatCorno: We support Python 3.5, so we cannot use `f-string` for formatting since they were introduced in Python 3.6

bouthilx:
#289

PR #234, opened 2019-07-22
Handle broken refers in experiment configuration 

bouthilx:
Why:

If the process is stopped between the registration of the experiment and
the update of exp[refers][self._id], then the experiment will be somehow
broken, i.e., many code parts will crash because of it.

How:

Handle refers creation during configuration with default values if
nothing is set so that abortion during registration will in worst case
lead to inconsistant refers that is easier to fix. Previously broken
refers would be {}, which is irrecuperable for child experiments. Now
broken refers would be {parent_id: ok, adapter: ok, root_id: None},
which is simple to infer and fix.

### DIFF
--- a/src/orion/core/cli/info.py
+++ b/src/orion/core/cli/info.py
@@ -29,8 +29,8 @@ def add_subparser(parser):
 
 def main(args):
     """Fetch config and info experiments"""
-    experiment_view = EVCBuilder().build_view_from(args)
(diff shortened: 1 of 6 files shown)


bouthilx:
[Handle broken `refers` in commands `info` and `list`.](https://trello.com/c/s0uenxJR/218-handle-broken-refers-in-commands-info-and-list)

PR #235, opened 2019-07-22
Move TrialMonitor/TrialPacemaker inside consumer

bouthilx:
Why:

We need access to the pacemaker when the trials is completed so that we
can stop it. Otherwise, the thread continues until if makes another
heartbeat before it crashes and quit, which can take lot of time. The
execution of Orion hangs at the end of experiment until the pacemaker
stops.

How:

Move pacemaker inside Consumer so that it can be stopped right at end of
trial execution. Note that there is some delay between the reservation
and the start of pacemaker, which makes it possible to loose the
reservation before execution begins. That is not very likely however
since there is not a large time gap between reservation and execution.

### DIFF
--- a/src/orion/core/worker/consumer.py
+++ b/src/orion/core/worker/consumer.py
@@ -16,7 +16,7 @@ import tempfile
 
 from orion.core.io.space_builder import SpaceBuilder
 from orion.core.utils.working_dir import WorkingDir
-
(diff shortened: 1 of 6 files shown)


bouthilx:
[Find a way to stop the trial hearbeat thread when trial is completed.](https://trello.com/c/5jTLIaY6/216-find-a-way-to-stop-the-trial-hearbeat-thread-when-trial-is-completed)

PR #236, opened 2019-07-23
Bring back space.{un}pack_point but deprecate them

bouthilx:
Why:

The plugin orion.skopt.algo is relying on these functions so we need to
keep them unless we update the plugin.

How:

Call the new methods internally and log a warning that {un}pack_point
will be removed in v0.2.0.


bouthilx:
[Fix {un}pack deprecation](https://trello.com/c/JsXtGBlE/220-fix-unpack-deprecation)

PR #237, opened 2019-07-23
Fix Trial ID mismatch with higher dimension shapes

DatCorno:


bouthilx:
[Trial id mismatch when shape is not 1](https://trello.com/c/C6JUjBWn/153-trial-id-mismatch-when-shape-is-not-1)

PR #239, opened 2019-07-23
Add configuration object for global configuration

bouthilx:
Why:

The current implementation of the default > global > env var
configuration hierarchy is scattered is resolve_config and
experiment_builder. We should centralize this in an object which can
also be easily accessed from any submodule.

How:

Take blocks's Configuration class and make a few changes to support
subconfigurations.

### DIFF
--- a/src/orion/core/io/config.py
+++ b/src/orion/core/io/config.py
@@ -1,0 +1,246 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.core.io.config` -- Configuration object
+===================================================
+
+.. module:: config
+   :platform: Unix
+   :synopsis: Configuration object to define package configuration
+
+
+Highly inspired from https://github.com/mila-iqia/blocks/blob/master/blocks/config.py.
+
+"""
+import logging
+import os
+
+import yaml
+
+from orion.core.utils.flatten import flatten
+
+
+logger = logging.getLogger(__name__)
+
+
+NOT_SET = object()
+
+
+class ConfigurationError(Exception):
+    """Error raised when a configuration value is requested but not set."""
+
+
+class Configuration:
+    """Configuration object
+
+    Provides default values configurable at different levels. The configuration object can have
+    global default values, which may be overriden by user with yaml configuration files, environment
+    variables or by setting directly the values in the configuration object. In order, direct
+    definition overrides, environment variables, which overrides yaml configuration, which overrides
+    default values in configuration object definition.
+
+    Examples
+    --------
+    >>> config = Configuration()
+    >>> config.add_option('test', str, 'hello', 'TEST_ENV')
+    >>> config.test
+    'hello'
+    >>> config.load_yaml('some_config.yaml')
+    >>> config.test
+    'in yaml!'
+    >>> os.environ['TEST_ENV'] = 'there!'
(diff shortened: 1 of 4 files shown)


src/orion/core/io/config.py
```diff
@@ -0,0 +1,278 @@
...
+
+class ConfigurationError(Exception):
+    """Error raised when a configuration value is requested but not set."""
+
+
+class Configuration:
+    """Configuration object
+
+    Provides default values configurable at different levels. The configuration object can have
+    global default values, which may be overriden by user with yaml configuration files, environment
```
DatCorno: overwritten*

src/orion/core/io/config.py
```diff
@@ -0,0 +1,278 @@
...
+
+        Parameters
+        ----------
+        path: str
+            Path to the global configuration file.
+
+        Raises
+        -------
+        ConfigurationError
+            If some options in the yaml file does not exist in the config
```
DatCorno: option*

src/orion/core/io/config.py
```diff
@@ -0,0 +1,278 @@
...
+        if 'value' in config_setting:
+            value = config_setting['value']
+        elif ('env_var' in config_setting and
+              config_setting['env_var'] in os.environ):
+            value = os.environ[config_setting['env_var']]
+        elif 'yaml' in config_setting:
+            value = config_setting['yaml']
+        elif 'default' in config_setting:
+            value = config_setting['default']
+        else:
```
DatCorno: I think we could do some kind of more good-looking loop on this, no?

src/orion/core/io/config.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def __setattr__(self, key, value):
+        """Set option value or subconfiguration
+
+        Parameters
+        ----------
+        key: str
+            The key or namespace to set the value of the configuration.
+            If the configuration has subconfiguration, the key may be
+            hierarchical with each levels seperated by dots.
```
DatCorno: with each level separated

src/orion/core/io/config.py
```diff
@@ -0,0 +1,278 @@
...
+            YAML configuration file.
+
+        """
+        self.config[key] = {'type': option_type}
+        if env_var is not None:
+            self.config[key]['env_var'] = env_var
+        if default is not NOT_SET:
+            self.config[key]['default'] = default
+
+    # def to_dict(self):
```
DatCorno: Should we remove the rest of this commented code?

bouthilx: Yes!

src/orion/core/io/config.py
bouthilx: Not sure, this really is a `switch` case.

src/orion/core/io/config.py
bouthilx: In this case I believe overridden (there was a typo) is more accurate, as we are not flushing the default value but rather adding another value which takes precedence, and removing it would revert the value of the config to the default value. (the config contains default value, yaml value, env var value and local value and provide the one with highest precedence).

DatCorno: Makes sense.

PR #242, opened 2019-07-24
Add auto-incrementing of version  for Experiment

DatCorno:

PR #243, opened 2019-07-31
Add `--version` argument

DatCorno:
This PR depends on #242


DatCorno:
[Add --version option](https://trello.com/c/3OLZEiXk/177-add-version-option)

src/orion/core/worker/experiment.py
```diff
@@ -130,10 +130,22 @@ def __init__(self, name, user=None):
                       name, user)
 
             if len(config) > 1:
-                version = max(map(lambda exp: exp['version'], config))
+                max_version = max(map(lambda exp: exp['version'], config))
+
+                if version is None:
+                    self.version = max_version
+                else:
+                    self.version = version
```
bouthilx: This should be set at line 124, so this here would only be a if instead of if-else

src/orion/core/worker/experiment.py
```diff
@@ -130,10 +130,22 @@ def __init__(self, name, user=None):
                       name, user)
 
             if len(config) > 1:
-                version = max(map(lambda exp: exp['version'], config))
+                max_version = max(map(lambda exp: exp['version'], config))
```
```suggestion
                max_version = max(config, key=lambda exp: exp['version'])
```

src/orion/core/worker/experiment.py
```diff
@@ -580,7 +592,9 @@ def configure(self, config, enable_branching=True, enable_update=True):
             # will be used when EVC is not available.
             # must_branch = self._is_different_from(experiment.configuration)
             branching_configuration = fetch_branching_configuration(config)
-            conflicts = detect_conflicts(self.configuration, experiment.configuration)
+            configuration = self.configuration
+            configuration['_id'] = self._id
```
bouthilx: Where do we need this _id?

bouthilx: ok, found: https://github.com/Epistimio/orion/blob/develop/src/orion/core/evc/conflicts.py#L1611

tests/unittests/core/io/interactive_commands/test_branching_prompt.py
```diff
@@ -63,14 +63,6 @@ def conflicts(new_dimension_conflict, new_cat_dimension_conflict,
     return conflicts
 
 
-@pytest.fixture
-def conflicts_exp_name(conflicts, failing_exp_name_conflict, experiment_name_conflict):
-    """Return a list of conflicts with an ExperimentNameConflict that fails."""
```
bouthilx: Why it isn't needed anymore? Because all exp-name now fails?

tests/unittests/core/worker/test_experiment.py
```diff
@@ -231,6 +253,43 @@ def test_existing_experiment(self, create_db_instance, exp_config):
...
+        """Create an already existing experiment with a version."""
+        create_db_instance.write('experiments', parent_version_config)
+        create_db_instance.write('experiments', child_version_config)
+
+        exp = Experiment("old_experiment", user="corneauf", version=1)
+        assert exp.version == 1
+
+    def test_old_experiment_w_version_bigger_than_max(self, create_db_instance,
+                                                      parent_version_config, child_version_config):
+        """Create an already existing experiment without a version."""
```
bouthilx: with a too large version?

src/orion/core/worker/experiment.py
DatCorno: No, because if `version` is left to `None` and there is no configuration inside the database then `version` will never be set to `1`.

bouthilx: oh ok!

tests/unittests/core/io/interactive_commands/test_branching_prompt.py
DatCorno: Because it was a workaround for the fact that the `ExperimentNameConflict` inside the original `conflicts` did not fail anymore. So I just replaced the current passing `ExperimentNameConflict` with the one that fails which now invalidates the existence of that fixture.

tests/unittests/core/worker/test_experiment.py
DatCorno: Oopsie

src/orion/core/worker/experiment.py
DatCorno: :upside_down_face:

PR #244, opened 2019-07-31
Deprecate `--auto-resolution` and introduce `--manual-resolution`

DatCorno:
This PR depends on #243

PR #245, opened 2019-08-01
Fix problem with incrementing version on experiment with no same-named children

DatCorno:
If an experiment name `root` already has at least one child not name `root`, maybe `child_1`, then it is impossible to create a new experiment name `root` to increment its version number.

While we're at it, also fixed a bug where branching from a selected version through `--version` would fail and branch from latest version.
This PR depends on #244

### DIFF
--- a/src/orion/core/evc/conflicts.py
+++ b/src/orion/core/evc/conflicts.py
@@ -1608,7 +1608,7 @@ class ExperimentNameConflict(Conflict):
             # If we made it this far, new_name is actually the name of the parent.
             parent = self.conflict.old_config
 
-            query = {'refers.parent_id': parent['_id']}
+            query = {'name': parent['name'], 'refers.parent_id': parent['_id']}
             children = len(get_storage().fetch_experiments(query))
 
             return bool(children)
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -125,7 +125,6 @@ class Experiment:
         self._storage = get_storage()
 
         config = self._storage.fetch_experiments({'name': name, 'metadata.user': user})
-        print(config)
 
         if config:
             log.debug("Found existing experiment, %s, under user, %s, registered in database.",
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -125,6 +125,7 @@ class Experiment:
         self._storage = get_storage()
 
         config = self._storage.fetch_experiments({'name': name, 'metadata.user': user})
+        print(config)
 
         if config:
             log.debug("Found existing experiment, %s, under user, %s, registered in database.",
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -144,8 +144,8 @@ class Experiment:
 
                 self.version = min(self.version, max_version)
 
-                log.info("Many versions for experiment %s have been found. Using latest\
-                          version %s.", name, self.version)
+                log.info("Many versions for experiment %s have been found. Using latest "
+                         "version %s.", name, self.version)
                 config = filter(lambda exp: exp['version'] == self.version, config)
 
             config = sorted(config, key=lambda x: x['metadata']['datetime'],
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -575,7 +575,7 @@ class Experiment:
             raise DuplicateKeyError("Cannot register an existing experiment with a new config")
 
         # Copy and simulate instantiating given configuration
-        experiment = Experiment(self.name)
+        experiment = Experiment(self.name, version=self.version)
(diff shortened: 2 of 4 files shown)


src/orion/core/evc/conflicts.py
```diff
@@ -1608,7 +1608,7 @@ def _check_for_children(self):
             # If we made it this far, new_name is actually the name of the parent.
             parent = self.conflict.old_config
 
-            query = {'refers.parent_id': parent['_id']}
+            query = {'name': parent['name'], 'refers.parent_id': parent['_id']}
```
bouthilx: This would only fetch children with same name, thus not all children. I understand this is what we are looking for here, but then the methods name become very confusing. `_check_for_greater_versions` maybe?

src/orion/core/worker/experiment.py
```diff
@@ -573,7 +573,7 @@ def configure(self, config, enable_branching=True, enable_update=True):
             raise DuplicateKeyError("Cannot register an existing experiment with a new config")
 
         # Copy and simulate instantiating given configuration
-        experiment = Experiment(self.name)
+        experiment = Experiment(self.name, version=self.version)
```
bouthilx: How was this not detected by unit-tests and for which tests was it now required?

DatCorno: Because there are no tests for ExperimentNode.

src/orion/core/evc/conflicts.py
DatCorno: Sounds good.

src/orion/core/worker/experiment.py
bouthilx: We should add a unittest in `TestConfigProperty` to verify that configuring an experiment with version=1 when version 2 exists does not increment to 3.

tests/unittests/core/worker/test_experiment.py
```diff
@@ -641,6 +641,20 @@ def test_status_is_pending_when_increase_max_trials(self, exp_config):
...
+
+        query['version'] = 2
+        query.pop('_id')
+
+        get_storage().create_experiment(query)
+
+        exp = Experiment('experiment_test', version=1)
+
+        assert exp.version == 1
+
```
bouthilx: The problem was when we do experiment.configure, where we would create a new Experiment for which we didn't pass self.version.

tests/unittests/core/worker/test_experiment.py
```diff
@@ -643,17 +643,32 @@ def test_status_is_pending_when_increase_max_trials(self, exp_config):
...
 
-        get_storage().create_experiment(query)
+        config['version'] = 2
+        config['metadata']['user_args'].append("--y~+normal(0,1)")
+        config.pop('_id')
 
+        get_storage().create_experiment(config)
+        config.pop('_id')
+
+        config['branch'] = ['experiment_2']
```
bouthilx: Why do you change the name? This clearly should override with version 1.

DatCorno: Well, the behavior occurs when we use `--branch`, so I thought it be more explicit this way. This also mimics the functional test inside `test_branching` testing the same thing.

bouthilx: The problem is version is hard-set to 1 if we change the name. Actually, if we try to run version one but change it without specifying the name, it should raise an error if version 2 already exists right?

DatCorno: Yes, but here we're trying to branch from version 1 so we must change the name and explicitly ask for version 1.

bouthilx: What happens if we don't?

DatCorno: Then, there'll be no `version` key inside the `config` dictionary, so `None` will be passed and the latest experiment will be loaded instead of the first one, basically cancelling the effort we've made to not have that in the first place!

bouthilx: I meant if we don't change the name of the experiment, but specify version 1 (even thought there is already a version 2).

DatCorno: Then that's a whole other situation which is not what this test is about.

bouthilx: The test we need should test for the situation where the modification at line https://github.com/Epistimio/orion/pull/245/files#diff-6d901bae3fe0c9e74e137277f6fdf9b4R578 was required. Since changing the name forces an overwrite of the version number in the config [here](https://github.com/Epistimio/orion/blob/develop/src/orion/core/evc/conflicts.py#L1586) I hardly see of this test would fail if the previous modification was not there. From what I see with a second look at Experiment.configure, the problem will happen when we instantiate with version=1 while a child exists but we don't branch. In this case the version number in the copied experiment will be 2 while it should have been 1.

PR #246, opened 2019-08-01
Add version to `list` output

DatCorno:
This PR changes the output of `orion list` by adding the version of each experiment. This change has impacted `ExperimentNode` only but I decided to change `Experiment.__repr__()` anyways just because.

### DIFF
--- a/src/orion/core/evc/experiment.py
+++ b/src/orion/core/evc/experiment.py
@@ -122,7 +122,7 @@ class ExperimentNode(TreeNode):
     def tree_name(self):
         """Return a formatted name of the Node for a tree pretty-print."""
         if self.item is not None:
-            return self.name + " - v.{}".format(self.item.version)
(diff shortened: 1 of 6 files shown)


DatCorno:
[Add --version list](https://trello.com/c/pVa8NOtU/242-add-version-list)

src/orion/core/evc/experiment.py
```diff
@@ -117,6 +118,14 @@ def adapter(self):
         """Get the adapter of the experiment with respect to its parent"""
         return self.item.refers["adapter"]
 
+    @property
+    def tree_name(self):
+        """Return a formatted name of the Node for a tree pretty-print."""
+        if self.item is not None:
+            return self.name + " - v.{}".format(self.item.version)
```
bouthilx: `"-v{}"` for compactness? It renders a bit loose with `pptree` with spaces. There isn't x.y versions, so the dot seems superfluous.

PR #247, opened 2019-08-01
Add support for `--version` inside `info` command

DatCorno:
And fix a bug where `ExperimentNode` would enter an infinite loop. You're welcome.

### DIFF
--- a/src/orion/core/evc/experiment.py
+++ b/src/orion/core/evc/experiment.py
@@ -46,9 +46,9 @@ class ExperimentNode(TreeNode):
 
     """
 
-    __slots__ = ('name', '_no_parent_lookup', '_no_children_lookup') + TreeNode.__slots__
+    __slots__ = ('name', 'version', '_no_parent_lookup', '_no_children_lookup') + TreeNode.__slots__
 
-    def __init__(self, name, experiment=None, parent=None, children=tuple()):
+    def __init__(self, name, version, experiment=None, parent=None, children=tuple()):
         """Initialize experiment node with item, experiment, parent and children
 
         .. seealso::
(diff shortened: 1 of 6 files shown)


DatCorno:
[Add --version for info](https://trello.com/c/8zRGfJnN/241-add-version-for-info)

src/orion/core/evc/experiment.py
```diff
@@ -86,11 +87,12 @@ def parent(self):
...
             query = {'_id': self.item.refers['parent_id']}
-            selection = {'name': 1}
+            selection = {'name': 1, 'version': 1}
             experiments = Database().read('experiments', query, selection=selection)
             if experiments:
-                self.set_parent(ExperimentNode(name=experiments[0]['name']))
-
+                parent = experiments[0]
+                exp_node = ExperimentNode(name=parent['name'], version=parent['version'])
+                self.set_parent(exp_node)
```
bouthilx: Any idea why this problem never showed up before? We should add a test for this in test_experiment_tree.

src/orion/core/evc/experiment.py
```diff
@@ -106,10 +108,10 @@ def children(self):
         if not self._children and self._no_children_lookup:
             self._no_children_lookup = False
             query = {'refers.parent_id': self.item.id}
-            selection = {'name': 1}
+            selection = {'name': 1, 'version': 1}
             experiments = Database().read('experiments', query, selection=selection)
             for child in experiments:
-                self.add_children(ExperimentNode(name=child['name']))
+                self.add_children(ExperimentNode(name=child['name'], version=child['version']))
```
bouthilx: Same for children

DatCorno: Maybe because we never tried with an experiment tree with multiple versions of the same experiment? I don't know.

PR #250, opened 2019-08-08
no generic queries

Delaunay:


src/orion/core/worker/experiment.py
```diff
@@ -155,6 +153,23 @@ def __init__(self, name, user=None, version=None):
...
+
+        .. seealso::
+
+           :meth:`orion.core.io.database.AbstractDB.read` for more information about the
+           arguments.
+
+        """
+        return self._storage.fetch_experiment_trials(uid=self._id)
+
     def fetch_trials(self, query, selection=None):
```
bouthilx: Can we drop query and avoid adding `fetch_experiment_trials`?

src/orion/core/worker/experiment.py
```diff
@@ -215,71 +230,34 @@ def retrieve_result(self, trial, *args, **kwargs):
...
-            log.warning("While reserving trial: `score_handle` was provided, but "
-                        "parameter space has not been defined yet.")
-
-        selected_trial = random.sample(new_trials, 1)[0]
-        log.debug('%s selected (trial: %s)', '<' * _depth, selected_trial)
-
-        update = dict(status='reserved', heartbeat=datetime.datetime.utcnow())
+        Parameters
+        ----------
+        score_handle: Optional[Callable[[Trial], int]]
```
bouthilx: Deprecation should be mentioned in the docstring. And type should be `callable object, optional` ([ref]( https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard))

src/orion/storage/legacy.py
```diff
@@ -93,7 +100,35 @@ def fetch_experiments(self, query, selection=None):
...
-        return [Trial(**t) for t in self._db.read('trials', query=query, selection=selection)]
+        # import inspect
+        # stack = inspect.stack()
+        # warn(f'DEPRECATED {query} {selection}\n')
+        #
+        # print('---')
+        # for i in stack[1:]:
+        #     print('   ', i.filename)
+
+        return self._fetch_trials(query, selection)
```
bouthilx: We should deprecate now if we removed all generic queries.

src/orion/storage/legacy.py
```diff
@@ -135,10 +170,132 @@ def retrieve_result(self, trial, results_file=None, **kwargs):
 
         return trial
 
-    def update_trial(self, trial: Trial, where=None, **kwargs) -> Trial:
+    def get_trial(self, trial=None, uid=None):
+        """See :func:`~orion.storage.BaseStorageProtocol.get_trial`"""
```
bouthilx: Params are not documented in base.get_trial

src/orion/storage/legacy.py
```diff
@@ -135,10 +170,132 @@ def retrieve_result(self, trial, results_file=None, **kwargs):
...
+        }
+
+        return self._fetch_trials(query)
+
+    def push_trial_results(self, trial):
+        """See :func:`~orion.storage.BaseStorageProtocol.push_trial_results`"""
+        return self._update_trial(trial, **trial.to_dict(), where={'_id': trial.id})
+
+    def set_trial_status(self, trial, status, heartbeat=None):
+        """See :func:`~orion.storage.BaseStorageProtocol.set_trial_status`"""
```
bouthilx: This method needs more documentation. What happens if trial.status is not the same as in the DB?

src/orion/storage/legacy.py
```diff
@@ -135,10 +170,132 @@ def retrieve_result(self, trial, results_file=None, **kwargs):
...
+        """See :func:`~orion.storage.BaseStorageProtocol.count_broken_trials`"""
+        query = dict(
+            experiment=experiment._id,
+            status='broken'
+        )
+        return self._db.count('trials', query)
+
+    def update_heartbeat(self, trial):
+        """Update trial's heartbeat"""
+        return self._update_trial(trial, heartbeat=datetime.datetime.utcnow())
```
bouthilx: This should verify first that `trial.status == 'reserved'`.

tests/unittests/core/worker/test_experiment.py
```diff
@@ -7,125 +7,59 @@
...
 import pytest
 
 from orion.algo.base import BaseAlgorithm
 from orion.core.io.database import DuplicateKeyError
+from orion.core.utils.tests import OrionState
 import orion.core.worker.experiment
 from orion.core.worker.experiment import Experiment, ExperimentView
 from orion.core.worker.trial import Trial
 from orion.storage.base import get_storage
+from orion.storage.legacy import FailedUpdate
```
bouthilx: Could we move this FailedUpdate error to storage.base instead?

tests/unittests/core/worker/test_experiment.py
```diff
@@ -7,125 +7,59 @@
...
-        # Part of `TestReserveTrial.test_reserve_success`
-        assert a_list[0].status == 'new'
-        assert a_list[1].status == 'new'
-        assert a_list[2].status == 'interrupted'
-        assert a_list[3].status == 'suspended'
-        assert should_be_one == 1
-        return [a_list[0]]
-
-    monkeypatch.setattr(random, 'sample', mock_sample)
+def test_change_status_success(exp_config_file):
```
bouthilx: This should be in `tests/storage/test_legacy.py`

tests/unittests/core/worker/test_experiment.py
```diff
@@ -7,125 +7,59 @@
...
-        return [a_list[0]]
-
-    monkeypatch.setattr(random, 'sample', mock_sample)
+def test_change_status_failed_update(exp_config_file):
+    """Successfully find new trials in db and reserve one at 'random'."""
+    with OrionState(from_yaml=exp_config_file) as cfg:
+        trial = cfg.get_trial(0)
+        assert trial is not None, 'Was not able to retrieve trial for test'
 
+        try:
```
bouthilx: Better practice is to use with `pytests.raises(FailedUpdate) as exc:` and then verify that `str(exc.value)` has the proper message.

tests/unittests/core/worker/test_experiment.py
```diff
@@ -673,71 +607,35 @@ def test_reserve_none(self):
...
+        with OrionState(from_yaml=exp_config_file) as cfg:
+            exp = cfg.get_experiment('supernaedo2', user='dendi')
+            trial = exp.reserve_trial()
+
+            cfg.trials[0]['status'] = 'reserved'
+            cfg.trials[0]['start_time'] = random_dt
+            cfg.trials[0]['heartbeat'] = random_dt
+
+            for t in cfg.trials:
+                print(t['submit_time'], t['_id'])
```
bouthilx: remove print?

tests/unittests/core/worker/test_experiment.py
```diff
@@ -673,71 +607,35 @@ def test_reserve_none(self):
...
+
+            assert trial.to_dict() == cfg.trials[0]
+
     def test_reserve_dead_race_condition(self, exp_config, hacked_exp):
```
bouthilx:
```suggestion
    def test_reserve_when_exausted(self, exp_config, hacked_exp):
```

tests/unittests/core/worker/test_experiment.py
```diff
@@ -1107,10 +995,11 @@ def test_experiment_view_protocol_read_only():
     """Verify that wrapper experiments' database is read-only"""
```
bouthilx:
```suggestion
    """Verify that wrapper experiments' protocol is read-only"""
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -127,11 +132,48 @@ def _get_database(self):
...
+
+            key, value = self._find_unpickable_field(doc)
+            log.error('because (value %s) in (field: %s) is not pickable',
+                      value, key)
+            raise
 
         os.rename(tmp_file, self.host)
 
+    @staticmethod
+    def _find_unpickable_doc(database):
```
bouthilx: Seems like these could be standalone functions.

src/orion/core/io/database/pickleddb.py
```diff
@@ -127,11 +132,48 @@ def _get_database(self):
...
+        for name, collection in database._db.items():
+            documents = collection._documents
```
bouthilx:
```suggestion
            documents = collection.find()
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -127,11 +132,48 @@ def _get_database(self):
...
+                    pickle.dumps(doc)
+
+                except PicklingError:
+                    return name, doc
+
+        return None, None
+
+    @staticmethod
+    def _find_unpickable_field(doc):
+        for k, v in doc._data.items():
```
bouthilx:
```suggestion
        for k, v in doc.to_dict().items():
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -8,20 +8,25 @@
...
 import orion.core
 from orion.core.io.database import AbstractDB
 from orion.core.io.database.ephemeraldb import EphemeralDB
 
+log = logging.getLogger(__name__)
 
 DEFAULT_HOST = os.path.join(orion.core.DIRS.user_data_dir, 'orion', 'orion_db.pkl')
 
 
+# pylint: disable=protected-access
```
bouthilx: This could be moved to the only access to `database._db` in `_find_unpickable_doc`

src/orion/core/io/database/pickleddb.py
```diff
@@ -127,11 +132,48 @@ def _get_database(self):
     def _dump_database(self, database):
         """Write pickled DB on disk"""
         tmp_file = self.host + '.tmp'
-        with open(tmp_file, 'wb') as f:
-            pickle.dump(database, f)
+
+        try:
```
bouthilx: These new additions to PickledDB should be tested.

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.core.utils.state` -- Encapsulate Orion state
+=================================================================
```
bouthilx:
```suggestion
========================================================
```

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Examples
+    --------
+    >>> myconfig = {...}
+    >>> with OrionTestState(myconfig):
+        ...
+
+    """
+
+    SINGLETONS = (Storage, Legacy, Database, MongoDB, PickledDB, EphemeralDB)
```
bouthilx: Maybe add a TODO comment: Fix these singletons to remove Legacy, MongoDB, PickledDB and EphemeralDB.

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,180 @@
...
+    def restore_singleton(self, obj):
+        """Restore a singleton to its previous value"""
+        obj.instance = self.singletons.get(obj)
+
+    def storage(self, storage_type='legacy'):
+        """Return test storage"""
+        try:
+            config = {
+                'database': {
+                    'type': 'PickledDB',
```
bouthilx: All tests running with OrionState are on PickledDB but the rest not using OrionState is still on MongoDB right?

bouthilx:
As we broke the generic queries into many methods the needs for better documentation and unittests has increased. The is tests for set_trial_states but there should be for others too. I suggest you write generic unittests for all methods so that we can reuse the same tests for all storage protocol backends.

src/orion/core/utils/tests.py
Delaunay: yes

src/orion/storage/legacy.py
Delaunay: track change the status to running. the heartbeat monitor check if the status is in a broken state instead

bouthilx: The problem is `update_trial checks` that trial.status is still the same in the DB, but we could make a mistake and do `update_hearbeat` on a trial that is not reserved/running and that should not be possible. This is the legacy protocol, so we should never get running status from Track, only reserved.

src/orion/core/io/database/pickleddb.py
```diff
@@ -124,11 +155,24 @@ def _get_database(self):
...
+            with open(tmp_file, 'wb') as f:
+                pickle.dump(database, f)
+
+        except PicklingError:
+            collection, doc = find_unpickable_doc(database._db)
```
bouthilx:
```suggestion
            collection, doc = find_unpickable_doc(database._db)  # pylint: disable: protected-access
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -124,11 +155,24 @@ def _get_database(self):
 
         return database
 
+    # pylint: disable=protected-access
```
bouthilx: Could move this down.

src/orion/storage/base.py
```diff
@@ -84,6 +90,14 @@ def fetch_experiment_trials(self, experiment=None, uid=None):
     def get_trial(self, trial=None, uid=None):
         """Fetch a single trial
```
bouthilx: What if both are None or both are passed? This should be documented.

src/orion/storage/base.py
```diff
@@ -84,6 +90,14 @@ def fetch_experiment_trials(self, experiment=None, uid=None):
...
 
+        Parameters
+        ----------
+        trial: Trial
```
bouthilx: optional

src/orion/storage/base.py
```diff
@@ -84,6 +90,14 @@ def fetch_experiment_trials(self, experiment=None, uid=None):
...
+           trial object to retrieve from the database
+
+        uid: str
```
bouthilx: optional

src/orion/storage/legacy.py
Delaunay: shouldn't we just remove the query status=current_status.
The only time where there is a concurrent status update is when the monitor is updating the status. I do not think that it is that important now

bouthilx: It's true that the most important point for this which was in `experiment.reserve_trial` is not needed anymore. It seams to be a reasonable safeguard to keep nevertheless. There should not be 2 process messing with the status at the same time, so if this ever happens I think it should not go unnoticed.

bouthilx:
[No generic Queries](https://trello.com/c/rEJTdXrE/258-no-generic-queries)

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+    resources: list
+        List of resources to insert into the database
+
+    database: dict
+        Configuration of the underlying database
+
+    Examples
+    --------
+    >>> myconfig = {...}
+    >>> with OrionTestState(myconfig):
```
DatCorno: Shouldn't this be `with OrionState`?

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+    workers = []
+    tempfile = None
+
+    def __init__(self, experiments=None, trials=None, workers=None, resources=None,
+                 from_yaml=None, database=None):
+        if from_yaml is not None:
+            with open(from_yaml) as f:
+                exp_config = list(yaml.safe_load_all(f))
+                experiments = exp_config[0]
+                trials = exp_config[1]
```
DatCorno: I suggest we get the list of experiments and trials through a key inside the yaml, for example:
```
experiments:
                      [
                      ...
                      ]
trials
                      [
                      ...
                      ]
```
This way we could have a yaml file of only trials without having experiments.

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+        # sort by submit time like the protocol does
+        self.trials.sort(key=lambda obj: int(obj['_id'], 16), reverse=True)
+
+        for i, _ in enumerate(self.experiments):
+            path = os.path.join(
+                os.path.dirname(__file__),
+                self.experiments[i]["metadata"]["user_script"])
+
+            self.experiments[i]["metadata"]["user_script"] = path
+            self.experiments[i]['version'] = 1
```
DatCorno: Why set their version to one? If an experiment is a child of another, this will break.

Delaunay: no particular reason, this code essentially adapts:
https://github.com/Epistimio/orion/blob/d23e539d3556c3dea4f6157cf75a528c5ba39125/tests/conftest.py#L150

So I should remove the version or set it to 0 ?

src/orion/core/utils/tests.py
Delaunay: I agree, the reason why it is setup like this is because that is how the current YAMLs are structed. We might have to support both in the meantime. maybe something like this ?

```
    def from_yaml(self, file_name):
        """Load a YAML files containing the initial state of the database"""
        data = {
            'experiments': [],
            'trials': []
        }

        if file_name is None:
            return data

        with open(file_name) as f:
            try:
                # try to load the YAML as a dictionary
                return data.update(dict(yaml.safe_load_all(f)))

            except ValueError:
                # TODO: Remove legacy loading and use yaml dictionary instead
                # if it fails try the legacy loading scheme
                exp_config = list(yaml.safe_load_all(f))
                data['experiments'] = exp_config[0]
                data['trials'] = exp_config[1]
                return data
```

src/orion/storage/legacy.py
```diff
@@ -8,12 +8,13 @@
...
 
 """
+import datetime
 import logging
 
 from orion.core.io.convert import JSONConverter
 from orion.core.io.database import Database
 from orion.core.worker.trial import Trial
-from orion.storage.base import BaseStorageProtocol
+from orion.storage.base import BaseStorageProtocol, FailedUpdate, UnderdefinedCall
```
bouthilx:
```suggestion
from orion.storage.base import BaseStorageProtocol, FailedUpdate, UndefinedCall
```

src/orion/storage/legacy.py
```diff
@@ -91,9 +92,28 @@ def fetch_experiments(self, query, selection=None):
         """See :func:`~orion.storage.BaseStorageProtocol.fetch_experiments`"""
         return self._db.read('experiments', query, selection)
 
-    def fetch_trials(self, query, selection=None):
+    def fetch_trials(self, experiment=None, uid=None):
         """See :func:`~orion.storage.BaseStorageProtocol.fetch_trials`"""
-        return [Trial(**t) for t in self._db.read('trials', query=query, selection=selection)]
+        if uid is None:
+            if experiment is None:
+                raise UnderdefinedCall('Either `experiment` or `uid` should be set')
```
bouthilx:
```suggestion
                raise UndefinedCall('Either `experiment` or `uid` should be set')
```

src/orion/storage/legacy.py
bouthilx: How was this one resolved?

src/orion/storage/base.py
```diff
@@ -69,41 +81,110 @@ def register_lie(self, trial):
...
 
         Returns
         -------
-        returns true if the underlying storage was updated
+        return none if the trial is not found,
+
+        Raises
+        ------
+        UnderdefinedCall
+            if both trial and uid are not set
```
bouthilx: What if both are set? :)

src/orion/storage/base.py
```diff
@@ -69,41 +81,110 @@ def register_lie(self, trial):
...
+            trial id used to retrieve the trial object
 
         Returns
         -------
-        returns true if the underlying storage was updated
+        return none if the trial is not found,
+
+        Raises
+        ------
+        UnderdefinedCall
```
```suggestion
        UndefinedCall
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -127,8 +161,20 @@ def _get_database(self):
...
-            pickle.dump(database, f)
+
+        try:
+            with open(tmp_file, 'wb') as f:
+                pickle.dump(database, f)
+
+        except (PicklingError, AttributeError):
+            collection, doc = find_unpickable_doc(database._db)  # pylint: disable=protected-access
+            log.error('Document in (collection: %s) is not pickable\ndoc: %s',
+                      collection, doc._data)                     # pylint: disable=protected-access
```
bouthilx:
```suggestion
                      collection, doc.to_dict())
```

src/orion/storage/base.py
Delaunay: it works

tests/unittests/core/test_trial.py
```diff
@@ -120,12 +120,12 @@ def test_value_equal(self, exp_config):
...
         assert trials[0].params[1] != Trial.Param(**exp_config[1][0]['params'][0])
 
-    def test_str_trial(self, exp_config):
+    def test_pfmt_trial(self, exp_config):
         """Test representation of `Trial`."""
         t = Trial(**exp_config[1][1])
-        assert str(t) == "Trial(experiment='supernaedo2', status='completed',\n"\
-                         "      params=/decoding_layer:lstm_with_attention\n"\
-                         "             /encoding_layer:gru)"
+        assert t.pfmt() == "Trial(experiment='supernaedo2', status='completed',\n"\
```
bouthilx: Could we adapt the test instead and drop pfmt?

Delaunay: is there really a need for testing printing ?
How hard of a spec is this ?
Like do we expect people relying on it to parse Trials ?

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+
+@pytest.mark.parametrize('storage', storage_backends)
+class TestStorage:
+    """Test all storage backend"""
+
+    def test_create_experiment(self, storage):
+        """Test create experiment"""
+        with OrionState(experiments=[], database=storage) as cfg:
+            storage = cfg.storage()
+
```
bouthilx: Should test `storage.fetch_experiments({}) == 0` before to make sure the experiment was not there already.

Delaunay: experiments=[] guarantee this

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+    """Generate a set of experiments"""
+    users = ['a', 'b', 'c']
+    return [_generate(base_trial, 'metadata', 'user', value=u) for u in users]
+
+
+@pytest.mark.parametrize('storage', storage_backends)
+class TestStorage:
+    """Test all storage backend"""
+
+    def test_create_experiment(self, storage):
```
bouthilx: Should add a test where experiment already exist and cannot be written.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+
+            storage.create_experiment(base_experiment)
+
+            experiments = storage.fetch_experiments({})
+            assert len(experiments) == 1, 'Only one experiment in the database'
+
+            experiment = experiments[0]
+            assert base_experiment == experiment, 'Local experiment and DB should match'
+
+    def test_fetch_experiments(self, storage, name='default_name', user='default_user'):
```
bouthilx: Should add a test where query fails. There is only one experiment so this test would also pass with `storage.fetch_experiments({})`.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+            storage = cfg.storage()
+
+            experiments = storage.fetch_experiments({'name': name, 'metadata.user': user})
+            assert len(experiments) == len(cfg.experiments)
+
+            experiment = experiments[0]
+            assert experiment['name'] == name, 'name should match query'
+            assert experiment['metadata']['user'] == user, 'user name should match query'
+
+    def test_register_trial(self, storage):
```
bouthilx: Should add a test where trial cannot be registered because of DuplicateKeyError.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test register trial"""
+        with OrionState(experiments=[base_experiment], database=storage) as cfg:
+            storage = cfg.storage()
+            trial1 = storage.register_trial(Trial(**base_trial))
+            trial2 = storage.get_trial(trial1)
+
+            assert trial1.to_dict() == trial2.to_dict(), 'Trials should match after insert'
+
+    def test_register_lie(self, storage):
```
bouthilx: Should add a test where trial lie cannot be registered because of DuplicateKeyError.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test register lie"""
+        with OrionState(experiments=[base_experiment], database=storage) as cfg:
+            storage = cfg.storage()
+            storage.register_lie(Trial(**base_trial))
+
+    def test_reserve_trial(self, storage):
```
bouthilx: Should add a test where reservation fails.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        with OrionState(
+                experiments=[base_experiment], trials=[base_trial], database=storage) as cfg:
+            storage = cfg.storage()
+            experiment = cfg.get_experiment('default_name', 'default_user', version=None)
+
+            trial = storage.reserve_trial(experiment)
+            assert trial is not None
+            assert trial.status == 'reserved'
+
+    def test_fetch_trials(self, storage):
```
bouthilx: Should test both experiment and uid. With each of them, both of them and none of them passed.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test fetch experiment trials"""
+        with OrionState(
+                experiments=[base_experiment], trials=generate_trials(), database=storage) as cfg:
+            storage = cfg.storage()
+            experiment = cfg.get_experiment('default_name', 'default_user', version=None)
+
+            trials = storage.fetch_trials(experiment)
+            assert len(trials) == len(cfg.trials), 'trial count should match'
+
+    def test_get_trial(self, storage):
```
bouthilx: Should test both trial and uid. With each of them, both of them and none of them passed.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test update heartbeat"""
+        with OrionState(experiments=[base_experiment],
+                        trials=generate_trials() + [make_lost_trial()], database=storage) as cfg:
+            storage = cfg.storage()
+
+            experiment = cfg.get_experiment('default_name', 'default_user', version=None)
+            trials = storage.fetch_lost_trials(experiment)
+            assert len(trials) == 1
+
+    def test_retrieve_result(self, storage):
```
bouthilx: Should test when file does not exist and when file content is invalid.

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+            count = 0
+            for trial in cfg.trials:
+                if trial['status'] != 'completed':
+                    count += 1
+
+            assert len(trials) == count
+            for trial in trials:
+                assert trial.status != 'completed'
+
+    def test_fetch_completed_trials(self, storage):
```
bouthilx:
```suggestion
    def test_fetch_trial_by_status(self, storage):
```

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+        """Test update heartbeat"""
+        with OrionState(
+                experiments=[base_experiment], trials=generate_trials(), database=storage) as cfg:
+            storage = cfg.storage()
+
+            trial1 = storage.get_trial(cfg.get_trial(0))
+            storage.update_heartbeat(trial1)
+
+            trial2 = storage.get_trial(trial1)
+            assert trial1.heartbeat != trial2.heartbeat
```
bouthilx:
```suggestion
            assert trial1.heartbeat < trial2.heartbeat
```

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,363 @@
...
+                    count += 1
+
+            storage = cfg.storage()
+
+            experiment = cfg.get_experiment('default_name', 'default_user', version=None)
+            trials = storage.count_broken_trials(experiment)
+
+            assert trials == count
+
+    def test_update_heartbeat(self, storage):
```
bouthilx: Should test with a trial that was set to something else than `reserved` meanwhile.

tests/unittests/core/test_trial.py
bouthilx: I agree that isn't critical. Feel free to drop this one as well as pfmt.

src/orion/storage/base.py
bouthilx: But which one has priority.

tests/unittests/storage/test_storage.py
bouthilx: Alright

tests/unittests/storage/test_storage.py
Delaunay: the update of heartbeat does not depend on the status anymore

src/orion/storage/base.py
Delaunay: yes

src/orion/core/evc/experiment.py
```diff
@@ -129,57 +128,60 @@ def tree_name(self):
...
+        return self._fetch_trials('fetch_noncompleted_trials')
 
-    rval = {'trials': experiment_trials, 'experiment': experiment_node.item}
+    def fetch_trials_by_status(self, status):
+        """See :meth:`orion.core.evc.experiment:Experiment._fetch_trials`"""
+        return self._fetch_trials('fetch_trials_by_status', status=status)
 
-    return rval, parent_or_children
+    def _fetch_trials(self, fun_name, *args, **kwargs):
+        """Fetch trials recursively in the EVC tree using the fetch function `fun_name`
```
DatCorno:
```suggestion
        """Fetch trials recursively in the EVC tree using the fetch function `fun_name`.
```

src/orion/core/evc/experiment.py
```diff
@@ -129,57 +128,60 @@ def tree_name(self):
...
 
+        Parameters
+        ----------
+        fun_name: callable
+            Function name to call to fetch trials. The function must be an attribute of
+            :class:`orion.core.worker.experiment:Experiment`
```
DatCorno:
```suggestion
            :class:`orion.core.worker.experiment:Experiment`.
```

src/orion/core/evc/experiment.py
```diff
@@ -129,57 +128,60 @@ def tree_name(self):
...
 
-def fetch_trials_tree(experiment_node, query, selection=None):
-    """Fetch trials recursively from an experiment node
+        *args:
+            positional arguments to pass to `fun_name`
```
DatCorno:
```suggestion
            Positional arguments to pass to `fun_name`.
```

src/orion/core/evc/experiment.py
```diff
@@ -129,57 +128,60 @@ def tree_name(self):
...
 
-    .. seealso::
+        **kwargs
+            keyword arguments to pass to `fun_name
```
DatCorno:
```suggestion
            Keyword arguments to pass to `fun_name.
```

src/orion/core/evc/experiment.py
```diff
@@ -129,57 +128,60 @@ def tree_name(self):
...
-
-    children_trials_tree = experiment_node.map(
-        functools.partial(_fetch_node_trials, query=query, selection=selection),
-        experiment_node.children)
-    children_trials_tree.set_parent(parent_trials_tree)
-
-    return children_trials_tree
+        """
+        def retrieve_trials(node, parent_or_children):
+            """Retrieve the trials of a node/experiment"""
```
DatCorno:
```suggestion
            """Retrieve the trials of a node/experiment."""
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -8,20 +8,54 @@
...
 from orion.core.io.database import AbstractDB
 from orion.core.io.database.ephemeraldb import EphemeralDB
 
+log = logging.getLogger(__name__)
 
 DEFAULT_HOST = os.path.join(orion.core.DIRS.user_data_dir, 'orion', 'orion_db.pkl')
 
 
+def find_unpickable_doc(dict_of_dict):
+    """Look for a dictionary that cannot be pickled"""
```
DatCorno:
```suggestion
    """Look for a dictionary that cannot be pickled."""
```

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+    return {
+        'storage_type': 'legacy',
+        'type': 'PickledDB',
+        'host': filename
+    }
+
+
+# pylint: disable=no-self-use,protected-access
+class OrionState:
+    """Setup global variables and singleton for tests
```
DatCorno:
```suggestion
    """Setup global variables and singleton for tests.
```

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+
+    it swaps the singleton with none at startup and restore them after the tests.
```
DatCorno:
```suggestion
    It swaps the singleton with `None` at startup and restores them after the tests.
```

src/orion/core/utils/tests.py
```diff
@@ -0,0 +1,212 @@
...
+    It also initialize PickleDB as the storage for testing.
```
DatCorno:
```suggestion
    It also initializes PickleDB as the storage for testing.
```

src/orion/core/worker/consumer.py
```diff
@@ -27,6 +27,12 @@ def _handler(signum, frame):
     raise KeyboardInterrupt
 
 
+class ExecutionError(Exception):
+    """Error raised when orion is unable to execute the user' script without errors"""
```
DatCorno:
```suggestion
    """Error raised when Orion is unable to execute the user's script without errors."""
```

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,451 @@
...
+            experiment = cfg.get_experiment('default_name', 'default_user', version=None)
+
+            trial = storage.reserve_trial(experiment)
+            assert trial is not None
+            assert trial.status == 'reserved'
+
+    def test_reserve_trial_fail(self, storage):
+        """Test reserve trial"""
+        with OrionState(
+                experiments=[base_experiment], trials=[], database=storage) as cfg:
```
bouthilx: It would be better to test with trials that are no reservable.

tests/unittests/storage/test_storage.py
bouthilx: I think you missed this one.

src/orion/storage/legacy.py
bouthilx: Note for reference: `update_heartbeat` now forces `update_trial(..., status='reserved')`.

tests/unittests/storage/test_storage.py
bouthilx: It does seem to: https://github.com/Epistimio/orion/pull/250/files#diff-59f01e95e125eaa93bb5ab10115477f7R281

Delaunay: that is after the update I made to make you happy

tests/unittests/storage/test_storage.py
```diff
@@ -0,0 +1,472 @@
...
+            assert trial1.heartbeat is None
+            assert trial2.heartbeat is not None
+            # this checks that heartbeat is the correct type and that it was updated prior to now
+            assert trial2.heartbeat < datetime.datetime.utcnow()
+
+            if storage_name is None:
+                trial3 = storage.fetch_trial_by_status(status='completed')
+                storage.update_heartbeat(trial3)
+
+                assert trial3.heartveat is None, \
```
bouthilx:
```suggestion
                assert trial3.heartbeat is None, \
```

PR #251, opened 2019-08-09
Migrate Database to new config

bouthilx:
Now configuration is in `orion.core.config` with database configuration. Next step is to migrate `evc` configuration and then we will refactor `experiment_builder` and clean out much of `resolve_config`.


bouthilx:
[Migrate database configuration into `orion.core.config`](https://trello.com/c/gamgwsuI/195-migrate-database-configuration-into-orioncoreconfig)

tests/orion_config.yaml
```diff
@@ -0,0 +1,4 @@
+database:
+    type: 'mongodb'
+    name: 'orion_test'
+    host: 'mongodb://user:pass@localhost'
```
DatCorno: Is there a way to bring a more general conftest higher up at `tests/` level to provide these fixtures that are universally used throughout both `functional` and `unittests`?

PR #253, opened 2019-08-14
Add the search space to the `status`command

DatCorno:


bouthilx:
After discussing offline on how we could support many hps (possibly with long names) we decided it would be too complicated to support this for now, for what it's worth. I'll close this PR until we decide to tackle this again.

bouthilx:
[Add the search space to `status` command.](https://trello.com/c/NM5Ps3Qx/236-add-the-search-space-to-status-command)

PR #257, opened 2019-08-16
Make Space Sorted instead of Ordered

bouthilx:
Why:

The reading of configuration files sometimes return the same data in a
different order (maybe python 3.5 only). Nevertheless, even with a
python API, the user passing a differnt order of hyper-parameter should
not lead to different spaces. Instead of fixing the problem upfront in
the loading of the config, we may fix directly in the space itself and
make it return the dimensions in alphabetical order.


bouthilx:
[Make Space Sorted instead of Ordered](https://trello.com/c/SptzlQLT/253-make-space-sorted-instead-of-ordered)

src/orion/algo/space.py
```diff
@@ -825,6 +824,18 @@ def __repr__(self):
...
+        """Return items sorted according to keys"""
+        return [(k, self[k]) for k in self.keys()]
+
+    def values(self):
+        """Return values sorted according to keys"""
+        return [self[k] for k in self.keys()]
+
+    def keys(self):
+        """Return sorted keys"""
+        return list(sorted(super(Space, self).keys()))
```
DatCorno: Wouldn't it be more efficient to store an internal list of sorted elements and just insert at the right place when adding new items instead of sorting everytime?

bouthilx: Yes but slightly more complicated and then we assume that nothing messes with the ordering during the life of the space. While in comparison this solution ensures the ordering no matter what, for a cost that is most likely unmeasurable considering the small size of spaces. I'd say if this ever turns out to be a bottleneck it should be easy to revisit and improve without affecting the interface.

PR #263, opened 2019-08-20
Fix iter(space)

bouthilx:
Why:

The space was supposed to have sorted keys, values and items, but `iter`
was forgotten and this caused issues in cmdline `insert` in which order
of space.keys() did not match order of `iter(space)`.


bouthilx:
[Fix iter(space)](https://trello.com/c/yhObuq4i/260-fix-iterspace)

PR #264, opened 2019-08-20
Remove `user` as an index for Experiment

DatCorno:


DatCorno:
[Fix --user for list, status and info.](https://trello.com/c/VfwCmQxB/255-fix-user-for-list-status-and-info)

PR #266, opened 2019-08-23
Add to algo the hability to opt out

bouthilx:
Why:

Some algos cannot make valuable predictions until some processes are
completed. For example, at some point ASHA fills the last rungs and
running new trials in the first rungs is pointless and a waste of
ressources.

How:

Algo.suggest may return None to opt out of suggest. The producer then
waits approximately 1 second and retry. It will continue doing so until
it reaches `max_idle_time`.


bouthilx:
[Let algorithm opt out](https://trello.com/c/H6Rn4BY3/205-let-algorithm-opt-out)

PR #267, opened 2019-08-23
 Make ASHA opt out when last rungs are filled

bouthilx:
Depends on #266 

Why:

When running the trials in the last rungs, it takes a significant amount
of time. During this time, all new trials in lower rungs won't be able
to make up to top rungs, as the final rungs will get completed
meanwhile. This results in significant waste of resources which could be
freed and made available for others.

How:

When penultimate rung is filled, no more trials are promoted nor sampled
at first level. This is because all trials required to select the final
one are already available and no more computations are required beside
completing the current trials in the penultimate rung.


bouthilx:
[Limit promotions in ASHA](https://trello.com/c/FcsChFgv/204-limit-promotions-in-asha)

PR #270, opened 2019-08-26
Handle status --name <exp> when empty

bouthilx:
Why:

The command would crash when no experiment matches the name passed with
--name.


bouthilx:
[Handle status --name <exp> when empty](https://trello.com/c/Z2cH11QU/282-handle-status-name-exp-when-empty)

PR #273, opened 2019-08-27
Fix backward compatibility for metadata.user index

bouthilx:
### Add `index_information` and `drop_index` to DB
Why:

The index metadata.user will be removed in v0.1.6, but this will cause
issues with experiments from previous versions since the index will
still be present.

How:

Use the same API as mongodb for both, with a simplified
`index_information` where the output is simply a dictionay mapping
{name}_{order} to {True, False} whether the index is unique or not.

###  Make legacy backward compatible
Why:

Version 0.1.6 will remove the index metadata.user, so we need to make
sure we remove it from the db as well.

### DIFF
--- a/src/orion/core/io/database/__init__.py
+++ b/src/orion/core/io/database/__init__.py
@@ -19,6 +19,7 @@ import logging
 from orion.core.utils import (AbstractSingletonType, SingletonFactory)
 
 
+# pylint: disable=too-many-public-methods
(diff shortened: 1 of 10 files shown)


bouthilx:
[Fix backward compatibility with index because of drop of user](https://trello.com/c/GcEKrvzE/286-fix-backward-compatibility-with-index-because-of-drop-of-user)

PR #274, opened 2019-08-28
Fix ASHA termination condition and branching with fidelity()

bouthilx:
### Improve documentation of ASHA
Why:

There was no documentation on the fact that ASHA requires the
SubParallelStrategy. There was also no documentation at all about the
parallel strategies...

### Stop producer.produce when algo is done
Why:

If an algo is done, but only *realizes* during an update in `produce`,
then the algo will opt out indefinitely and the process will crash. This
should be handled and leave gracefully.

How:

Check for algo.is_done at each loop iteration in `produce` and stop if
True.

### Add default_value to fidelity 
Why:

The EVC is using the default_value to build adapters. Also this default
value can be used as a fall-back if the algorithms does not support
multi-fidelity optimization. This is handy is we want to try different
algorithms with the same experiment configuration.

### Fix ASHA termination condition
Why:

The termination condition was wrong and would stop when the penultimate
rung was filled, making it impossible to ever run the final trial with
max resources.

### DIFF
--- a/docs/src/user/algorithms.rst
+++ b/docs/src/user/algorithms.rst
@@ -2,6 +2,10 @@
 Setup Algorithms
 ****************
 
+.. contents::
+   :depth: 2
+   :local:
+
 Default algorithm is a random search based on the probability
 distribution given to a search parameter's definition.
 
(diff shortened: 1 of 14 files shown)


bouthilx:
[Fix branching with `fidelity()`](https://trello.com/c/oDJ2eOzx/287-fix-branching-with-fidelity)

PR #277, opened 2019-08-29
Avoid branching in `info`

bouthilx:
Why:

When something is different in the environment, the experiment may
branch since we are not using a `View` in `info`. We should rather use a
view but for now they don't have access to instantiated space and
algorithms which are required by `info`.

How:

Use a view and reconfigure it to instantiate the space and algo. This is
not done by default in `View` because they crash if the users scripts in
the experiment are not present on the current FS. At least now this only
happens in `info`, which was already the case anyway since we were not
using a `View`.

### DIFF
--- a/src/orion/core/cli/info.py
+++ b/src/orion/core/cli/info.py
@@ -10,6 +10,7 @@
 
 """
 import logging
+import sys
 
 from orion.core.cli.base import get_basic_args_group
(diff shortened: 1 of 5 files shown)


bouthilx:
[Fix orion info to avoid branching](https://trello.com/c/YaUryQze/289-fix-orion-info-to-avoid-branching)

PR #279, opened 2019-08-30
Fix orion status --name --version

bouthilx:
Why:

The command was not selecting last version or specifed version when
using --name.


bouthilx:
[Fix orion status --name --version](https://trello.com/c/APAWQObt/288-fix-orion-status-name-version)

PR #280, opened 2019-08-30
Add env-vars to consumer

bouthilx:
Note: Documentation missing.

### Add env-vars to consumer

Why:

The user may require the experiment name, experiment id, trial id or
working-dir through another mean than commandline arguments. The
environment variables seams a legitimate way to communicate this
information.


bouthilx:
[Add env-vars to consumer](https://trello.com/c/ccGVrjem/293-add-env-vars-to-consumer)

PR #281, opened 2019-09-10
Handle race conditions during version increment

bouthilx:
Why:

If 2 experiments are created simultaneously and both lead to the same
version increment, one may be created as version=1 and then fail to
resolve the name conflict because version=2 has been registered
meanwhile. This should be resolved by raising an error signaling a race
condition to force the experiment builder to restart the experiment
creation from scratch, as it is done to handle race conditions for
experiment registration during experiment creation.

How:

During conflict detection in experiment.configure, if the
ExperimentNameConflict is not resolved and version number was not
requested (experiment should take last version), then we raise a
RaceCondition error to signal the race condition.

### DIFF
--- a/src/orion/core/io/experiment_builder.py
+++ b/src/orion/core/io/experiment_builder.py
@@ -91,7 +91,7 @@ import logging
 
 from orion.core.io import resolve_config
 from orion.core.io.database import DuplicateKeyError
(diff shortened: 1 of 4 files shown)


bouthilx:
[Fix race condition during auto-resolution](https://trello.com/c/thzTcqN0/295-fix-race-condition-during-auto-resolution)

PR #284, opened 2019-09-16
Handle non-existing fields in EphemeralDB

bouthilx:
Why:

The expected behavior when a selection is done on a non-existing field
is that it will be returned with the value None. Similarly, if an index
is created for a non-existing field it should be considered as being
equal to None.

How:

In the selection, verify if there was any match for a selection key, if
not, assign it None. For the non-existing index, document[key] should
return None when document does not contain key.


bouthilx:
[Handle non-existing fields in EphemeralDB](https://trello.com/c/z8v2rpNH/301-handle-non-existing-fields-in-ephemeraldb)

bouthilx:
@fosterrath-mila

PR #285, opened 2019-09-23
Add functional tests for backward-compatibility

bouthilx:
### Why:

During release v0.1.6 two bugs have been introduced that made it
impossible to reuse data created with v0.1.5 with the PickledDB
backend. This kind of problems should be automatically detected when
making changes in the develop branch.

### How:

For each previous versions (starting with 0.1.3), a db is created using
the old version (0.1.3, 0.1.4, etc) and then all commands are tested in
the current version (test-db, list, status, hunt, etc). The outputs are
not tested thouroughly, this should be covered by the unit-tests. The
tests here only verify that commands run successfully. All tests are run
for each db backend.

### Note:

This should fail in the PickledDB tests until #284 is merged in `develop`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,13 @@ jobs:
         dist: xenial
         sudo: true
       - <<: *test
-        env: TOXENV=demo_random
+        env: TOXENV=demo-random
+        python: 3.6
+      - <<: *test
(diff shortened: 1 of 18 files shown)


bouthilx:
[Add functional tests for backward compatibility](https://trello.com/c/kNOtOWa9/305-add-functional-tests-for-backward-compatibility)

PR #288, opened 2019-10-04
Save parser state in DB

bouthilx:
Why:

We need the parser state to be able to resume the experiment without the
configuration file. It is also convenient for monitoring scripts where
the configuration file often isn't available locally.
Finally, this is a step in direction of support for python API, as the
prior dict in metadata[prior] will serve as the config used to create
the space instead of metadata[user_args]. This will uniformize how
cmdline and python APIs are handled.

How:

OrionCmdlineParser, CmdlineParser and all Converters now have
`get_state_dict()` and `set_state_dict()` methods so that we can save
the state in metadata and restore the parser without the cmdline and config file.

### DIFF
--- a/src/orion/core/evc/conflicts.py
+++ b/src/orion/core/evc/conflicts.py
@@ -56,7 +56,9 @@ import pprint
 import traceback
 
 from orion.algo.space import Dimension
+import orion.core
 from orion.core.evc import adapters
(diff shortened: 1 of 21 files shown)


bouthilx:
[Save parser state in db for standalone resuming](https://trello.com/c/cWPjc3ic/307-save-parser-state-in-db-for-standalone-resuming)

PR #289, opened 2019-10-04
Track integration2

Delaunay:


src/orion/storage/base.py
```diff
@@ -82,7 +82,13 @@ def register_lie(self, trial):
         raise NotImplementedError()
 
     def reserve_trial(self, experiment):
-        """Select a pending trial and reserve it for the worker"""
+        """Select a pending trial and reserve it for the worker
+
+        Returns
+        -------
+        Returns the reserved trial of None if no trials were found
```
```suggestion
        Returns the reserved trial or None if no trials were found
```

src/orion/core/utils/tests.py
```diff
@@ -146,45 +178,97 @@ def get_trial(self, index):
 
     def cleanup(self):
         """Cleanup after testing"""
-        self.database.remove('experiments', {})
-        self.database.remove('trials', {})
+        if self.database:
```
bouthilx: Should we create child classes of OrionState for Legacy and Track instead? Handling both in the same class seems to be leading to some hacky fixes.

src/orion/core/worker/trial.py
```diff
@@ -170,8 +171,8 @@ def __init__(self, **kwargs):
 
         self.status = 'new'
 
-        # Remove useless item
-        kwargs.pop('_id', None)
+        # Store the id as an override to support different backends
+        self.id_override = kwargs.pop('_id', None)
```
bouthilx: This should be added to the docstring of the class below `experiment`.

src/orion/core/worker/trial.py
```diff
@@ -256,7 +257,9 @@ def status(self, status):
     @property
     def id(self):
         """Return hash_name which is also the database key `_id`."""
-        return self.__hash__()
+        if self.id_override is None:
+            return self.__hash__()
+        return self.id_override
```
bouthilx: This hash is required to ensure uniqueness of trial parameters. If we can't use the `id` in Track for this, than we need to save the hash and use it as the unique key. Is this what happens [here]( https://github.com/Epistimio/orion/pull/289/files#diff-e326ee3c119d2fa7c784968b91906ce6R445)?

Delaunay: IIRC, for legacy you just discard '_id' when you get it from the db because you can reliably recompute it using the hash function.

but for track, the _id is made unique by increasing the version tag for a given hash and the primary key is `<hash, tag>`.
so I cannot just discard the `<hash, tag>` returned by the db as I cannot guess the version tag.

Delaunay: just an integer that increases if the trial has already been completed

bouthilx: Is there any way this wrapper forces Orion to only use version 1 then? Because there should not be 2 trials with the same configuration in Orion.

Delaunay: Track takes the latest one so Orion only sees one

bouthilx: Ok so the question is, is there any mechanism to ensure that within Oríon if we try to register a trial with a configuration identical to one already in the db it will be detected and raise a DuplicateKeyError (and not update the version tag and return the latest version).

src/orion/storage/track.py
```diff
@@ -0,0 +1,705 @@
...
+    # return '/' + name
+    return name
+
+
+def to_epoch(date):
+    """Convert datetime class into seconds since epochs"""
+    return (date - datetime.datetime(1970, 1, 1)).total_seconds()
+
+
+class _WarnOnce:
```
bouthilx: Should use `warnings.warn` instead of `log.warning` then.

src/orion/storage/track.py
```diff
@@ -0,0 +1,705 @@
...
+        heartbeat = to_json(trial.heartbeat)
+        if heartbeat is None:
+            heartbeat = 0
+        metadata['heartbeat'] = heartbeat
+
+        metrics = defaultdict(list)
+        for p in trial.results:
+            metrics[p.name] = [p.value]
+
+        self.current_trial = self.backend.new_trial(TrackTrial(
```
bouthilx: When does `new_trial` return None? It's hard to guess from [here](https://github.com/Delaunay/track/blob/d8ffb0024bc0ee8720f6889c31c92aa989e361d7/track/client.py#L224)

tests/unittests/storage/test_storage.py
```diff
@@ -498,3 +471,75 @@ def test_update_heartbeat(self, storage):
 
                 assert trial3.heartbeat is None, \
                     'Legacy does not update trials with a status different from reserved'
+
+
+class TestLegacyStorage:
```
bouthilx: This should be move to test_legacy.py

src/orion/core/worker/trial.py
Delaunay: if protocol.new_trial(*args, autoincrement=False) then no new trial is created if it already exist.
Note that the behaviour is tested in `unittests/storage/test_storage.py:test_register_duplicate_trial`

PR #290, opened 2019-10-05
Fix deepcopy of Enumerate transform

bouthilx:
Why:

The vectorized function of Enumerate could not be `deepcopy`ed making it
unusable because of naive algorithm's that requires copying the original
algo.

How:

Overwrite `__deepcopy__` the rebuild the Enumerate object instead of
copying it.


bouthilx:
[Fix deepcopy in Enumerate](https://trello.com/c/ctiHheCX/315-fix-deepcopy-in-enumerate)

PR #291, opened 2019-10-08
Turn test-db into db test

bouthilx:
Why:

We will shortly add more database related commands such as `remove` and
`upgrade`. They should all be grouped within `orion db`.


bouthilx:
[Move db commands to `orion db`](https://trello.com/c/JBI1mfyV/313-move-db-commands-to-orion-db)

PR #292, opened 2019-10-08
Update skopt doc and prepare v0.1.7

bouthilx:


bouthilx:
[Update skopt doc](https://trello.com/c/SNGoOGdU/314-update-skopt-doc)

PR #293, opened 2019-10-09
Add `orion db upgrade` command

bouthilx:
Why:

There is breaking changes in the DB scheme (mostly importantly in the
unique indexes definition). To fix this, we will provide a command to
upgrade the db scheme explicitely (not implicitly during execution like
we were first attempting). This will avoid breaking DBs and simplify the
handling of backward compatibility in the db.

How:

- Add checks during DB creation to fail fast if the db is outdated and
incompatible with current version
- Add command `orion db upgrade` to fix the indexes and add missing
fields (version, priors)


bouthilx:
[Add command `orion db upgrade`](https://trello.com/c/CpbQWcD8/316-add-command-orion-db-upgrade)

PR #295, opened 2019-10-09
Instantiate adapter even if empty

bouthilx:
Why:

When handling the configured experiment, we expect the adapter to be
instiantiated even if it does nothing (empty CompositeAdapter).


bouthilx:
[Instantiate adapter even if empty](https://trello.com/c/kkUUMfk9/321-instantiate-adapter-even-if-empty)

PR #297, opened 2019-10-10
 Refactor experiment builder 

bouthilx:
Why:

The construction of an experiment is currently very complicated and
coupled with the argument structure coming from the commandline.
By breaking up the parsing of the commandline arguments and the
configuration of the experiment it becomes both easier to build an
experiment and support a new python API in addition to the commandline
API.

How:

Break down the ExperimentBuilder into few public methods, most
importantly `build()`, `build_view()` and `setup_storage()`. In addition
to this, `build_from_args()` and `build_view_from_args()` makes the
bridge between the commandline API and the new experiment builder
functions.

All of `experiment.configure()` is not taking place inside `build()` and
`build_view()`. The mechanisms are currently duplicated to ensure a
smooth migrate, but ultimately the `Experiment` object will solely play
the role of fetching and updating data. Configuration, construction,
branching now all occur in `experiment_builder`.

PR #299, opened 2019-10-15
Add Space definition in DB

bouthilx:
Note:

Depends on #297 

Why:

The space definition used to be implicitely stored in
metadata[user_args] when the only API was the commandline one. With the
python API it becomes much more convenient to store the space definition
in the DB explicitely as it is how the user will provide it.

How:

If using the commandline API, parse the user_args, build the space and
save the configuration of the space in DB. If using the python API,
simply save the space in DB directly.

### DIFF
--- a/src/orion/algo/space.py
+++ b/src/orion/algo/space.py
@@ -31,7 +31,7 @@ scipy documentation for each specific implentation of a random variable type,
 unless noted otherwise!
 
 """
-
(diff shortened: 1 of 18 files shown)


bouthilx:
[Add space to database](https://trello.com/c/1akjpvnt/197-add-space-to-database)

PR #303, opened 2019-10-17
Drop python3.5 support and add python3.8

bouthilx:
Why:

Python 3.8 was released a few days ago.

For python 3.5, it lack of builtin ordered dictionary is growing
cumbersome and likely not worth the hassle.

PR #304, opened 2019-10-17
Turn trial params to dicts

bouthilx:
Note: py35 will fail until removed in #303 

Why:

The list structure is very cumbersome and does not match the interface
provided to users to define space. The params structure as a dict is
more coherent.


bouthilx:
[Turn params into dict](https://trello.com/c/7Npz9BzO/329-turn-params-into-dict)

bouthilx:
Close in favor of #305

PR #305, opened 2019-10-18
Expose trial.params as a dict

bouthilx:
Why:

The list structure is very cumbersome and does not match the interface
provided to users to define space. The params structure as a dict is
more coherent. The ordering of the params is important however for the
hashing of the trial, hence we keep the list as a private attribute and
only expose a dict of {name:value} of the trials, which is more
convenient for the user interface.


bouthilx:
[Turn params into dict](https://trello.com/c/7Npz9BzO/329-turn-params-into-dict)

PR #306, opened 2019-10-18
Fix EphemeralDB index update after document deletion

bouthilx:
Why:

When remove documents, the corresponding indexes would not be deleted,
making impossible to insert new documents with these values.


bouthilx:
[Fix EphemeralDB index update after document deletion](https://trello.com/c/fkFSa3Tf/332-fix-ephemeraldb-index-update-after-document-deletion)

PR #307, opened 2019-10-22
Handle broken trials with python API

bouthilx:
Why:

If the process exits, the trial status stays to `reserved` and will
eventually be set to `interrupted`.

How:

Register a function to release the remaining trials with pacemakers. If
the process exits normally (no signup signals or sigterm), the function
will be called and all reserved trials will be set to broken.

### DIFF
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -8,6 +8,8 @@
    :synopsis: Wraps the core Experiment object to provide further functionalities for the user
 
 """
+import atexit
+import functools
 import logging
 
 from numpy import inf as infinity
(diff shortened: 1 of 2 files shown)


bouthilx:
[Handle broken trials with python API](https://trello.com/c/A1kRqDur/335-handle-broken-trials-with-python-api)

bouthilx:
Codecov test is totally broken, coverage is not decreased by the PR.

Delaunay:
Actually if we are going to have an explicit `experiment.close()` I would be in favor of implementing it as a context as well with `__enter__` and `__exit__` so users do not have to think about it

bouthilx:
I agree that having support for close through a context manager would be great, but since the call to `close()` is not mandatory (not doing so just clutters atexit but it doesn't cause any issue), I think forcing the use of the context manager would be too much hassle for the user for what it's worth. Maybe we can make a trello card to add support for both explicit `close()` and contextmanager, like builtin `open()`.

PR #308, opened 2019-10-25
Dev/ignore cli options

mirkobronzi:


bouthilx:
[Add configurable arguments to ignore (like a tmp-path specific to a job)](https://trello.com/c/Byfgpau1/61-add-configurable-arguments-to-ignore-like-a-tmp-path-specific-to-a-job)

PR #309, opened 2019-10-28
Fix doc for bayesopt's alpha

bouthilx:
Why:

The format 1e-10 is not a valid format for scientific numbers in yaml
and will be converted to string.


bouthilx:
[Fix documentation of conf for bayesopt](https://trello.com/c/iftc1qTK/346-fix-documentation-of-conf-for-bayesopt)

PR #311, opened 2019-11-06
Add support for space hierarchical structure

bouthilx:
Why:

The python API is convenient to use with a hierarchical space structure
if there is many dimensions, just as we would do in configuration files.
Note that this does not support conditional dimensions, each dimension
is independent, but the space is represented in a hierarchical
structure. See contributed unit-tests for an example.


bouthilx:
[Support hierarchical space structure](https://trello.com/c/siVbYO4V/367-support-hierarchical-space-structure)

PR #312, opened 2019-11-08
Fix formatting of trial with subdict params

bouthilx:
Why:

The formating was not flattening the params when turning them to tuple
for algo.observe.


bouthilx:
[Fix formatting of trial with subdict params](https://trello.com/c/oQS2Jvtl/370-fix-formatting-of-trial-with-subdict-params)

PR #313, opened 2019-11-08
Dev/recover experiment

mirkobronzi:
Adds a `hash_params` id that does *not* include the fidelity.

Useful (e.g.,) to allow asha to recover an experiment (where the parameters are the same but the fidelity is different) and run it without starting from scratch.


bouthilx:
[Add trial.hash_params](https://trello.com/c/f0GbSMQW/368-add-trialhashparams)

PR #314, opened 2019-11-08
Support small fidelity scales for ASHA

bouthilx:
Why:

If the scale is too small and there is many rungs, then some fidelity
levels will have the same values. We can catch this and increase the
wrong ones to avoid duplicates.


bouthilx:
[Support small fidelity scales for ASHA](https://trello.com/c/SVzGeL1n/371-support-small-fidelity-scales-for-asha)

PR #318, opened 2019-11-14
Fix track  backend API

Delaunay:


src/orion/storage/track.py
```diff
@@ -393,6 +407,8 @@ def fetch_experiments(self, query, selection=None):
         for k, v in query.items():
             if k == 'name':
                 new_query['metadata.name'] = v
+                # FIXME
```
bouthilx: Is it fixed? What is to be fixed? :)

Delaunay: it is ugly

src/orion/storage/track.py
```diff
@@ -443,7 +459,7 @@ def register_trial(self, trial):
         for p in trial.results:
             metrics[p.name] = [p.value]
 
-        self.current_trial = self.backend.new_trial(TrackTrial(
```
bouthilx: No more current trial?

Delaunay: it was not used anyway

PR #319, opened 2019-11-20
Algo redefinition should not drop algo config

bouthilx:
Why:

When the algo is passed again, it resulted in dropping the algo
configuration altogether. This lead to a reversal to RandomSearch and thus
the experiment would have a different algo and branch.

How:

Overwrite config[algo] with the user's defined algorithm since it has
precedence over db anyway.


bouthilx:
[Algo redefinition should not drop algo config](https://trello.com/c/bbdXI2NY/377-algo-redefinition-should-not-drop-algo-config)

PR #321, opened 2019-12-07
Add hash params to Track API

Delaunay:


src/orion/storage/track.py
```diff
@@ -275,6 +275,17 @@ def results(self):
 
         return self._results
 
+    @property
+    def hash_params(self):
+        from orion.core.worker.trial import Trial as OrionTrial
+
+        params = [x for x in self._params if x.type != 'fidelity']
+        params_repr = OrionTrial._repr_values(None, params, sep=',')
```
bouthilx: Why not using the Trial.hash_params directly?

PR #322, opened 2019-12-07
Add hash params

Delaunay:


src/orion/core/worker/trial.py
```diff
@@ -294,44 +294,35 @@ def gradient(self):
...
                              "have not been set.")
-        params_repr = self.params_repr(ignore_fidelity=True)
+
+        params = self.params_repr(self._params, ignore_fidelity)
         experiment_repr = str(self.experiment)
-        return hashlib.md5((params_repr + experiment_repr).encode('utf-8')).hexdigest()
+
+        lie_repr = ""
+        if not ignore_fidelity and self.lie:
+            self.values_repr([self.lie])
```
bouthilx:
```suggestion
            lie_repr = self.values_repr([self.lie])
```

src/orion/core/worker/trial.py
```diff
@@ -294,44 +294,35 @@ def gradient(self):
...
 
     @property
     def hash_params(self):
         """Generate a unique param md5sum hash for this `Trial`.
 
         .. note:: The params contributing to the hash do not include the fidelity.
         """
+        return self._hash(ignore_fidelity=True)
+
+    def _hash(self, ignore_fidelity=False):
```
bouthilx: Maybe this should be a public static method well since we plan on using it for the Track backend.

Delaunay: ok

src/orion/storage/track.py
```diff
@@ -275,6 +276,12 @@ def results(self):
 
         return self._results
 
+    @property
+    def hash_params(self):
+        """See `~orion.core.worker.trial.Trial`"""
+        from orion.core.worker.trial import Trial as OrionTrial
+        return OrionTrial.compute_trial_hash(self, ignore_fidelity=True)
```
bouthilx: You turn the property into a staticmethod so that you can pass a duck that quacks like a duck but is not a duck. :thinking:

Delaunay: PYTHONNNNNN am I rite

PR #325, opened 2020-01-25
Misc Modifications

Delaunay:


Delaunay:
>   `return [self.high for i in range(n_samples)]`

This was a fix suggested by @bouthilx that was not in develop.

And the other change was raising an exception instead of returning None when we are waiting for a Trial & re-raising the exception in case of error instead of returning None and silently missing the exception. This was mainly an issue of the python API where user would be clueless about what went wrong.

The code below highlight all the different cases where orion can return None and as you can see it can be hard to differentiate.

Current code does not differentiate between an error and waiting for a trial to finish.

This PR also highlight there are some underlying issue that needs to be addressed.
For example now we receive a RuntimeException from the Producer.
This exception used to be ignored and transformed into None, maybe now it should become something else.

I am not going to be working on it at the moment but I think it brings a useful set of improvements and those should still be considered.

```
class TrialIterator:
    """Take an Orion experiment and iterate through all the trials it suggests

    Parameters
    ----------
    experiment: ExperimentClient
        Orion Experiment
    """
    def __init__(self, experiment, retries=2, client=None):
        self.experiment = experiment
        self.time = StatStream(drop_first_obs=1)
        self.retries = retries
        self.client = client

    def __iter__(self):
        return self

    @property
    def is_finished(self):
        return self.experiment.is_done or self.experiment.is_broken

    def __next__(self, _depth=0):
        from orion.core.worker import WaitingForTrials

        if _depth >= self.retries:
            debug(f'Retried {_depth} times without success')
            raise StopIteration

        if self.experiment.is_broken:
            error('Experiment is broken and cannot continue')
            raise StopIteration

        if self.experiment.is_done:
            info('Orion does not have more trials to suggest')
            raise StopIteration

        start = datetime.utcnow()
        try:
            trial = self.experiment.suggest()

        except WaitingForTrials:
            # Do not increase depth this is not a retry
            time.sleep(10)
            return self.next(_depth)

        self.time += (datetime.utcnow() - start).total_seconds()

        if trial is None:
            if not self.experiment.is_done:
                warning('No additional trials were found but experiment is not done')
                return self.next(_depth + 1)
            else:
                info(f'Orion did not suggest more trials (is_done: {self.experiment.is_done}')
                raise StopIteration

        if self.client is not None:
            self.client.trial = trial

        return trial

    next = __next__
```

Delaunay:
I might finish it over the weekend, depending if I get feedback from @bouthilx.
I need this merged for Olympus

Thomsch:
Thank you for the description, this is very helpful.

The improvements are great. I'll come back to you if you don't get an answer from @bouthilx.

bouthilx:
I like the iterator solution. `orion.core.worker.workon` should also be refactored to use this iterator. I'll add a trello card about the errors raised in Producer to specialize them so they can be differentiated by user.

@Delaunay Could you adapt the tests and doc for the new behavior? (Raise instead of return None)

Delaunay:
Ready for merge.
Test failure is unrelated to the changes.

src/orion/client/experiment.py
```diff
@@ -383,20 +447,55 @@ def suggest(self):
             Reserved trial for execution. Will return None if experiment is done or broken
```
bouthilx:
```suggestion
            Reserved trial for execution. Will return None if experiment is done.
```
And remove rest of sentence.

src/orion/client/experiment.py
```diff
@@ -383,20 +447,55 @@ def suggest(self):
...
+        Raises
+        ------
+        `WaitingForTrials`
+            if the HPO has not finished but no trials are available at the moment
+
+        `BrokenExperiment`
+            if the trials failed to run and the HPO cannot continue
+
+        `SampleTimeout`
+            if the HPO could not sample new unique new points
```
bouthilx:
```suggestion
            if the HPO could not sample new unique points.
```

src/orion/client/experiment.py
```diff
@@ -383,20 +447,55 @@ def suggest(self):
             Reserved trial for execution. Will return None if experiment is done or broken
             of if the algorithm cannot suggest until other trials complete.
 
+        Raises
+        ------
+        `WaitingForTrials`
+            if the HPO has not finished but no trials are available at the moment
+
+        `BrokenExperiment`
+            if the trials failed to run and the HPO cannot continue
```
bouthilx:
```suggestion
            if too many trials failed to run and the experiment cannot continue. This is determined by ``experiment.max_broken`` in the configuration.
```

src/orion/client/experiment.py
```diff
@@ -383,20 +447,55 @@ def suggest(self):
             Reserved trial for execution. Will return None if experiment is done or broken
             of if the algorithm cannot suggest until other trials complete.
 
+        Raises
+        ------
+        `WaitingForTrials`
+            if the HPO has not finished but no trials are available at the moment
```
bouthilx:
```suggestion
            if the experiment is not completed and algorithm needs to wait for some trials to complete 
            before it can suggest new trials.
```

src/orion/client/experiment.py
```diff
@@ -383,20 +447,55 @@ def suggest(self):
...
+                raise BrokenExperiment("Trials failed too many times") from e
+
+            raise e
 
         if trial is not None:
             self._maintain_reservation(trial)
 
         return trial
 
+    def __iter__(self):
```
bouthilx: `def workon` should use this now.

src/orion/client/experiment.py
```diff
@@ -36,6 +38,68 @@ def set_broken_trials(client):
...
+    def next_trial(self, _depth=0):
+        """Iterate through Orion's suggestions"""
+        if _depth >= self.retries and not self.experiment.is_done:
+            log.error(f'Retried {_depth} times without success')
+            raise SampleTimeout(
+                'HPO is not able to sample new unique points after {} retries'.format(
+                    self.retries))
+
+        if self.experiment.is_done:
+            log.info('Orion does not have more trials to suggest')
```
bouthilx:
```suggestion
            log.info('Experiment is completed, no more trials to suggest.')
```

src/orion/algo/space.py
```diff
@@ -767,7 +767,7 @@ def validate(self):
 
     def sample(self, n_samples=1, seed=None):
         """Do not do anything."""
-        return [self.high]
+        return [self.high for i in range(n_samples)]
```
bouthilx: There should be a test for this fix.

src/orion/client/experiment.py
```diff
@@ -36,6 +38,68 @@ def set_broken_trials(client):
         client.release(trial, status='broken')
 
 
+# pylint: disable=too-few-public-methods
+class TrialIterator:
```
bouthilx: The iterator should be tested.

src/orion/core/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+.. module:: exceptions
+   :platform: Unix
+   :synopsis: define public exceptions
+
+"""
+
+
+class SampleTimeout(Exception):
+    """Raises when the algorithm is not able to sample new unique points in time"""
```
bouthilx:
```suggestion
    """Raised when the algorithm is not able to sample new unique points in time"""
```

src/orion/core/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class WaitingForTrials(Exception):
+    """Raised when no trials could be reserved after multiple retries"""
+
+    pass
+
+
+class BrokenExperiment(Exception):
+    """Raised when a trial has been tried too many times without success"""
```
bouthilx:
```suggestion
    """Raised when too many trials failed in an experiment and it is now considered broken"""
```

src/orion/core/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class SampleTimeout(Exception):
+    """Raises when the algorithm is not able to sample new unique points in time"""
+
+    pass
+
+
+class WaitingForTrials(Exception):
+    """Raised when no trials could be reserved after multiple retries"""
```
bouthilx:
```suggestion
    """Raised when the algorithm needs to wait for some trials to complete before it can suggest new ones"""
```

bouthilx:
I agree with the modifications of Producer. I'm not sure we should revert to number of retries instead of timeout however. We used to have retries and it was to dependent on algorithm/context to be reliable across them. If using a slow algo, like Bayesopt with many points, or with PickledDB with many workers than we would need to adjust the number of retries. It was tricky and annoying. Timeout with backoff sleep time instead solved the issue.

Delaunay:
Alright, I wont modify it further. It is too time consuming

PR #326, opened 2020-01-29
Formatted commands and added a commande description

Thomsch:
Why:
A few commands were not formatted and a ``sed`` command was shown without any explanation which is susceptible to deteriorate the readability and comprehension of the example, threatening the engagement of potential and current users.

How:
Formatted the commands similarly to the others in the example and added a minimum descriptions.

PR #327, opened 2020-01-29
Formatted commands and added a command description

Thomsch:
Why:
A few commands were not formatted and a sed command was shown without any explanation which is susceptible to deteriorate the readability and comprehension of the example, threatening the engagement of potential and current users.

How:
Formatted the commands similarly to the others in the example and added a minimum descriptions.


Thomsch:
Can you add the label 'documentation' to this PR too?

Thomsch:
Mmmmh, I'm not sure why it failed for the 3.7 env as it's exactly the same commits in https://github.com/Epistimio/orion/pull/326 which passed the test then.

bouthilx:
Looks like a hickup fro MongoDB server. I restarted the test, we'll see.

PR #331, opened 2020-02-11
PR: add precision to Real

hbertrand:


hbertrand:
I can't pass pylint without some major refactoring of `core.worker.transformer.build_required_space`. I'm not sure what should be done here.

bouthilx:
I'm surprised the error `too-many-branches` was not already silenced, this function is fugly. :sweat_smile: You can add `# pylint: disable=too-many-branches` at the top of `build_required_space` to silence it.

src/orion/algo/space.py
```diff
@@ -346,6 +346,13 @@ def __init__(self, name, prior, *args, **kwargs):
             raise ValueError("Lower bound {} has to be less than upper bound {}"
                              .format(self._low, self._high))
 
+        precision = kwargs.pop('precision', 4)
```
bouthilx: I would add precision in __init__ signature instead and document it in the docstring.

src/orion/core/worker/transformer.py
```diff
@@ -239,6 +244,33 @@ def domain_type(self):
...
 
 
+class Precision(Transformer):
+    """Round real numbers to requested precision."""
+
+    domain_type = 'real'
+    target_type = 'real'
+
+    def __init__(self, precision=4):
+        """Initialize a precision transformation. Precision must be a non-negative integer."""
```
bouthilx: You can drop the docstring in `__init__`, we document arguments in the class docstring. (any documented `__init__` you see were mistakes we did not correct yet). This is following numpy's standard: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

tests/functional/demo/test_demo.py
```diff
@@ -535,6 +535,6 @@ def test_demo_with_nondefault_config_keyword(database, monkeypatch):
     assert len(params) == 1
     assert params[0]['name'] == '/x'
     assert params[0]['type'] == 'real'
-    assert (params[0]['value'] - 34.56789) < 1e-5
+    assert (params[0]['value'] - 34.568) < 1e-5
```
bouthilx: I would add instead a specific test for this (like `test_demo()`) where you only verify that the values are rounded to correct precision.

tests/unittests/algo/test_space.py
```diff
@@ -774,3 +774,18 @@ def test_configuration(self):
             'yolo2': 'uniform(-3, 3, shape=(2,), discrete=True)',
             'yolo3': 'norm(0.9)',
             'yolo4': 'choices([\'asdfa\', 2])'}
+
+    def test_precision(self):
+        """Test that precision is correctly handled."""
+        space = Space()
+        space.register(Real('yolo1', 'norm', 0.9, precision=6))
```
bouthilx: Maybe add a third and verify default value?

tests/unittests/core/test_transformer.py
```diff
@@ -662,14 +705,20 @@ def test_no_requirement(self, space_each_type):
         assert tspace[0].type == 'real'
         assert tspace[1].type == 'categorical'
         assert tspace[2].type == 'integer'
-        assert str(tspace) == str(space_each_type)
+        assert (str(tspace) ==
+                "Space([Precision(Real(name=yolo, prior={norm: (0.9,), {}}, shape=(3, 2), default value=None)),\n"  # noqa
```
bouthilx: I think `Precision.repr_format` should be overriden so that it prints `Precision({precision value}, Real(...))` otherwise we loose the information.

bouthilx:
[Add Precision to Real](https://trello.com/c/yEfOKnmj/399-add-precision-to-real)

tests/functional/demo/test_demo.py
```diff
@@ -538,3 +538,31 @@ def test_demo_with_nondefault_config_keyword(database, monkeypatch):
...
+
+@pytest.mark.usefixtures("clean_db")
+@pytest.mark.usefixtures("null_db_instances")
+def test_demo_precision(database, monkeypatch):
+    """Test a simple usage scenario."""
+    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
+
+    user_args = [
+        "-x~uniform(-50, 50, precision=5)",
+        "--test-env",
```
bouthilx: You can pass only the first line ('x~uniform()') and remove all args below.

tests/functional/demo/test_demo.py
```diff
@@ -538,3 +538,31 @@ def test_demo_with_nondefault_config_keyword(database, monkeypatch):
...
+        "hunt", "--config", "./orion_config.yaml", "./black_box.py"] + user_args)
+
+    exp = list(database.experiments.find({'name': 'voila_voici'}))
+    exp = exp[0]
+    exp_id = exp['_id']
+    trials = list(database.trials.find({'experiment': exp_id}))
+    trials = list(sorted(trials, key=lambda trial: trial['submit_time']))
+    params = trials[-1]['params']
+
+    assert (params[0]['value'] - 34.568) < 1e-5
```
bouthilx: I would simply test that values of params have the proper precision.

src/orion/algo/space.py
hbertrand: If I do that it breaks a bunch of tests for reasons unclear to me. And shouldn't it be with low and high?

bouthilx: Oh, can you share the stacktrace of a test that fails because of this modification?

bouthilx: From offline chat it seems adding other arguments before `*args` and `**kwargs` is not compatible with positional arguments for the priors. For example ``normal(0.9)`` would pass ``0.9`` as the first positional argument to ``Real()`` ( ``precision`` if we added it) and thus ``*args`` would be left empty and the prior would not receive its argument. The priority is to keep the interface simple for users, and thus support positional arguments in prior, so I would agree to use ``kwargs`` as currently done it this PR.

tests/functional/demo/test_demo.py
hbertrand: En fait `params[0]['value']` a une valeur de 34.56795111111093, ce qui est bien plus que demandé. Si je comprends bien la sélection des x est faite par gradient descent dans ce cas, donc j'imagine que ça ignore la précision demandé ?

bouthilx: oh :scream: We just catched an issue. If the algo outputs in lower precision, it should be casted in transformations. But what we do right now is cast it in `transform`, which is before sending to algo, but we do nothing in `reverse`. We should cast in both directions.

hbertrand: The modif to `reverse` causes another problem: `RuntimeError: Algorithm could not sample new points in less than 60 seconds`, because I guess it really wants to sample at values in between the precision

tests/functional/demo/test_demo.py
```diff
@@ -538,3 +538,26 @@ def test_demo_with_nondefault_config_keyword(database, monkeypatch):
...
+@pytest.mark.usefixtures("null_db_instances")
+def test_demo_precision(database, monkeypatch):
+    """Test a simple usage scenario."""
+    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
+
+    user_args = [
+        "-x~uniform(-50, 50, precision=5)"]
+
+    orion.core.cli.main([
+        "hunt", "--config", "./orion_config.yaml", "./black_box.py"] + user_args)
```
bouthilx: This should solve the issue.
```suggestion
        "hunt", "--config", "./orion_config.yaml", "--max-trials", "2", "./black_box.py"] + user_args)
```

PR #333, opened 2020-02-13
Dev/support entry point

mirkobronzi:


bouthilx:
[Support any executable, including `python script.py`](https://trello.com/c/lQYRuPQf/416-support-any-executable-including-python-scriptpy)

PR #336, opened 2020-02-23
Random algorithm will hang if search space is smaller than specified trials number

donglinjy:
This PR include below 2 fix
- There is a doc error for space builder where use 'random' as the prior
- When the search space size is too small, Random will keep proposing repeated samples.


src/orion/algo/space.py
```diff
@@ -568,6 +585,11 @@ def __init__(self, name, categories, **kwargs):
                                                   self._probs))
         super(Categorical, self).__init__(name, prior, **kwargs)
 
+    @property
+    def capacity(self):
+        """Return the capacity size of all the possible values from Categorical `Dimension`"""
+        return len(self.categories)
```
bouthilx: This should depend on shape as well, isn't?

tests/unittests/algo/test_space.py
```diff
@@ -683,6 +683,26 @@ def test_interval(self):
...
+        """Check whether space capacity is correct"""
+        space = Space()
+        probs = (0.1, 0.2, 0.3, 0.4)
+        categories = ('asdfa', 2, 3, 4)
+        dim = Categorical('yolo', OrderedDict(zip(categories, probs)), shape=2)
+        space.register(dim)
+        dim = Integer('yolo2', 'uniform', -3, 6)
+        space.register(dim)
+
+        assert 24 == space.samplescapacity
```
bouthilx: If capacity of categories depends on shape, then this should be (4 * 2) * 6 = 48. I would also write the test as such (with the multiplications instead of directly 48) to ease reading and understanding of the test.

src/orion/algo/random.py
```diff
@@ -62,4 +64,11 @@ def observe(self, points, results):
 
         A simple random sampler though does not take anything into account.
         """
-        pass
+        self.alreadyrun += len(points)
+
+    @property
+    def is_done(self):
```
bouthilx: This is not specific to RandomSearch and would affect any algorithm. I would instead add this check in Experiment.is_done.

src/orion/algo/random.py
```diff
@@ -62,4 +64,11 @@ def observe(self, points, results):
 
         A simple random sampler though does not take anything into account.
         """
-        pass
+        self.alreadyrun += len(points)
```
bouthilx: This would also lead to the issue if there is more workers than sample capacity. We could add it in suggest instead but then we do not know if the sample is valid and will be kept (potential duplicate due to race-conditions). By moving the check of is_done to Experiment.is_done we avoid these issues.

src/orion/algo/space.py
donglinjy: I did not realize that Categorical will also support shape. I will update this.

tests/unittests/algo/test_space.py
donglinjy: Agree, I will make the change

src/orion/algo/random.py
donglinjy: I was thinking to such capacity check in Experment.is_done too, but consider that different algorithm may have different ways to decide whether the search space is already empty or not, like for Hyperband. So I decided to put this in the Random first and we can think if could define a generic way for all.

src/orion/algo/random.py
donglinjy: Yes, do this in suggestion will have the duplicate issue.  As for more workers than sample capacity, my understanding is: each worker will be able to observe that global completed points and some of them may propose duplicated samples, but according to the producer logic, after backoff, the is_done logic should be True at each workers.

donglinjy:
> Thanks for the PR @donglinjy ! I agree the fixes, there is only a few minor modifications required. Do you think `cardinality` instead of `capacity` would be more accurate?

Yeah, cardinality looks better, I will rename it. thanks

src/orion/algo/random.py
bouthilx: Indeed some algo may have specific is_done procedures, but the one based on fully explored search space is generic and thus should be in Experiment in my opinion. Algorithms will either be done before search space is fully explored or in worst case when it gets fully explored. So in some sense algorithm.is_done have precedence over experiment.is_done.

src/orion/algo/random.py
bouthilx: Yes, mostly, but this depends on the parallel strategy. If there is none, then non-completed trials are not passed to observe and the algo will exhaust the search space.

bouthilx:
[Random algorithm will hang if search space is smaller than specified trials number](https://trello.com/c/zkPpXwuX/424-random-algorithm-will-hang-if-search-space-is-smaller-than-specified-trials-number)

PR #339, opened 2020-02-25
Minimalist example with scikit-learn

Thomsch:
Adds a minimalist example of integrating a scikit-learn based script to Oríon.


bouthilx:
[Rédiger tutoriel dans la Doc Sphinx de Orion.](https://trello.com/c/3SgXcP0r/388-r%C3%A9diger-tutoriel-dans-la-doc-sphinx-de-orion)

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+The script is divided in five parts:
+
+#. Parsing of the script arguments
+#. Loading and splitting the dataset
+#. Training a classifier using the researcher-defined *epsilon*
+#. Evaluating the classifier using the testing set
+#. Reporting the performance of the model. i.e., the accuracy.
+
+.. note::
+   The workflow presented in the script is simplified compared to the reality on purpose. The
```
bouthilx:
```suggestion
   The workflow presented in the script is simplified on purpose compared to real ones. The
```

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+   objective of this example is to illustrate the basic steps involved in using Oríon.
+
+To find a good *epsilon*, a user would run empirically ``$ python main.py <epsilon>`` multiple
+times, choosing a new value for *epsilon* manually.
+
+This ad-hoc hyper-parameter optimization is unreliable, slow, and requires a lot of work from the
+user. Oríon solves this problem by providing cutting-edge hyper-parameter optimization
```
bouthilx:
```suggestion
user. Oríon solves this problem by providing established hyper-parameter optimization
```
We prioritize established and reliable algorithms.

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+
+For the former, this step takes place in the script training the model. The latter can either be
+specified in a configuration file or directly while calling the script with Oríon.
+For the purpose of the example, we'll configure the hyper-parameter space directly as a
+command-line argument.
+
+Updating the script
+^^^^^^^^^^^^^^^^^^^
+We only need to make one small change to the script: we report to Oríon the objective that we
+want to **minimize** at the end of the script using :py:func:`orion.client.report_results`:
```
bouthilx:
```suggestion
want to **minimize** at the end of the script using :py:func:`orion.client.report_objective`:
```

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. literalinclude:: ../../../examples/scikitlearn-iris/main.py
+   :language: python
+   :lines: 32-
+
+In our example, we measure the accuracy of the model to qualify its performance. To get the best
+accuracy possible, we need to minimize the difference between 1 and the accuracy to get it as
+close to 1 as possible. Otherwise, we'll be minimizing the accuracy which will yield a poor model.
+
+:py:func:`orion.client.report_results` can be imported using :
```
bouthilx:
```suggestion
:py:func:`orion.client.report_objective` can be imported using :
```

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. code-block:: python
+
+   from orion.client import report_results
```
bouthilx:
```suggestion
   from orion.client import report_objective
```

examples/scikitlearn-iris/main.py
```diff
@@ -0,0 +1,35 @@
+import sys
+
+import numpy as np
+from orion.client import report_results
```
bouthilx:
```suggestion
from orion.client import report_objective
```

examples/scikitlearn-iris/main.py
```diff
@@ -0,0 +1,35 @@
...
+clf.fit(X_train, y_train)
+
+# Evaluating the accuracy using the testing set.
+y_pred = clf.predict(X_test)
+accuracy = balanced_accuracy_score(y_test, y_pred)
+
+# Reporting the results
+print("Accuracy is {}".format(accuracy))
+
+report_results([dict(
```
bouthilx: Needs to be adapted to `report_objective`. Basically, simply `report_objective(1 - accuracy)`.

tests/functional/example/orion_config.yaml
```diff
@@ -0,0 +1,12 @@
+name: scikit-iris-tutorial
+
+max_trials: 1
+
+algorithms:
+  random:
+    seed: 1
+
+database:
```
bouthilx: We should probably use PickledDB for the example.

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+Updating the script call
+^^^^^^^^^^^^^^^^^^^^^^^^
+The last missing piece in automating the hyper-parameter optimization of our example model is to
+supply Oríon with the values to use for *epsilon*.
+
+We specify the search space in the command line using ``orion~loguniform(1e-5, 1.0)``
+as the argument for *espilon*. This argument will tell Oríon to use a log uniform distribution
+between ``1e-5`` and ``1`` for the values of *epsilon*.
+
+Putting everything together, we need to call ./main.py with Oríon. The syntax is the
```
bouthilx:
```suggestion
Putting everything together, we need to call ``main.py`` with Oríon. The syntax is the
```

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
...
+   -r requirements.txt``.
+
+Viewing the results
+-------------------
+Once the optimization reached its stopping condition, you can query Oríon to give you the results
+of the optimization with the sub-command ``$ orion info``:
+
+.. code-block:: bash
+
+   orion info -n scitkit-iris-tutorial
```
bouthilx: Maybe point to user/monitoring.html#library-api for examples on how to query the results from the DB?

docs/src/examples/scikit-learn.rst
```diff
@@ -0,0 +1,112 @@
+********************
+Scikit-learn
+********************
+.. Might also think of moving this file to examples/ then we an example auto contained in the
+   repository. We invoke this file from index.rst
+
+In this example, we're going to demonstrate how Oríon can be integrated to a minimal model using
+scikit-learn [scikit]_ on the iris dataset [iris]_. The files mentioned in this example are
```
bouthilx: I think it would be more useful here to provide a URL to scikit instead of a citation. For iris it would be great to provide an url to the dataset as well. Similarly, the line below should point to the examples folder in the repo on github, so the user can directly go see.

tests/functional/example/test_scikit_learn.py
```diff
@@ -0,0 +1,65 @@
...
+    assert len(trials) == 1
+
+    trial = trials[0]
+    assert trial.status == 'completed'
+    assert trial.params['/_pos_2'] == 0.1
+
+
+@pytest.mark.usefixtures("clean_db")
+@pytest.mark.usefixtures("null_db_instances")
+def test_script_results(monkeypatch):
```
bouthilx: Do we need this test are the one above is sufficient? I don't see the difference beside one looking at experiment.stats, but that could be done in the same test. :thinking:

docs/src/examples/scikit-learn.rst
Thomsch: Good catch! Thanks!

tests/functional/example/test_scikit_learn.py
Thomsch: I'm in favor of keeping them separate because they tests different things: the second one tests that the script can run to completion with Oríon. The third one tests that the script gives replicable results (which is important in the context of a tutorial). That way, if there is a replication issue, only the third test will fail instead of the one signaling that the script is incompatible with Oríon. 

However, we may improve the name of the test to reflect its purpose in case the docstring is not enough.

bouthilx: Sounds reasonable. I suggest you rename the last test `test_result_reproducibility`.

tests/functional/example/orion_config.yaml
Thomsch: Agreed, it removes one point of friction for the learner.

tests/functional/example/test_scikit_learn.py
Thomsch: Perfect!

docs/src/examples/scikit-learn.rst
Thomsch: Good idea!

docs/src/examples/scikit-learn.rst
Thomsch: Good idea!

examples/scikitlearn-iris/main.py
Thomsch: Thanks for the pointer

Thomsch:
> Well done! There is a couple of minor comments but overall the example is well written and very clear. Thank you! :)

Thank for the great feedback!

Thomsch:
I had to rebase the branch since the addition of `report_objective` is posterior to when I wrote the tutorial.

tests/functional/example/orion_config.yaml
Thomsch: Agreed, it removes friction points for the learner.

Thomsch: Actually, this test is not for the learner to see, it's for the CI system. All the other functional tests are using MongoDB.

bouthilx: Oh, I did not realize this was in tests/. Still, the tests should run on the PickledDB as we are migrating to it as the default backend. For the tutorial I leave it as is then, to avoid adding documentation of the DB configuration in it. Once we migrate to PickledDB the tutorial will implicitly as well.

docs/src/examples/scikit-learn.rst
Thomsch: Good idea!

tests/functional/example/orion_config.yaml
Thomsch: Agreed.

docs/src/examples/scikit-learn.rst
Thomsch: Good catch! I'm going to change it.

docs/src/examples/scikit-learn.rst
```diff
@@ -5,8 +5,8 @@ Scikit-learn
    repository. We invoke this file from index.rst
 
 In this example, we're going to demonstrate how Oríon can be integrated to a minimal model using
-scikit-learn [scikit]_ on the iris dataset [iris]_. The files mentioned in this example are
-available in the folder :file:`examples/scikitlearn-iris/`.
+`scikit-learn <https://scikit-learn.org/stable/index.html>`_ on the `iris dataset <https://en.wikipedia.org/wiki/Iris_flower_data_set>`_. The files mentioned in this example are
```
bouthilx: For iris I would give this [link](http://archive.ics.uci.edu/ml/datasets/Iris) instead so that user are just one click away from downloading it.

docs/src/examples/scikit-learn.rst
```diff
@@ -5,8 +5,8 @@ Scikit-learn
...
+available at https://github.com/Epistimio/orion/tree/master/examples/scikitlearn-iris/.
```
bouthilx: Maybe `examples/scikitlearn-iris/ <https://github.com/Epistimio/orion/tree/master/examples/scikitlearn-iris/>`_?

docs/src/examples/scikit-learn.rst
bouthilx: Not ``main.py``? :sweat_smile:

docs/src/examples/scikit-learn.rst
bouthilx: Instead of using the url you can reference it with `` :ref:`library-api-results` ``, the anchor already defined for the section here https://github.com/Epistimio/orion/blame/master/docs/src/user/monitoring.rst#L20

tests/functional/example/orion_config.yaml
bouthilx: So we leave it as mongodb?

Thomsch: Yes

docs/src/examples/scikit-learn.rst
Thomsch: I changed it!

docs/src/examples/scikit-learn.rst
Thomsch: That's better.

docs/src/examples/scikit-learn.rst
bouthilx: But it seams to be still ./main.py

tests/functional/example/orion_config.yaml
bouthilx: Why not using PickledDB if we plan to migrate to it by default?

Thomsch: Because we havn't migrated it yet and do not know when we'll do it. The card in Trello has been created last year in June and was switched back and forth from the backlog and current sprint in October.

Changing just this test to PickleDB worsen the cohesiveness of the codebase to gain a potential minor time later. However, since there is no scheduled for this migration, there is a serious chance that we will end up with a fragmented test suite where one part will use MongoDB and the other PickleDB, rising the chance of not covering the right features and making it at the end more complicated overall to migrate.

docs/src/examples/scikit-learn.rst
Thomsch: Changed it again...

tests/functional/example/orion_config.yaml
bouthilx: I agree about the fragmentation of the test suite, but a single migration would be overwhelming so we have no choice but to break it down and do it piece by piece. It has already began in many parts, so the fragmentation is already present. At least adding new tests with PickledDB would reduce the amount of remaining tests to migrate.

Thomsch: I wasn't aware of the existing fragmentation when I looked for tests to get inspiration from. Then, it's settled: we have a clear stance already defined. I'll migrate the tests to PickleDB.
Thanks for the precisions!

bouthilx: No problem, you raised good points!

Thomsch: Implemented in https://github.com/Epistimio/orion/pull/339/commits/25067a9295b287da22237d88752fe08ed7294d2a

tests/functional/example/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+# -*- coding: utf-8 -*-
+"""Common fixtures and utils for example tests."""
+
+import os
+import tempfile
+
+import pytest
+
+
+@pytest.fixture(scope="session")
```
bouthilx: scope='session' will likely cause issues. It would be preferable that the DB starts from scratch for each test.

tests/functional/example/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+def setup_database():
+    """Configure the database"""
+    os.environ['ORION_DB_TYPE'] = "pickleddb"
+    os.environ['ORION_DB_ADDRESS'] = tempfile.NamedTemporaryFile().name
```
bouthilx: Maybe we should handle  `tmp_file.close()` cleanly here. I don't know if using yield could work if the fixture is not used explicitly as an argument to the test function. You would need to verify.

tests/functional/example/test_scikit_learn.py
```diff
@@ -58,6 +52,7 @@ def test_orion_runs_script(monkeypatch):
 
 @pytest.mark.usefixtures("clean_db")
```
bouthilx: You can remove `null_db_instances` and `clean_db` now, we don't use the mongodb backend in these tests anymore. :tada: :)

tests/functional/example/conftest.py
Thomsch: I was wondering about that, but it seemed to work fine. I'll change it to `function` scope

tests/functional/example/test_scikit_learn.py
Thomsch: I'm always confused with the current database setup.

tests/functional/example/conftest.py
Thomsch: The file is deleted automatically at the end of the session though.
I confirm that yield works even when not used explicitly :+1: .

PR #341, opened 2020-02-26
Add support for non-standard -args

bouthilx:
Why:

Some user may call their script with unconventional formats such as
defining an argument with one dash but multiple characters (-arg instead
of --args) or using underscores in arguments (--a_fugly_argument).
Orion is meant to be non-intrusive and therefore should support these
use-cases as well instead of forcing the users to conform to standards.

How:

Save a mapping of the argument names to their actual argument format
(-arg or --arg or whatever) in a new attribute `parser.keys`. This
attribute is then reused to build the template which will be an exact
replicate of the provided commandline.

Backward compatibility is ensured by building `parser.keys` based on
previous way of inferring argument format (`_key_to_arg`).


bouthilx:
[Support -arg~prior()](https://trello.com/c/JbL1TnAS/426-support-argprior)

PR #342, opened 2020-02-26
Inconsistent output in orion status and orion list

donglinjy:
After run pytest, there are some invalid expriments and run orion status will get below error but orion list works well to return only the valid experiments, so I change this a little bit to make them consistent at this moment.

```
$ orion status
  File "/Users/donglin/App/anaconda2/envs/orion/bin/orion", line 11, in <module>
    load_entry_point('orion', 'console_scripts', 'orion')()
  File "/Users/donglin/gitspace/orion/src/orion/core/cli/__init__.py", line 39, in main
    return orion_parser.execute(argv)
  File "/Users/donglin/gitspace/orion/src/orion/core/cli/base.py", line 74, in execute
    function(args)
  File "/Users/donglin/gitspace/orion/src/orion/core/cli/status.py", line 56, in main
    experiments = get_experiments(args)
  File "/Users/donglin/gitspace/orion/src/orion/core/cli/status.py", line 113, in get_experiments
    for exp in experiments]
  File "/Users/donglin/gitspace/orion/src/orion/core/cli/status.py", line 113, in <listcomp>
    for exp in experiments]
  File "/Users/donglin/gitspace/orion/src/orion/core/io/experiment_builder.py", line 255, in build_view
    experiment = create_experiment(**db_config)
  File "/Users/donglin/gitspace/orion/src/orion/core/io/experiment_builder.py", line 289, in create_experiment
    experiment.algorithms = _instantiate_algo(experiment.space, kwargs.get('algorithms'))
  File "/Users/donglin/gitspace/orion/src/orion/core/io/experiment_builder.py", line 374, in _instantiate_algo
    return PrimaryAlgo(space, config)
  File "/Users/donglin/gitspace/orion/src/orion/core/worker/primary_algo.py", line 40, in __init__
    super(PrimaryAlgo, self).__init__(space, algorithm=algorithm_config)
  File "/Users/donglin/gitspace/orion/src/orion/algo/base.py", line 111, in __init__
    space, **subalgo_kwargs)
  File "/Users/donglin/gitspace/orion/src/orion/core/utils/__init__.py", line 159, in __call__
    raise NotImplementedError(error)
NotImplementedError: Could not find implementation of BaseAlgorithm, type = 'dumbalgo'
Currently, there is an implementation for types:
['asha', 'random']
```

There may be some other to-do related to this issue
1) Instead of print the whole error stack, may be can consider to only give the meaningful error message.
2) Not sure the purpose of 'clean_db' in our test set, but seems it won't clean but insert more.
```
@pytest.fixture()
def clean_db(database, exp_config):
    """Clean insert example experiment entries to collections."""
    database.experiments.drop()
    database.experiments.insert_many(exp_config[0])
    database.lying_trials.drop()
    database.trials.drop()
    database.trials.insert_many(exp_config[1])
    database.workers.drop()
    database.workers.insert_many(exp_config[2])
    database.resources.drop()
    database.resources.insert_many(exp_config[3])
```
@bouthilx


bouthilx:
> 1. Instead of print the whole error stack, may be can consider to only give the meaningful error message.

Agreed!

> 2. Not sure the purpose of 'clean_db' in our test set, but seems it won't clean but insert more.

The name may be misleading. The goal is to get db reset to a fresh initial state for the tests. (with some initial data)

src/orion/core/cli/status.py
```diff
@@ -104,13 +104,19 @@ def get_experiments(args):
...
 
+    if args['name']:
+        root_experiments = experiments
+    else:
+        root_experiments = [exp for exp in experiments
+                            if exp['refers'].get('root_id', exp['_id']) == exp['_id']]
+
     return [experiment_builder.build_view(name=exp['name'], version=exp.get('version', 1))
-            for exp in experiments]
+            for exp in root_experiments]
```
bouthilx: I believe the filter should be removed at line 71. https://github.com/Epistimio/orion/pull/342/files#diff-ae55d8dac36a389a72153331fa90394cR71

tests/functional/commands/test_status_command.py
```diff
@@ -93,6 +93,35 @@ def test_experiment_wout_success_wout_ac(clean_db, single_without_success, capsy
     assert captured == expected
 
 
+def test_experiment_number_same_list_status(clean_db, only_invalidalgo_experiments_db,
+                                            single_without_success, capsys):
+    """Test status and list command output the consistent number of experiments"""
```
bouthilx: Could we make separate tests for invalidalgo and coherence with list or both are related? I believe they are not.

tests/functional/commands/test_status_command.py
```diff
@@ -93,6 +93,35 @@ def test_experiment_wout_success_wout_ac(clean_db, single_without_success, capsy
...
+    orion.core.cli.main(['status'])
+
+    captured = capsys.readouterr().out
+
+    expected = """\
+test_single_exp-v1
```
bouthilx: Shouldn't the experiment with invalid algo be listed here as well? There is no `supernaedo2-dendi`. :thinking:

bouthilx:
[Uniformize format of status and list commands](https://trello.com/c/AERchE8b/431-uniformize-format-of-status-and-list-commands)

tests/functional/commands/test_status_command.py
donglinjy: sure, we can, I do it in one as we did not really fix the invalidalgo problem but just make their output consistent, so the invalidalgo and coherence get same output. But once we fix the invalidalgo experiment construct issue, they should be 2 cases.

tests/functional/commands/test_status_command.py
donglinjy: Since root_id and db _id is not the same for `supernaedo2-dendi`, so it is ignored. As mentioned above, once we get th invalid algo issue fixed, we need another case to cover this.

src/orion/core/cli/status.py
donglinjy: yeah, that filter is useless anymore since we filter the root experiments already. will change it.

tests/functional/commands/test_status_command.py
bouthilx: Oh, I didn't realize this. You should use `ensure_deterministic_id` to make sure it is valid.

I suggest you remove the invalid algo stuff and stash it for next PR. This specific test should not depend on it.

donglinjy: okay, let's add this later. I actually should not name it as invalid algorithm but just invalid experiment root_id which does not exist.

donglinjy: remove the invalid algo stuff and add it back when we get the safe experiment build work done.

PR #343, opened 2020-02-26
Add more helper functions for cmdline client

bouthilx:
Some users need a way to signal an interruption from within the script.
While KeyboardInterrupt in the shell where orion's worker is running
would catch the signal properly and interrupt the trial, raising
KeyboardInterrupt from within the script would only lead to a completed
script with error code > 0, hence the trial would be set to broken.

This commit adds a new helper function for this, `interrupt_trial`,
which terminates the process with a special code defined by Oríon, so
that the worker can detect that this is not a broken trial, but
an interrupted one.

Additional helper functions are added to simplify usage of Oríon:
`report_objective` and `report_bad_trials`.

### DIFF
--- a/docs/src/code/client.rst
+++ b/docs/src/code/client.rst
@@ -6,6 +6,7 @@ Client helper functions
    :maxdepth: 1
    :caption: Modules
 
+   client/cli
    client/experiment
    client/manual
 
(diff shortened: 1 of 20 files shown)


bouthilx:
[Add more helper functions for cmdline client](https://trello.com/c/gi6DAaf2/430-add-more-helper-functions-for-cmdline-client)

PR #345, opened 2020-02-27
Print important information at end of worker

bouthilx:
Why:

The important information such as `worker is stopping because experiment is broken`
was only printed in logger at INFO level. This should be printed at
*any* time.

How:

Turn the logger.info to a print and reuse `info` format for coherence.
Add also hints on other commands to run afterwards for monitoring.


bouthilx:
[Print important information at end of worker](https://trello.com/c/cnr2mspq/432-print-important-information-at-end-of-worker)

PR #346, opened 2020-02-29
Add comprehensive error message for branching err

bouthilx:
Why:

Users are confused when there is an error because of branching events.
Most of branching events are handled automatically, but when commandline
arguments are different for each workers or user scripts writes in code
repo there is a race condition between the version increments and it
fails.

How:

Write a clear message to explain what are the typical causes for these
issues and give hints on how to solve it.


bouthilx:
[Add comprehensive error message for branching err](https://trello.com/c/fiVYrdpv/436-add-comprehensive-error-message-for-branching-err)

PR #349, opened 2020-03-05
Make global, local and cmdline args coherent

bouthilx:
## Make global, local and cmdline args coherent
Why:

So far some arguments were supported either in global, local or cmdline,
sometimes in the same format, sometimes in different format. This
was horribly confusing for users.

How:

Force the local config to use the same format as global config by using
the latter as a reference point while parsing local config. The cmdline
arguments are processed and turned into a dictionary following the
structure of the global config.

## Add help and deprecation to Configuration

Why:

The help message for options is scattered and duplicated all over the code.
We should use the configuration as a common ground, where help messages
are globally defined. Also, there is many modifications coming to make
global, local and commandline options more coherent, which will require
many deprecation. Adding the deprecation inside the configuration makes
it possible to centralize all (most) deprecation warnings.

## Make configuration insensitive to - or _

Why:

It is very confusing for user if the configuration does recognize only
dash or underscore in option names.

How:

Convert to underscore any dash so that config is insensitive to both.
This means however that to_dict will return all keys with underscores.

### DIFF
--- a/docs/src/index.rst
+++ b/docs/src/index.rst
@@ -21,6 +21,7 @@
    user/algorithms
    user/script
    user/evc
+   user/config
 
 .. toctree::
    :caption: Examples
--- a/docs/src/user/config.rst
+++ b/docs/src/user/config.rst
@@ -1,5 +1,5 @@
 .. _configuration:
-    
+
 **********************
 Advanced Configuration
 **********************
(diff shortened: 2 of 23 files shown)


bouthilx:
[Uniformize support for configuration in orion.core/orion_config.yaml and local --config](https://trello.com/c/ZqJQhhGX/407-uniformize-support-for-configuration-in-orioncore-orionconfigyaml-and-local-config)

PR #352, opened 2020-03-07
Generic cardinality check in base algorithm

donglinjy:
Provide a default cardinality check in base algorithm. Any algorithm which does not want to use this will have to override is_done. 

@bouthilx

### DIFF
--- a/src/orion/algo/random.py
+++ b/src/orion/algo/random.py
@@ -8,8 +8,6 @@
    :synopsis: Draw and deliver samples from prior defined in problem's domain.
 
 """
-import hashlib
-
 import numpy
 
 from orion.algo.base import BaseAlgorithm
@@ -26,8 +24,6 @@ class Random(BaseAlgorithm):
         :param seed: Integer seed for the random number generator.
         """
         super(Random, self).__init__(space, seed=seed)
-        self.trial_info = {}  # Stores Unique Trial -> Result
-        self.cardinality = space.cardinality
 
     def seed_rng(self, seed):
         """Seed the state of the random number generator.
@@ -39,7 +35,7 @@ class Random(BaseAlgorithm):
     @property
     def state_dict(self):
         """Return a state dict that can be used to reset the state of the algorithm."""
-        return {'rng_state': self.rng.get_state(), 'trial_info': self.trial_info}
+        return {'rng_state': self.rng.get_state()}
 
     def set_state(self, state_dict):
         """Reset the state of the algorithm based on the given state_dict
@@ -48,7 +44,6 @@ class Random(BaseAlgorithm):
         """
         self.seed_rng(0)
         self.rng.set_state(state_dict['rng_state'])
-        self.trial_info = state_dict['trial_info']
 
     def suggest(self, num=1):
         """Suggest a `num` of new sets of parameters. Randomly draw samples
@@ -67,16 +62,4 @@ class Random(BaseAlgorithm):
 
         A simple random sampler though does not take anything into account.
         """
-        for point, result in zip(points, results):
-            _point = list(point)
-            _id = hashlib.md5(str(_point).encode('utf-8')).hexdigest()
-
-            if _id not in self.trial_info:
-                self.trial_info[_id] = result
-
-    @property
-    def is_done(self):
-        """Return True, if all possible sets of parameters has been tried."""
-        if len(self.trial_info) >= self.cardinality:
-            return True
-        return False
+        pass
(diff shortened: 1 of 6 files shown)


src/orion/core/worker/primary_algo.py
```diff
@@ -49,13 +48,18 @@ def seed_rng(self, seed):
     @property
     def state_dict(self):
         """Return a state dict that can be used to reset the state of the algorithm."""
-        return self.algorithm.state_dict
+        _state_dict = self.algorithm.state_dict
+        _state_dict.update(super(self.algorithm.__class__, self.algorithm).state_dict)
```
bouthilx: Why aren't you using `self.algorithm.state_dict` instead?

donglinjy: `self.algorithm.state_dict` only gets the child algorithm state but still need to sync _trials_info from base between copy algorithm and original algorithm. Although I did think the possibility to not sync the _trials_info, but it will change the original behavior.

src/orion/algo/random.py
```diff
@@ -39,7 +35,7 @@ def seed_rng(self, seed):
     @property
     def state_dict(self):
         """Return a state dict that can be used to reset the state of the algorithm."""
-        return {'rng_state': self.rng.get_state(), 'trial_info': self.trial_info}
+        return {'rng_state': self.rng.get_state()}
```
bouthilx: If we remove abstractmethod from `observe` to make its definition optional in children class, we could remove it altogether in `Random` and reuse super().state_dict/set_state to insert `trial_info` in the states. This would solve the issue for the call on `super` in PrimaryAlgo.

donglinjy: Even we remove abstractmethod of observe, but not sure how many actual algorithm will stop implement their own. This will remove the super call in PrimaryAlgo.observe if the child does not need their own observe. As for state_dict and set_state, are you saying we ask user to explicitly in their concrete child algorithm? 

```
def state_dict(self):
      child_state_dict = xxx
      child_state_dict.update(super.state_dict)
def set_state(self, state_dict):
      super.set_state(state_dict)
```

And even for observe, even the child algorithm get its own algorithm, we also ask them to call super.observe like this?
```
def observe(self, points, results):
       super.observe(points, results)
      # child observer code here ...
```

This way it indeed simplify the code in PrimaryAlgo, but actually ask sub-algorithm to call super explicitly, I guess people could make mistake even we can document about this.

bouthilx: Yes I meant the latter. It may well be an common mistake to forget it, but if we include the call to super in the cookie-cutter template with a clear documentation of this call then the mistake should happen rarely actually.

src/orion/core/worker/primary_algo.py
bouthilx: I'll mark this one as resolved as it's the same issue as discussed here: https://github.com/Epistimio/orion/pull/352#discussion_r389329855

src/orion/algo/random.py
donglinjy: okay, using the template could be a way, normally I would say split such a feature to depend part on orion own code and part on user's code could be risky, but let's get it work first.

src/orion/core/worker/primary_algo.py
```diff
@@ -51,6 +50,8 @@ def state_dict(self):
         """Return a state dict that can be used to reset the state of the algorithm."""
         return self.algorithm.state_dict
 
+    # pylint:disable=bad-super-call
+    # Bad first argument 'NoneType' given to super() (bad-super-call)
```
bouthilx: This can be removed now.

src/orion/core/worker/primary_algo.py
```diff
@@ -80,6 +81,8 @@ def suggest(self, num=1):
 
         return [self.transformed_space.reverse(point) for point in points]
 
+    # pylint:disable=bad-super-call
+    # Bad first argument 'NoneType' given to super() (bad-super-call)
```
bouthilx: Idem

bouthilx:
Alright! Could you please add tests for the `trials_info`. I will update the template accordingly.

src/orion/core/worker/primary_algo.py
donglinjy: good catch

donglinjy:
> Alright! Could you please add tests for the `trials_info`. I will update the template accordingly.

Yeah, sure. Was thinking the Random case already cover this some level. Let me add some ut case for base algorithm

PR #354, opened 2020-03-10
Hyperband oob

donglinjy:
Move hyperband from a separate algorithms plugin [orion.core.hyeprband](https://github.com/Epistimio/orion.algo.hyperband) to an OOB algorithms in Orion core.

### DIFF
--- a/src/orion/algo/hyperband.py
+++ b/src/orion/algo/hyperband.py
@@ -83,7 +83,7 @@ class Hyperband(BaseAlgorithm):
 
     def __init__(self, space, seed=None, repeat=numpy.inf):
         self.brackets = []
-        super(Hyperband, self).__init__(space, seed=seed)
+        super(Hyperband, self).__init__(space, seed=seed, repeat=numpy.inf)
 
         self.trial_info = {}  # Stores Trial -> Bracket
 
--- a/src/orion/algo/hyperband.py
+++ b/src/orion/algo/hyperband.py
@@ -130,7 +130,7 @@ class Hyperband(BaseAlgorithm):
             raise RuntimeError(
                 'Hyperband keeps sampling already existing points. This should not happen, '
                 'please report this error to '
-                'https://github.com/bouthilx/orion.algo.hyperband/issues')
+                'https://github.com/Epistimio/orion/issues')
 
         return points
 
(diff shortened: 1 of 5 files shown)


donglinjy:
@bouthilx

src/orion/algo/hyperband.py
```diff
@@ -0,0 +1,423 @@
...
+                point = list(point)
+                point[self.fidelity_index] = bracket.rungs[0]['resources']
+                points.append(tuple(point))
+            i += 1
+
+        if not points:
+            raise RuntimeError(
+                'Hyperband keeps sampling already existing points. This should not happen, '
+                'please report this error to '
+                'https://github.com/bouthilx/orion.algo.hyperband/issues')
```
bouthilx:
```suggestion
                'https://github.com/Epistimio/orion/issues')
```

tests/functional/algos/test_algos.py
```diff
@@ -12,6 +12,7 @@
 
 config_files = ['random_config.yaml']
 fidelity_config_files = ['random_config.yaml', 'asha_config.yaml']
+hyperband_config = ['hyperband.yaml']
```
bouthilx: Can you add the yaml file to the list above instead so that we don't need to add a test explicitly?

bouthilx:
Documentation will also need to be updated: https://orion.readthedocs.io/en/latest/user/algorithms.html#included-algorithms

bouthilx:
[Add Hyperband to Oríon](https://trello.com/c/fwjQ4gqG/453-add-hyperband-to-or%C3%ADon)

src/orion/algo/hyperband.py
donglinjy: > Documentation will also need to be updated: https://orion.readthedocs.io/en/latest/user/algorithms.html#included-algorithms
> 
> (here https://github.com/Epistimio/orion/blob/master/docs/src/user/algorithms.rst)

Will add doc for hyperband later when we get some other issues done which may affect how to write the doc.

donglinjy:
> Documentation will also need to be updated: https://orion.readthedocs.io/en/latest/user/algorithms.html#included-algorithms
> 
> (here https://github.com/Epistimio/orion/blob/master/docs/src/user/algorithms.rst)

Will add doc for hyperband later when we get some other issues done which may affect how to write the doc.

PR #360, opened 2020-03-18
Add max_trials to algorithm.is_done

bouthilx:
Why:

We need to be able to discern when experiment is completed and no more
trials should be executed, and when algorithm is completed and no more
trials should be created. With current experiment.is_done, we sample
more than `max_trials`. Adding `max_trials` to termination criterion of
algo makes it possible for the algo to flag when it should not sample
new points, but still experiment.is_done is False because there is
pending trials.

How:

Add max_trials to algo in `Producer`. This is somewhat hacky but
`max_trials` is tightly integrated to `Experiment` so moving it to
algorithms would be complicated. Maybe something we should consider
doing later on.

In experiment.is_done, it now return True if number of completed trials
is greater than or equal to `max_trials`, or if `algorithm.is_done` and
there is no more pending trials.

### DIFF
--- a/src/orion/algo/base.py
+++ b/src/orion/algo/base.py
@@ -205,6 +205,10 @@ class BaseAlgorithm(object, metaclass=ABCMeta):
         """
         if len(self._trials_info) >= self.space.cardinality:
             return True
+
+        if len(self._trials_info) >= getattr(self, 'max_trials', float('inf')):
+            return True
+
         return False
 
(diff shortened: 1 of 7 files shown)


bouthilx:
[Stop when number of all trials matches `max_trials` instead of number of completed trials](https://trello.com/c/e2ahZH2q/462-stop-when-number-of-all-trials-matches-maxtrials-instead-of-number-of-completed-trials)

PR #361, opened 2020-03-18
Avoid duplicates in RandomSearch

bouthilx:
Why:

The filtering of duplicates in RandomSearch is a lot more efficient than
relying on the database with DuplicateKeyErrors. There will still be
race conditions leading to this error, but filtering upfront in
RandomSearch reduces significantly the number of registration failures
and lead to better workload balance between the workers. Without this
fix, some workers end up always updating their algo and trying to sample
the next point but loose all race conditions.

How:

Use _trial_info to filter out duplicates while sampling points...

### DIFF
--- a/src/orion/algo/base.py
+++ b/src/orion/algo/base.py
@@ -18,6 +18,11 @@ from orion.core.utils import Factory
 log = logging.getLogger(__name__)
 
 
+def infer_trial_id(point):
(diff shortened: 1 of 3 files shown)


bouthilx:
[Better balance workload between workers](https://trello.com/c/0xfKEFfT/463-better-balance-workload-between-workers)

PR #362, opened 2020-03-18
Handle timeout in DBs (PickledDB in particular)

bouthilx:
Why:

When there are timeouts, the errors are not handled and
python leaves with scary stack-traces in terminal for users.
There is typical issues causing timeouts in PickledDB, the error message
should explain these to the users so they learn what to do to fix it.

Note:

They timeouts never occurred for MongoDB during the stress test and it
is not clear what message should be provided to users, given there is no
obvious root problems as of now. So I leave this as empty reraise, and
do not test...

### DIFF
--- a/src/orion/core/io/database/__init__.py
+++ b/src/orion/core/io/database/__init__.py
@@ -305,6 +305,12 @@ class DuplicateKeyError(DatabaseError):
     pass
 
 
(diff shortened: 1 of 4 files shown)


bouthilx:
[Handle database timeouts](https://trello.com/c/Jq5lKsMo/465-handle-database-timeouts)

PR #363, opened 2020-03-18
hyperband fixes

donglinjy:
This PR include below fixes for Hyerband:
- Compute budgets does not output the same table as the original paper
**Fix**: Besides the fix, also add a debug log to show the budget table like this:
image

- Duplicate points in a single hyperband execution will cause some points been tried but won't be able to register into the correct budget since we choose bracket one time based on fidelity. 
**Fix**: Won't allow repeated points in a single hyperband execution. And if the search space is too small to fill the first rungs of all brackets, hyperband won't be able to promote and experiment only run the points of the first rungs and exit.

- Support to run duplicate points in different hyperband execution only if the later one get bigger fidelity compared with previous execution.
 
- When multiple workers run in parallel, there are 2 issues:
   - Some workers won't be able to exit normally when some trials is running and naive algorithm is done but source algorithm is not.
   **Fix**: Sync the executed number of times from naive algorithm to source.
   -  When hyberband.observe with points belong to different hyperband executions, points could be registered to the bracket of a wrong hyperband executions.
   **Fix**: Try to refresh brackets of hyperband after observe of each point.

- Add documentation for Hyperband

@bouthilx

### DIFF
--- a/docs/src/code/algo/hyperband.rst
+++ b/docs/src/code/algo/hyperband.rst
@@ -1,5 +1,5 @@
 Hyperband Algorithm
-=======================
+===================
 
 .. automodule:: orion.algo.hyperband
    :members:
--- a/docs/src/code/algo/hyperband.rst
+++ b/docs/src/code/algo/hyperband.rst
@@ -1,0 +1,5 @@
+Hyperband Algorithm
+=======================
+
+.. automodule:: orion.algo.hyperband
+   :members:
--- a/docs/src/user/algorithms.rst
+++ b/docs/src/user/algorithms.rst
@@ -63,7 +63,7 @@ Seed for the random number generator used to sample new trials. Default is ``Non
 Hyperband
 ---------
 
-`Hyperband`_ extends the `SuccessiveHalving`_ algorithm by providing an way to exploit a
+`Hyperband`_ extends the `SuccessiveHalving`_ algorithm by providing a way to exploit a
 fixed budget with different number of configurations for ``SuccessiveHalving`` algorithm to
 evaluate. Each run of ``SuccessiveHalving`` will be defined as a ``bracket`` in Hyperband.
 Hyperband requires two inputs (1) ``R``, the maximum amount of resource that can be allocated
@@ -95,18 +95,20 @@ Configuration
 .. code-block:: yaml
 
     algorithms:
-       asha:
+       hyperband:
           seed: null
-          frequency: 1
+          repetitions: 1
 
 
 ``seed``
 
 Seed for the random number generator used to sample new trials. Default is ``None``.
 
-``frequency``
+``repetitions``
 
-Number of execution for Hyperband. Default is ``numpy.inf`` which means to run Hyperband
+Number of executions for Hyperband. A single execution of Hyperband takes a finite
+budget of ``(log(R)/log(eta) + 1) * (log(R)/log(eta) + 1) * R``, and ``repetitions`` allows you
+to run multiple executions of Hyperband. Default is ``numpy.inf`` which means to run Hyperband
 until no new trials can be suggested.
 
 
--- a/docs/src/user/algorithms.rst
+++ b/docs/src/user/algorithms.rst
@@ -58,6 +58,58 @@ Configuration
 
 Seed for the random number generator used to sample new trials. Default is ``None``.
 
+.. _hyperband-algorithm:
+
+Hyperband
+---------
+
+`Hyperband`_ extends the `SuccessiveHalving`_ algorithm by providing an way to exploit a
+fixed budget with different number of configurations for ``SuccessiveHalving`` algorithm to
+evaluate. Each run of ``SuccessiveHalving`` will be defined as a ``bracket`` in Hyperband.
+Hyperband requires two inputs (1) ``R``, the maximum amount of resource that can be allocated
+to a single configuration, and (2) ``eta``, an input that controls the proportion of
+configurations discarded in each round of SuccessiveHalving.
+
+To use Hyperband in Oríon, you must specify one parameter with ``fidelity(low, high, base)``
+as the prior, ``low`` will be ignored, ``high`` will be taken as the maximum resource ``R``
+and ``base`` will be taken as the reduction factor ``eta``.
+
+Number of epochs usually can be used as the resource but the algorithm is generic and can be
+applied to any multi-fidelity setting. That is, you can use training time, specifying the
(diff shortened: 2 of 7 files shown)


bouthilx:
Great! :smiley: I will look at the PR in details tonight or tomorrow.

bouthilx:
[Fixes for Hyperband](https://trello.com/c/N1HSbZJ5/469-fixes-for-hyperband)

docs/src/user/algorithms.rst
```diff
@@ -58,6 +58,58 @@ Configuration
 
 Seed for the random number generator used to sample new trials. Default is ``None``.
 
+.. _hyperband-algorithm:
+
+Hyperband
+---------
+
+`Hyperband`_ extends the `SuccessiveHalving`_ algorithm by providing an way to exploit a
```
bouthilx:
```suggestion
`Hyperband`_ extends the `SuccessiveHalving`_ algorithm by providing a way to exploit a
```

docs/src/user/algorithms.rst
```diff
@@ -58,6 +58,58 @@ Configuration
...
+
+   Current implementation does not support more than one fidelity dimension.
+
+Configuration
+~~~~~~~~~~~~~
+
+.. code-block:: yaml
+
+    algorithms:
+       asha:
```
bouthilx: Should be adapted to Hyperband :wink:

docs/src/user/algorithms.rst
```diff
@@ -58,6 +58,58 @@ Configuration
...
+
+
+``seed``
+
+Seed for the random number generator used to sample new trials. Default is ``None``.
+
+``frequency``
+
+Number of execution for Hyperband. Default is ``numpy.inf`` which means to run Hyperband
+until no new trials can be suggested.
```
bouthilx: Maybe this should give more information on what it means to run Hyperband many times. The user may be confused: how can it continue well all trials of all brackets are completed?

src/orion/algo/hyperband.py
```diff
@@ -1,13 +1,12 @@
 # -*- coding: utf-8 -*-
 """
-:mod:`orion.algo.hyperband.hyperband -- TODO
-=================================================
+:mod:`orion.algo.hyperband` -- A Novel Bandit-Based Approach to Hyperparameter Optimization
+============================================================================================
```
bouthilx:
```suggestion
===========================================================================================
```

docs/src/code/algo/hyperband.rst
```diff
@@ -0,0 +1,5 @@
+Hyperband Algorithm
+=======================
```
bouthilx:
```suggestion
===================
```

src/orion/algo/hyperband.py
```diff
@@ -18,17 +17,15 @@
...
 
-
 REGISTRATION_ERROR = """
 Bad fidelity level {fidelity}. Should be in {budgets}.
 Params: {params}
 """
 
 SPACE_ERROR = """
-Hyperband cannot be used if space does contain a fidelity dimension.
+Hyperband cannot be used if space does not contain a fidelity dimension.
```
bouthilx: Good catch! :joy:

src/orion/algo/hyperband.py
```diff
@@ -202,13 +247,63 @@ def suggest(self, num=1):
...
+            brackets = [bracket for bracket in self.brackets
+                        if bracket.rungs[0]['resources'] == fidelity]
+
+        if not brackets:
+            raise ValueError("No bracket found for point {0} with fidelity {1}"
+                             .format(_id_wo_fidelity, fidelity))
 
-        return hashlib.md5(str(non_fidelity_dims).encode('utf-8')).hexdigest()
+        if len(brackets) > 1:
+            logger.warning("More than one bracket found for point %s, this should not happen",
```
bouthilx: This could happen now with repetitions as described here  https://github.com/Epistimio/orion/pull/363/files#diff-fd0f6183ef73c491312c6e13328f8353R169, right?

src/orion/algo/hyperband.py
```diff
@@ -300,13 +380,24 @@ def __init__(self, hyperband, budgets):
 
     @property
     def is_filled(self):
-        """Return True if last rung with trials is filled"""
+        """Return True if first rung with trials is filled"""
         return self.has_rung_filled(0)
 
+    def get_point_max_resource(self, point):
+        """Return the max resource value been tried for a point"""
```
bouthilx:
```suggestion
        """Return the max resource value that has been tried for a point"""
```

src/orion/algo/hyperband.py
```diff
@@ -78,14 +108,19 @@ class Hyperband(BaseAlgorithm):
     seed: None, int or sequence of int
         Seed for the random number generator used to sample new trials.
         Default: ``None``
+    frequency: int
```
bouthilx: I believe `repetitions` would be more accurate.

docs/src/user/algorithms.rst
donglinjy: good catch, I copied from asha. 😉

src/orion/algo/hyperband.py
donglinjy: I think it should not, if there are more than bracket found for a point, means there are duplicate points in the same hyeprband execution, which is not allowed. We only allow duplicate in different hyperband execution, which should not cause more brackets found a point as the self.brackets here only have bracket for the current hyperband exeucton.

src/orion/algo/hyperband.py
donglinjy: You always have the good name :)

docs/src/user/algorithms.rst
donglinjy: okay, let me explain more about this.

donglinjy:
Seems only osx test failed @bouthilx

src/orion/algo/hyperband.py
bouthilx: Oh! I didn't realize that self.brackets was replaced here: https://github.com/Epistimio/orion/pull/363/files#diff-fd0f6183ef73c491312c6e13328f8353R269. I thought you were appending them. Gotcha, thanks for the explanation!

donglinjy:
There seems a potential bug of space transformer when the lower/upper bound value suggested from the original space.

bouthilx:
Indeed. I saw this issue in the stress test I am implementing. I should fix it this week.

PR #366, opened 2020-03-24
Fix broken-link in monitoring

Thomsch:
Why:
There is a broken link in the monitoring section when referencing the EVC. https://trello.com/c/HWp2LH8J

How:
I replaced the broken external link with a valid internal link. This way, if the linked document is moved, the build system will throw an error.

### DIFF
--- a/docs/src/user/library/evc_results.rst
+++ b/docs/src/user/library/evc_results.rst
@@ -1,7 +1,7 @@
 Iterative Results with EVC
 --------------------------
 
-When using the experiment version control (described :doc:`here <../../user/evc>`),
+When using the experiment version control (described :doc:`here </user/evc>`),
 the experiments are connected in a tree structure which we call the EVC tree.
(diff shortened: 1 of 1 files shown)


docs/src/user/library/evc_results.rst
```diff
@@ -1,7 +1,7 @@
 Iterative Results with EVC
 --------------------------
 
-When using the experiment version control (described `here <user/evc>`_),
+When using the experiment version control (described :doc:`here <../../user/evc>`),
```
bouthilx: Thanks for the fix! Can you try with `/user/evc` instead?

Thomsch: Great idea! I remember trying it but it didn't work at the time. In hindsight, it was probably a small syntax error. I tried again and it worked. It's much better this way 🙏.

bouthilx:
Closing and re-opening because for some reason github didn't receive travis notice that tests passed.

PR #369, opened 2020-03-30
Baseline method for shrinking search space

aimar1986bupt:
I added the baseline method of shrinking search space.

### DIFF
--- a/src/orion/core/cli/__init__.py
+++ b/src/orion/core/cli/__init__.py
@@ -20,7 +20,8 @@ log = logging.getLogger(__name__)
 
 def load_modules_parser(orion_parser):
     """Search through the `cli` folder for any module containing a `get_parser` function"""
-    modules = module_import.load_modules_in_path('orion.core.cli', lambda m: hasattr(m, 'add_subparser'))
+    modules = module_import.load_modules_in_path('orion.core.cli',
+                                                 lambda m: hasattr(m, 'add_subparser'))
     # modules = module_import.load_modules_in_path('orion.core.cli')
 
     for module in modules:
@@ -35,9 +36,7 @@ def main(argv=None):
 
     orion_parser = OrionArgsParser()
 
-
     load_modules_parser(orion_parser)
-
 
     return orion_parser.execute(argv)
 
--- a/src/orion/core/cli/__init__.py
+++ b/src/orion/core/cli/__init__.py
@@ -14,8 +14,6 @@ import logging
 from orion.core.cli.base import OrionArgsParser
 from orion.core.utils import module_import
 
-# from cli.base import OrionArgsParser
-# from ..utils import module_import
 
 log = logging.getLogger(__name__)
 
@@ -37,20 +35,9 @@ def main(argv=None):
 
     orion_parser = OrionArgsParser()
 
-    # orion_parser.parser.add_argument(
-    #     '-v', '--verbose',
-    #      default=0,)
 
     load_modules_parser(orion_parser)
 
-    # argv="-v hunt -n orion-tutorial --max-trials 5 ./main.py --lr~'loguniform(1e-5, 1.0)'"
-
-    # argv = ["-v", "hunt", "-n", "orion-tutorial2", "--max-trials", "2",
-    #         "/home/xuechao/orion_ibm/examples/mnist/main.py", "--lr~loguniform(1e-5, 1.0)", "--gamma~uniform(0.2, 0.7)"]
-
-    argv = ["-v", "hunt", "-n", "orion-tutorial9", "--max-trials", "2",
-            "/home/xuechao/orion_ibm/examples/mnist/main.py",
-            "--lr~loguniform(5e-3, 5e-2)", "--gamma~uniform(0.2, 0.7)"]
 
     return orion_parser.execute(argv)
 
--- a/src/orion/core/cli/__init__.py
+++ b/src/orion/core/cli/__init__.py
@@ -14,13 +14,16 @@ import logging
 from orion.core.cli.base import OrionArgsParser
 from orion.core.utils import module_import
 
+# from cli.base import OrionArgsParser
+# from ..utils import module_import
+
 log = logging.getLogger(__name__)
 
 
 def load_modules_parser(orion_parser):
     """Search through the `cli` folder for any module containing a `get_parser` function"""
-    modules = module_import.load_modules_in_path('orion.core.cli',
-                                                 lambda m: hasattr(m, 'add_subparser'))
+    modules = module_import.load_modules_in_path('orion.core.cli', lambda m: hasattr(m, 'add_subparser'))
+    # modules = module_import.load_modules_in_path('orion.core.cli')
 
     for module in modules:
         get_parser = getattr(module, 'add_subparser')
@@ -34,7 +37,20 @@ def main(argv=None):
 
     orion_parser = OrionArgsParser()
 
+    # orion_parser.parser.add_argument(
+    #     '-v', '--verbose',
+    #      default=0,)
+
     load_modules_parser(orion_parser)
+
+    # argv="-v hunt -n orion-tutorial --max-trials 5 ./main.py --lr~'loguniform(1e-5, 1.0)'"
+
+    # argv = ["-v", "hunt", "-n", "orion-tutorial2", "--max-trials", "2",
+    #         "/home/xuechao/orion_ibm/examples/mnist/main.py", "--lr~loguniform(1e-5, 1.0)", "--gamma~uniform(0.2, 0.7)"]
+
+    argv = ["-v", "hunt", "-n", "orion-tutorial9", "--max-trials", "2",
+            "/home/xuechao/orion_ibm/examples/mnist/main.py",
+            "--lr~loguniform(5e-3, 5e-2)", "--gamma~uniform(0.2, 0.7)"]
 
     return orion_parser.execute(argv)
 
--- a/src/orion/core/cli/base.py
+++ b/src/orion/core/cli/base.py
@@ -16,9 +16,6 @@ import orion
 from orion.core.io.database import DatabaseError
 from orion.core.utils.exceptions import NoConfigurationError
 
-# from ..io.database import DatabaseError
-# from ..utils.exceptions import NoConfigurationError
-
 
 CLI_DOC_HEADER = """
 orion:
--- a/src/orion/core/cli/base.py
+++ b/src/orion/core/cli/base.py
@@ -16,6 +16,9 @@ import orion
 from orion.core.io.database import DatabaseError
 from orion.core.utils.exceptions import NoConfigurationError
 
+# from ..io.database import DatabaseError
+# from ..utils.exceptions import NoConfigurationError
+
 
 CLI_DOC_HEADER = """
 orion:
--- a/src/orion/core/cli/hunt.py
+++ b/src/orion/core/cli/hunt.py
@@ -72,11 +72,5 @@ def main(args):
     # TODO: simplify when parameter parsing is refactored
     experiment = experiment_builder.build_from_args(args)
     config = experiment_builder.get_cmd_config(args)
-
-    # print("iamhereeeeee!!!!!!!!!", config)
-
     worker_trials = config.get('worker_trials', orion.core.config.experiment.worker_trials)
-
-    transfer_search_space(experiment)
-
     workon(experiment, worker_trials)
--- a/src/orion/core/cli/hunt.py
+++ b/src/orion/core/cli/hunt.py
@@ -72,5 +72,11 @@ def main(args):
     # TODO: simplify when parameter parsing is refactored
     experiment = experiment_builder.build_from_args(args)
     config = experiment_builder.get_cmd_config(args)
+
+    # print("iamhereeeeee!!!!!!!!!", config)
+
     worker_trials = config.get('worker_trials', orion.core.config.experiment.worker_trials)
+
+    transfer_search_space(experiment)
+
     workon(experiment, worker_trials)
--- a/src/orion/core/transfer/learn_search_space/__init__.py
+++ b/src/orion/core/transfer/learn_search_space/__init__.py
@@ -8,16 +8,12 @@
    :synopsis: Executes transfer optimization of learning search space.
 
 """
-
 import logging
+
 import numpy as np
 
-# from orion.core.worker.consumer import Consumer
-# from orion.core.worker.producer import Producer
-# from orion.storage.base import get_storage
 
 log = logging.getLogger(__name__)
-
 # It should apply experiment and trials objects (in its same root path?)
 # to represent current task (experiment) and historical tasks (trials)
 # and update the space domain in the experiment object
--- a/src/orion/core/transfer/learn_search_space/__init__.py
+++ b/src/orion/core/transfer/learn_search_space/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-:mod:`orion.core.transfer` -- Coordination of the transfer optimization procedure of learning search space
+:mod:`orion.core.transfer` -- Coordination of transfer optimization of learning search space
 ======================================================================
 
 .. module:: transfer
--- a/src/orion/core/transfer/learn_search_space/__init__.py
+++ b/src/orion/core/transfer/learn_search_space/__init__.py
@@ -8,10 +8,7 @@
    :synopsis: Executes transfer optimization of learning search space.
 
 """
-import io
-import itertools
 import logging
-import pprint
 import numpy as np
 
 # from orion.core.worker.consumer import Consumer
@@ -44,7 +41,7 @@ log = logging.getLogger(__name__)
 
 
 def rep_task(hist_tasks, curr_task):
-    """represent search space as matrix"""
+    """Represent search space as matrix"""
     hpp_idx_map = {}
     # idx_hpp_map = {}
     curr_bound = np.zeros((2, len(curr_task["space"])))  # 0: lower, 1: upper
@@ -101,7 +98,7 @@ def shrink_ellips_space(hist_tasks, curr_task):
 
 
 def rej_sampling_ellips_space(matrix_a, vector_b, curr_task):
-    """reject sample from ellipsoidal search space example"""
+    """Reject sample from ellipsoidal search space example"""
     hyper_param_dim = vector_b.shape[0]
     _, curr_bound = rep_task([], curr_task)
 
@@ -122,14 +119,15 @@ def rej_sampling_ellips_space(matrix_a, vector_b, curr_task):
     # x is in the same order as curr_task representation
     return x
 
+
 if __name__ == "__main__":
     # curr_task = {"space":
     #                   [{"lr": [0.001, 0.04]},
     #                    {"wd": [0.00001, 0.00005]}]}
 
     curr_task = {"space":
-                     [{"lr": [0.001, 0.04]},
-                      {"wd": [0.00001, 0.01]}]}
+                 [{"lr": [0.001, 0.04]},
+                  {"wd": [0.00001, 0.01]}]}
 
     hist_task = [{"params":
                   [{"name": "lr", "type": "real", "value": 0.05},
@@ -138,10 +136,8 @@ if __name__ == "__main__":
                   [{"name": "lr", "type": "real", "value": 0.02},
                    {"name": "wd", "type": "real", "value": 0.0003}]},
                  {"params":
-                   [{"name": "lr", "type": "real", "value": 0.007},
-                    {"name": "wd", "type": "real", "value": 0.0002}]}]
+                  [{"name": "lr", "type": "real", "value": 0.007},
+                   {"name": "wd", "type": "real", "value": 0.0002}]}]
 
     shrink_bbox_space(hist_task, curr_task)
-
     print("update curr task:", curr_task)
-
--- a/src/orion/core/transfer/learn_search_space/__init__.py
+++ b/src/orion/core/transfer/learn_search_space/__init__.py
@@ -8,9 +8,9 @@
    :synopsis: Executes transfer optimization of learning search space.
 
 """
+
 import logging
 import numpy as np
-
 
 # from orion.core.worker.consumer import Consumer
 # from orion.core.worker.producer import Producer
--- a/src/orion/core/transfer/learn_search_space/__init__.py
+++ b/src/orion/core/transfer/learn_search_space/__init__.py
@@ -1,0 +1,147 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.core.transfer` -- Coordination of the transfer optimization procedure of learning search space
+======================================================================
+
+.. module:: transfer
+   :platform: Unix
+   :synopsis: Executes transfer optimization of learning search space.
+
+"""
+import io
+import itertools
+import logging
+import pprint
+import numpy as np
+
+# from orion.core.worker.consumer import Consumer
+# from orion.core.worker.producer import Producer
+# from orion.storage.base import get_storage
+
+log = logging.getLogger(__name__)
+
+# It should apply experiment and trials objects (in its same root path?)
+# to represent current task (experiment) and historical tasks (trials)
+# and update the space domain in the experiment object
+# The data structure is aligned with the object of experiments and trials
+# curr_task = {"space":
+#                   [{"lr": [lower, upper]},
+#                    ...,
+#                    {"wd": [lower, upper]}]}
+
+# hist_task = [{"params":
+#               [{"name": "lr", "type": real, "value": 0.05},
+#                ...,
+#                {"name": "wd", "type": real, "value": 0.0001}]},
+#              ...,
+#              {"params":
+#               [{"name": "lr", "type": real, "value": 0.02},
+#                ...,
+#                {"name": "wd", "type": real, "value": 0.0003}]}]
+
+# constraints: 1) hyper-params are the same in curr and hist;
+#              2) all types are real
+
+
+def rep_task(hist_tasks, curr_task):
+    """represent search space as matrix"""
+    hpp_idx_map = {}
+    # idx_hpp_map = {}
+    curr_bound = np.zeros((2, len(curr_task["space"])))  # 0: lower, 1: upper
+    hist_opt = np.zeros((len(hist_tasks), len(curr_task["space"])))
+
+    for i in range(len(curr_task["space"])):
+        value_item = next(iter(curr_task["space"][i].values()))
+        key_item = next(iter(curr_task["space"][i].keys()))
+        hpp_idx_map[key_item] = i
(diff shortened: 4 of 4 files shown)


bouthilx:
[Transfer learning - Space definition](https://trello.com/c/YmRuKpAk/494-transfer-learning-space-definition)

Thomsch:
Thank you for your contribution! We're currently reviewing it.

I changed the title to be a little bit more expressive. Feel free to change it!
If you can add a short description to this pull request it would be great 🥇.

Thomsch:
Forcing Travis from .org to .com

bouthilx:
I think this could be a great example to add [here](https://github.com/Epistimio/orion/tree/develop/examples) and in the corresponding example section in the documentation. We could provide an example of how to re-use existing experiment in Oríon with this method to generate a new search space. I don't see however how this could fit in the core given how we define experiments, that's why I would propose to turn it into an example.

There is an issue however with the sampling technique. `rej_sampling_ellips_space` could be set as a condition on the search space but unfortunately we don't support conditional search space yet. Would you want to work on the integration of conditional dimensions? From our offline discussions I feel you would prefer working on the integration of the evolutionary algorithm. If that's the case, I suggest we close this PR and open an issue about Conditional Dimensions in which we point to this PR so that when we are able to, we revive this PR and integrate it. What would you think about this?

aimar1986bupt:
@bouthilx I agree with you. We can put it as an example to generate a new search space. And close this PR. Open issue about conditional dimensions.

aimar1986bupt:
I move search space shrinking to examples and add an evolution algorithm.

Thomsch:
[Take a decision about PR#369](https://trello.com/c/XT87PzX0/571-take-a-decision-about-pr369)

aimar1986bupt:
Customized mutate function interface added.

src/orion/algo/mutate_functions.py
```diff
@@ -0,0 +1,72 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def default_mutate(self, winner_id, loser_id, multiply_factor, add_factor):
+    """Get a default mutate function"""
+    select_genes_key_list = np.random.choice(self.search_space_remove_fidelity,
+                                             self.eves.nums_mutate_gene,
+                                             replace=False)
+    self.copy_winner(winner_id, loser_id)
```
bouthilx, line 24: I think we should extract the copying logic out of the mutate function. It is too specific to the EvolutionaryES algorithm.

src/orion/algo/evolution_es.py
```diff
@@ -152,10 +151,22 @@ def __init__(self, evolutiones, budgets, repetition_id, space):
         super(BracketEVES, self).__init__(evolutiones, budgets, repetition_id)
         self.eves = self.hyperband
         self.space = space
-        self._search_space_remove_fidelity = []
+        self.search_space_remove_fidelity = []
+
+        if evolutiones.mutate_attr:
```
bouthilx, line 156: Minor syntax comment, `evolutiones` should be `evolution_es`.

aimar1986bupt: I have fixed them. Please check.

aimar1986bupt:
I have fixed the mutate function and the syntax problem. Please check.

src/orion/algo/evolution_es.py
```diff
@@ -243,9 +243,19 @@ def get_candidates(self, rung_id):
...
-                         self.mutate_attr["add_factor"])
+        select_genes_key_list = np.random.choice(self.search_space_remove_fidelity,
+                                                 self.eves.nums_mutate_gene,
+                                                 replace=False)
+        self.copy_winner(winner_id, loser_id)
+        for i, _ in enumerate(select_genes_key_list):
+            space = self.space.values()[select_genes_key_list[i]]
+            old = self.eves.population[select_genes_key_list[i]][loser_id]
+
+            new = self.mutate_func(self.mutate_attr["multiply_factor"],
```
bouthilx, line 254: I believe the mutate_attr should be passed as `**kwargs` arguments to support any type of function and arguments. I can understand why space and old arguments would be required for any function. Is the volatility more a parameter of the general evolutionary algorithm or more a parameter of the mutation function?

aimar1986bupt: Good suggestion. I have done it. Please check it. Volatility is more a mutation parameter, so I  view it as other mutate_attr.

src/orion/algo/evolution_es.py
```diff
@@ -247,14 +245,18 @@ def _mutate(self, winner_id, loser_id):
                                                  self.eves.nums_mutate_gene,
                                                  replace=False)
         self.copy_winner(winner_id, loser_id)
+        kwarg = {}
+        if self.mutate_attr.get("multiply_factor") and not kwarg.get("multiply_factor"):
+            kwarg["multiply_factor"] = self.mutate_attr["multiply_factor"]
+        if self.mutate_attr.get("add_factor") and not kwarg.get("add_factor"):
+            kwarg["add_factor"] = self.mutate_attr["add_factor"]
```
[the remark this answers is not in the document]
aimar1986bupt: Done!

examples/learn_search_space/__init__.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.core.transfer` -- Coordination of transfer optimization of learning search space
+======================================================================
+
+.. module:: transfer
+   :platform: Unix
+   :synopsis: Executes transfer optimization of learning search space.
+
```
bouthilx, line 9: There should be a citation to the paper here.

src/orion/algo/mutate_functions.py
```diff
@@ -0,0 +1,61 @@
...
+            new_value = old_value * factors
+        elif lower_bound > old_value * factors:
+            new_value = lower_bound + volatility * np.random.random()
+        else:
+            new_value = upper_bound - volatility * np.random.random()
+    elif search_space.type == "integer":
+        if search_space.prior.name == "uniform" or \
+           search_space.prior.name == "loguniform":
+            lower_bound = search_space.prior.a
+            upper_bound = search_space.prior.b
```
bouthilx, line 44: We should raise an error if the prior is not supported.

src/orion/algo/mutate_functions.py
```diff
@@ -0,0 +1,61 @@
...
+def default_mutate(search_space, old_value, **kwargs):
+    """Get a default mutate function"""
+    multiply_factor = kwargs.pop('multiply_factor', 3.0)
+    add_factor = kwargs.pop('add_factor', 1)
+    volatility = kwargs.pop('volatility', 0.001)
+    lower_bound = -np.inf
+    upper_bound = np.inf
+    if search_space.type == "real":
+        if search_space.prior.name == "uniform" or \
+           search_space.prior.name == "loguniform":
```
bouthilx, line 28: We should raise an error if the prior is not supported.

src/orion/algo/space.py
```diff
@@ -913,7 +918,7 @@ def items(self):
         """Return items sorted according to keys"""
         return [(k, self[k]) for k in self.keys()]
 
-    def values(self):
+    def values(self) -> object:
```
bouthilx, line 921: There is no typing in Oríon so far.

src/orion/core/cli/__init__.py
```diff
@@ -21,6 +21,7 @@ def load_modules_parser(orion_parser):
     """Search through the `cli` folder for any module containing a `get_parser` function"""
     modules = module_import.load_modules_in_path('orion.core.cli',
                                                  lambda m: hasattr(m, 'add_subparser'))
+    # modules = module_import.load_modules_in_path('orion.core.cli')
```
bouthilx, line 24: This should be removed.

src/orion/core/worker/consumer.py
```diff
@@ -234,6 +234,8 @@ def execute_process(self, cmd_args, environ):
         command = cmd_args
 
         signal.signal(signal.SIGTERM, _handler)
+        # print("command:", command)
+        # print("environ:", environ)
```
bouthilx, line 238: Should be removed also. :)

src/orion/core/worker/producer.py
```diff
@@ -132,7 +132,7 @@ def register_trials(self, new_point):
             return 1
 
         except DuplicateKeyError:
-            log.debug("#### Duplicate sample.")
+            log.info("#### Duplicate sample.")
```
bouthilx, line 135: I'm not sure this should be info. What do you think?

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,97 @@
...
+        bracket.eves = evolution
+        bracket.rungs = [rung_0, rung_1]
+        point = (1, 0.0)
+        point_hash = hashlib.md5(str([0.0]).encode('utf-8')).hexdigest()
+
+        evolution.observe([point], [{'objective': 0.0}])
+
+        assert len(bracket.rungs[0])
+        assert point_hash in bracket.rungs[0]['results']
+        assert (0.0, point) == bracket.rungs[0]['results'][point_hash]
```
bouthilx, line 97: The default mutate function should be tested for different type of priors.

src/orion/algo/evolution_es.py
```diff
@@ -0,0 +1,263 @@
...
+                            if len(list(rung.values())) > self.eves.nums_population
+                            else len(list(rung.values())))
+
+        for i in range(population_range):
+            for j in self.search_space_remove_fidelity:
+                self.eves.population[j][i] = list(rung.values())[i][1][j]
+            self.eves.performance[i] = list(rung.values())[i][0]
+
+        population_index = list(range(self.eves.nums_population))
+        red_team = np.random.choice(population_index, self.eves.nums_comp_pairs, replace=False)
```
bouthilx, line 188: All sampling should be seeded to ensure reproducibility. The RNG of self.eves could be used for this.

src/orion/algo/evolution_es.py
```diff
@@ -0,0 +1,263 @@
...
+        diff_list = list(set(population_index).difference(set(red_team)))
+        blue_team = np.random.choice(diff_list, self.eves.nums_comp_pairs, replace=False)
+
+        winner_list = []
+        loser_list = []
+
+        hurdles = 0
+        for i, _ in enumerate(red_team):
+            winner = (red_team
```
bouthilx, line 197: Suggestion
```winner, loser = (red_team, blue_team) if ... else (blue_team, red_team)```

File: src/orion/algo/evolution_es.py [hunk 15]
```diff
@@ -0,0 +1,263 @@
...
+            point = [0] * len(self.space)
+            nums_all_equal = 0
+            while True:
+                point[self.eves.fidelity_index] = \
+                    list(rung.values())[i][1][self.eves.fidelity_index]
+
+                for j in self.search_space_remove_fidelity:
+                    if self.space.values()[j].type == "integer" or \
+                       self.space.values()[j].type == "categorical":
+                        point[j] = int(self.eves.population[j][i])
```
bouthilx, line 225: Why is this casting necessary, shouldn't it be part of the mutate function?

src/orion/algo/evolution_es.py
```diff
@@ -0,0 +1,263 @@
...
+                      if self.eves.performance[red_team[i]] < self.eves.performance[blue_team[i]]
+                      else blue_team)
+            loser = (red_team
+                     if self.eves.performance[red_team[i]] >= self.eves.performance[blue_team[i]]
+                     else blue_team)
+
+            winner_list.append(winner[i])
+            loser_list.append(loser[i])
+            hurdles += self.eves.performance[winner[i]]
+            self._mutate(winner[i], loser[i])
```
bouthilx, line 207: It seems to me the second loop could be avoided by verifying here if the mutation created a duplicate in the population and then mutating again and again until it is not creating a duplicate.

Also, you could use `self.eves.get_id` (see [here]( https://github.com/Epistimio/orion/blob/develop/src/orion/algo/hyperband.py#L254)) to build a set of ids to test for duplicates.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_compute_budgets():
+    """Verify proper computation of budgets on a logarithmic scale"""
+    # Check typical values
+    assert compute_budgets(1, 3, 1, 2, 1) == [[(2, 1), (2, 2), (2, 3)]]
+    assert compute_budgets(1, 4, 2, 4, 2) == [[(4, 1), (4, 2), (4, 4)]]
+
+
+class TestEvolutionES():
```
bouthilx, line 81: It should be tested with another dummy mutate function to make sure they are supported.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,97 @@
...
+    """Tests for the algo Hyperband."""
+
+    def test_register(self, evolution, bracket, rung_0, rung_1):
```
bouthilx, line 84: The new get_candidates is fairly complex. It should have it's dedicated unit-test. For instance we should test whether there is indeed no duplicates, and if it behaves correctly when it's not possible to avoid duplicates. The current way of continuing and logging a warning leaves me dubious. In the case of Hyperband for instance, we log the warning but then None is returned instead of a point, leading to the termination of the worker.

examples/learn_search_space/__init__.py
aimar1986bupt: fixed.

src/orion/algo/mutate_functions.py
aimar1986bupt: In the default mutate function, the prior only effects on lower/upper bound (-inf, inf by default), rather than sampling. The mutation function (*3 or +1) does not depend on the prior. On the other hand, the sampling method (sample from the prior) is inherited from Hyperband, so the verifying is there.

src/orion/algo/mutate_functions.py
aimar1986bupt: see above

src/orion/algo/space.py
aimar1986bupt: fixed

src/orion/core/cli/__init__.py
aimar1986bupt: fixed.

src/orion/core/worker/consumer.py
aimar1986bupt: Fixed.

src/orion/core/worker/producer.py
aimar1986bupt: Fixed.

src/orion/algo/evolution_es.py
aimar1986bupt: Fixed.

src/orion/algo/evolution_es.py
aimar1986bupt: Good suggestion. Fixed.

src/orion/algo/evolution_es.py
aimar1986bupt: Fixed.

src/orion/algo/evolution_es.py
aimar1986bupt: The second loop can be moved to 207, but can not be avoided. Because the we need to extract self.eves.population and assign it to points. The extraction process need one loop. get_id() need point object as its parameter rather than population, so we still need to extract population to point firstly. I tried to do that following by your suggestion, but it can not save much space or branches. I didn't make change here, because the flow: rungs-->population-->points is clear.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Added.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Added.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Added. Also see above. The mutate function does not depend on priors, which does not sampling from prior.

src/orion/algo/mutate_functions.py
bouthilx: I see, so if the search space prior is not uniform or loguniform the bounds will be (-inf, inf) and it will work. This could be replaced by `lower_bound, upper_bound = search_space.interval()` to support priors more generally. Some will indeed lead to (-inf, inf).

src/orion/algo/space.py
```diff
@@ -745,8 +750,8 @@ def __init__(self, name, low, high, base=2):
             raise AttributeError("Minimum resources must be a positive number.")
         elif low > high:
             raise AttributeError("Minimum resources must be smaller than maximum resources.")
-        if base <= 1:
-            raise AttributeError("Base should be greater than 1")
+        if base < 1:
+            raise AttributeError("Base should be great than or equal to 1")
```
bouthilx, line 754:
```suggestion
            raise AttributeError("Base should be greater than or equal to 1")
```

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def test_compute_budgets():
+    """Verify proper computation of budgets on a logarithmic scale"""
+    # Check typical values
+    assert compute_budgets(1, 3, 1, 2, 1) == [[(2, 1), (2, 2), (2, 3)]]
+    assert compute_budgets(1, 4, 2, 4, 2) == [[(4, 1), (4, 2), (4, 4)]]
+
+
+def test_get_mutated_candidates(space1):
```
bouthilx, line 91: The test should call the method get_candidate directly. Copying the code here makes the maintenance difficult and likely to lead to missed errors as we try to harmonize the code in the test and the one in get_candidate. I understand this method is a bit difficult to test as it does several things. I think extracting the different parts of get_candidate into private methods would make it easier to test each private method separately.

1. [The first part of the function](https://github.com/Epistimio/orion/pull/369/files#diff-8d7c467e92a292572c23c35ff41bc207R181-R191) focus on the creation of the population set, with the red and blue teams. This could be moved into a single function which would return all the necessary objects. 
2. After that there is the [loop for mutations](https://github.com/Epistimio/orion/pull/369/files#diff-8d7c467e92a292572c23c35ff41bc207R197-R206). This could be another private method which would be easier to test. I think the verification for duplicates should be done in 2, but if it is not possible it should be done before the creation of the list of points so that we can test the two processes separately. That would be a third private method, and would make the next one the fourth private method.
3. Finally there is the [creation of the list of points](https://github.com/Epistimio/orion/pull/369/files#diff-8d7c467e92a292572c23c35ff41bc207R213-R235).

aimar1986bupt: Done. I split the get_candidates function into three functions, and test them directly in the test cases.

src/orion/algo/evolution_es.py
```diff
@@ -0,0 +1,272 @@
...
+        Seed for the random number generator used to sample new trials.
+        Default: ``None``
+    repetitions: int
+        Number of execution of Hyperband. Default is numpy.inf which means to
+        run Hyperband until no new trials can be suggested.
+
+    """
+
+    def __init__(self, space, seed=None, repetitions=np.inf, nums_population=20, mutate=None):
+        super(EvolutionES, self).__init__(space, seed=seed, repetitions=repetitions)
```
bouthilx, line 102: I think we should verify that nums_population is even otherwise we will have unbalanced teams.

src/orion/algo/evolution_es.py
```diff
@@ -190,6 +190,10 @@ def get_candidates(self, rung_id):
         blue_team = self.eves.rng.choice(diff_list, self.eves.nums_comp_pairs,
                                          replace=False)
 
+        return rung, population_range, red_team, blue_team
+
+    def _get_mutated_population(self, red_team, blue_team):
```
bouthilx, line 195:
```suggestion
    def _mutate_population(self, red_team, blue_team):
```
Since it is modifying the population in-place.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -237,3 +182,83 @@ def test_register(self, evolution, bracket, rung_0, rung_1):
...
+
+        org_data = copy.deepcopy(org_data)
+
+        bracket._get_mutated_population(red_team, blue_team)
+
+        mutated_data = np.stack((list(bracket.eves.population.values())[0],
+                                 list(bracket.eves.population.values())[1]), axis=0).T
+
+        assert org_data.shape == mutated_data.shape
+        assert (mutated_data[0] == org_data[0]).all()
```
bouthilx, line 220: Could you please document the rational here for the tests. It's hard to infer quickly. Thanks!

src/orion/algo/evolution_es.py
```diff
@@ -222,12 +228,12 @@ def get_candidates(self, rung_id):
                     point[j] = self.eves.population[j][i]
 
                 if tuple(point) in points:
-                    nums_all_equal += 1
+                    nums_all_equal[i] += 1
                     logger.debug("find equal one, continue to mutate.")
                     self._mutate(points.index(tuple(point)), i)
```
bouthilx, line 233: I would be in favor of moving these mutations to avoid duplicates to `_mutate_population` so that roles of `_mutate_population` and `_get_mutated_points` are clearly separated. If I understand correctly this function would be only returning the population in the format of points. So I would suggest to rename it `_get_population_points`. (Assuming the mutation to avoid duplications are moved to `_mutate_population`). This would also avoid side-effects in this function, concentrating all in-place operations in `_mutate_population`.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -237,3 +182,83 @@ def test_register(self, evolution, bracket, rung_0, rung_1):
...
+
+    def test_duplicated_mutated_population(self, bracket, rung_4):
+        """Verify duplicated candidates can be found and processed correctly."""
+        population_range = 4
+        for i in range(4):
+            for j in [1, 2]:
+                bracket.eves.population[j][i] = list(rung_4["results"].values())[i][1][j]
+        points, nums_all_equal = bracket._get_mutated_points(rung_4["results"], population_range)
+
+        if points[1][1] != points[2][1]:
```
bouthilx, line 245: Could you please document the rational here for the tests as well. Thanks!

tests/unittests/algo/test_evolution_es.py
bouthilx: The customized function should be tested when passed to EvolutionES to make sure the whole pipeline supports it properly. You can move the custom function to `src/orion/core/utils/tests.py` if you need to have it in a proper python package to be able to define the string of the function to import.

src/orion/algo/evolution_es.py
aimar1986bupt: I considered this case. In the code:  pair = nums_population // 2, which ensure the teams are balanced. User doesn't need to care about the nums_population is even or odd, if it is odd, for example 9, then the blue team and red team are both 4 individuals. It is transparent to the user.

src/orion/algo/evolution_es.py
aimar1986bupt: Done.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Done. For my test case: Winner team will be [0, 2], so [0, 2] will be remained, while [1, 3] will be mutated. In addition, for each individual, mutation occurs in only one dimension chosen from two. So for example, if lr is mutated, weight decay will be copied from individual 0/2 directly, and vice versa. I added this comments in code.

src/orion/algo/evolution_es.py
aimar1986bupt: In my opinion, there is no need to add a function to merely implement "convert population to points" (there may be only one sentence to do that.) So I move the duplications into _mutate_population() as you suggestion, and implement the "convert population to points" in passing.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Done. In this case, duplication will occur, and we can make it mutate one more time. The points 1 and 2 should be different, while one of nums_all_equal should be 1 (meaning it mutate one more time). I added this comments in code.

tests/unittests/algo/test_evolution_es.py
aimar1986bupt: Good suggestions. Done.

src/orion/algo/evolution_es.py
bouthilx: I would be in favor of keeping them separated to simplify testing, but let's keep them together now that tests are ok.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,274 @@
...
+
+def test_compute_budgets():
+    """Verify proper computation of budgets on a logarithmic scale"""
+    # Check typical values
+    assert compute_budgets(1, 3, 1, 2, 1) == [[(2, 1), (2, 2), (2, 3)]]
+    assert compute_budgets(1, 4, 2, 4, 2) == [[(4, 1), (4, 2), (4, 4)]]
+
+
+def test_customized_mutate_func(space2):
+    """Verify customized mutate function works correctly"""
```
bouthilx, line 127: This test should be modified to use EvolutionES directly. If we modify the code in EvolutionES, in particular the code for the support of custom functions, we will need to synchronize any changes in the test here. This will likely lead to missed bugs. We should test the existing code in EvolutionES and avoid testing a duplication of its code otherwise maintainance and testing become hazardous.

aimar1986bupt: Good suggestion. Fixed.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -0,0 +1,284 @@
...
+        assert (nums_all_equal == 0).all()
+
+    def test_customized_mutate_population(self, bracket, rung_3):
+        """Verify mutated candidates is generated correctly."""
+        red_team = [0, 2]
+        blue_team = [1, 3]
+        population_range = 4
+
+        mutate_attr = {}
+        function_string = mutate_attr.pop('function',
```
bouthilx, line 243: What I meant was to replace this part with 
```python
algo = EvolutionES(space, mutate="orion.core.utils.tests.customized_mutate_example")
```
And then use `algo.brackets[0]._mutate_population` to test the custom function. This way we test the code within `EvolutionES` to convert the string to a function and the test won't need to be adapted to reflect changes in EvolutionES builder if we make any.

aimar1986bupt: I see. Done.

bouthilx:
Alright! :) If you can update your branch on current develop I will merge the PR. Thanks!

aimar1986bupt:
@bouthilx This one 9be29f7 is the latest one after I merged the current dev. Could you have a look and check it?

PR #371, opened 2020-03-31
Release & Packaging

Thomsch:
This pull request introduces a dedicated page for the documentation regarding the release process and the distribution of new releases.

### DIFF
--- a/docs/src/developer/release.rst
+++ b/docs/src/developer/release.rst
@@ -38,6 +38,13 @@ time to create the release artifacts and publish the release.
 
 Publishing the release
 ======================
+Once the release is correctly documented and integrated to the VCS workflow, we can publish it to
+the public.
+
+* Publish the GitHub release. The source code archives will be added automatically by GitHub to the
+  release.
+* Publish the new version to PyPI_ by executing ``$ tox -e release`` from the tagged commit on the
+  master branch.
 
 After the release
 =================
--- a/docs/src/developer/release.rst
+++ b/docs/src/developer/release.rst
@@ -48,3 +48,15 @@ the public.
 
 After the release
 =================
+Once published, it's important to notify our user base and community that a new version exists so
+they can update their version or learn about Oríon.
+
+* Verify Oríon's Zenodo_ page has been updated to reflect the new release on GitHub_. Zenodo is
+  configured to automatically create a new version whenever a new release is published on GitHub.
+* Announce the new release on your #orion's slack channel.
+* Announce the new release on relevant communication channels (e.g., email, forums, google groups)
+* Celebrate! You published a new version of Oríon! Congratulations!
+
+.. _GitHub: https://github.com/Epistimio/orion/releases
+.. _Zenodo: https://doi.org/10.5281/zenodo.3478592
+.. _PyPI: https://pypi.org/project/orion/
--- a/docs/src/developer/release.rst
+++ b/docs/src/developer/release.rst
@@ -21,8 +21,20 @@ the README.md.
 #. Run the stress tests according to the instruct-codingions in stress test's documentation.
 #. Update the **Citation** section in the project's README.md with the latest version of Oríon.
 
+.. _release-make:
+
 Making the release
 ==================
+Once the release is throughly tested and the core contributors are confident in the release, it's
+time to create the release artifacts and publish the release.
+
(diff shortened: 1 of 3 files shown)


docs/src/developer/release.rst
```diff
@@ -0,0 +1,62 @@
...
+============================
+The first step in releasing a new version is to create a release candidate. A release candidate
+allows us to throughly test the new version and iron out the remaining bugs. Additionally, it's also
+at this time that we make sure to change the version number and update related documentation such as
+the README.md.
+
+#. Create a new branch from the *develop* branch named ``release-{version}rc``, where
+   ``{version}`` is replaced by the number of the new version (e.g., ``1.2.0``). This effectively
+   freezes the feature set for this new version, while allowing regular development to continue take
+   place in the *develop* branch. More information is available in :ref:`standard-vcs`.``{version}
```
bouthilx:
```suggestion
   place in the *develop* branch. More information is available in :ref:`standard-vcs`.``{version}``
```

bouthilx: Well, why is ``{version}`` there? Is it a mistake?

docs/src/developer/release.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Publishing the release
+======================
+Once the release is correctly documented and integrated to the VCS workflow, we can publish it to
+the public.
+
+* Publish the GitHub release. The source code archives will be added automatically by GitHub to the
+  release.
+* Publish the new version to PyPI_ by executing ``$ tox -e release`` from the tagged commit on the
+  master branch.
```
bouthilx: You can add another note that the package will be automatically pushed to Anaconda we the test of master branch first pass. We should probably automate PyPI the same way.

docs/src/developer/release.rst
```diff
@@ -0,0 +1,62 @@
...
+After the release
+=================
+Once published, it's important to notify our user base and community that a new version exists so
+they can update their version or learn about Oríon.
+
+* Verify Oríon's Zenodo_ page has been updated to reflect the new release on GitHub_. Zenodo is
+  configured to automatically create a new version whenever a new release is published on GitHub.
+* Announce the new release on your #orion's slack channel.
+* Announce the new release on relevant communication channels (e.g., email, forums, google groups)
+* Celebrate! You published a new version of Oríon! Congratulations!
```
bouthilx: Should we celebrate or should we make it more like a common thing to do. Like in the DevOps-way, release often, any-time, and make it a simple and reliable process.

docs/src/developer/release.rst
Thomsch: Mmmmmh, I'm not sure how this guy arrived here. Good catch.

docs/src/developer/release.rst
Thomsch: I didn't see that travis was uploading the conda binaries. Thanks for pointing it out.

Indeed, we should have both upload using the same conduit. Either both tox or both travis (or both github, it's possible too now via GitHub Actions).

bouthilx: I would opt to include in tox and use the tox command in travis. @Delaunay Played with GitHub Actions a bit. If could could centralize CI on github that would be great.

docs/src/developer/release.rst
Thomsch: That's a good point which I thought about while writing this bit. Initially, I wanted to shape it like "Congratulations! You published a new version of Orion! It's no small task" but I removed it for the exact reason. What I tried to emphasize are positivity and encouragement. If it's too much, maybe we can remove "Congratulations!"? What is your opinion?

bouthilx: To me it's more the `Celebrate` part that implies that it's an important achievement (more than it should be IMHO). Personally that is the part I would remove but maybe I am too boring. :upside_down_face:

Thomsch: Haha, absolutely not. I wanted to put "profit" instead initially as a reference for fun but decided against it ultimately.
I'm going to remove "Celebrate!", it's a good suggestion :)

bouthilx: Haha, ok!

docs/src/developer/release.rst
Thomsch: I added the cards on Trello

PR #373, opened 2020-04-01
Making upper bound inclusive

bouthilx:
Why:

The priors typically have exclusive upper bound but due to numerical
errors algorithms are likely to suggest points on the upper bound.


bouthilx:
[Fix out-of-space issue](https://trello.com/c/WYLirJEi/473-fix-out-of-space-issue)

codecov.yml
```diff
@@ -40,7 +40,8 @@ coverage:
       default:
         branches: null
 
-  ignore: null
+  ignore:
+    - tests/*
```
Thomsch, line 44: This configuration change is outside the scope of this bugfix. Can I suggest, we add those changes in a separate PR?

src/orion/algo/space.py
```diff
@@ -312,44 +310,32 @@ def _is_numeric(item):
...
-    high : float
-       Constrain with an upper bound (exclusive), default ``numpy.inf``.
+       Lower bound (inclusive), optional; default ``-numpy.inf``.
+    high : float:
+       Upper bound (inclusive), optional; default ``numpy.inf``.
+    precision : int
+        Precision, optional; default ``4``.
+    shape : tuple
+       Defines how many dimensions are packed in this `Dimension`.
+       Describes the shape of the corresponding tensor.
```
Thomsch, line 335: Since the expected behavior is that the higher bound is exclusive, maybe we could add a quick rationale for why it's not the case.

src/orion/algo/space.py
```diff
@@ -475,7 +461,7 @@ def interval(self, alpha=1.0):
             int_high = int(numpy.floor(high))
         except OverflowError:  # infinity cannot be converted to Python int type
             int_high = numpy.inf
-        if int_high < high:  # Exclusive upper bound
+        if int_high < high:  # Inclusive upper bound
```
Thomsch, line 464: Why is it not `<=` ?

src/orion/algo/space.py
```diff
@@ -484,15 +470,28 @@ def __contains__(self, point):
 
 
 class Integer(Real, _Discrete):
-    """Subclass of `Dimension` for representing integer parameters.
+    """Search space dimension that can take on any real value.
```
Thomsch, line 473: > any real value.
What is the difference compared to the base class `Real`?

codecov.yml
bouthilx: Fair enough. I created #377 and will now remove it from this PR.

src/orion/algo/space.py
bouthilx: Because we would overflow to next integer when `int_high == high`. That should be removed actually and use `ceil` instead of `floor` for `int_high`.

src/orion/algo/space.py
bouthilx: Oopsie. Good catch, that was wrong copy paste from Real class.

src/orion/algo/space.py
bouthilx: Do you mean in the docstring for the parameter `high` or as a note?

codecov.yml
Thomsch: Thanks!

src/orion/algo/space.py
Thomsch: I was thinking to include it in the class's docstring.

src/orion/algo/space.py
Thomsch: Thanks for the explanations!

PR #376, opened 2020-04-02
Remove useless deprecation warnings

bouthilx:
Why:

There was still access to deprecated options in the `core` which were
triggering deprecation warnings event though the user was not using
any of these.

How:

Adjust access to config to use the proper configuration values
and update yaml reading to transfer deprecated settings to proper ones
with warning for users.


bouthilx:
[Remove deprecation warning that pops in create_experiment even if not using these options](https://trello.com/c/uMI2qWpR/454-remove-deprecation-warning-that-pops-in-createexperiment-even-if-not-using-these-options)

bouthilx:
Sure :)

PR #377, opened 2020-04-02
Ignore test files for code coverage

bouthilx:
Why:

We are not planning to tests the tests, hence it does not make sense to
measure the coverage of the tests on the test code. I would be in favor
of ignoring these files as this often leads to misleading coverage
changes in PRs if they slightly increase code coverage but add
significantly more test code.


bouthilx:
Note: I believe codecov reports are not merged properly. I'm gonna investigate and push a fix here if necessary.

bouthilx:
[Ignore test files for code coverage](https://trello.com/c/FwCZtoss/498-ignore-test-files-for-code-coverage)

Thomsch:
Excellent idea!!!

PR #378, opened 2020-04-02
Fix typos in developer documentation

Thomsch:
I discovered some typos that were not picked up by VSCode which I use to write the doc. 

I installed a plugin for spellchecking to avoid this situation from happening again.


Thomsch:
Can't believe I missed them too. Maybe we should do something about that. Is there a way to check spelling via tox?

bouthilx:
That would be amazing! If we can find something like this that would be great for a git commit hook together with black.

PR #381, opened 2020-04-11
Add TPE algorithm into Orion

donglinjy:
`Tree-structured Parzen Estimator (TPE)` algorithm is one of Sequential Model-Based
Global Optimization (SMBO) algorithms, which will build models to propose new points based
on the historical observed trials.

@bouthilx


docs/src/user/algorithms.rst
```diff
@@ -186,6 +186,71 @@ converging trials that do not lead to best results at convergence (stragglers).
...
+
+`Tree-structured Parzen Estimator`_ (TPE) algorithm is one of Sequential Model-Based
+Global Optimization (SMBO) algorithms, which will build models to propose new points based
+on the historical observed trials.
+
+Instead of modeling p(y|x) like other SMBO algorithms, TPE models p(x|y) and p(y),
+and p(x|y) is modeled by transforming that generative process, replacing the distributions of
+the configuration prior with non-parametric densities.
+
+The TPE defines p(x|y) using two such densities l(x) and g(x) while l(x) is distribution of
```
bouthilx, line 203:
```suggestion
The TPE defines p(x|y) using two such densities l(x) and g(x) where l(x) is distribution of
```

docs/src/user/algorithms.rst
```diff
@@ -186,6 +186,71 @@ converging trials that do not lead to best results at convergence (stragglers).
...
+good points and g(x) is the distribution of bad points. New point candidates will be sampled
```
bouthilx, line 204: This sentence or a next sentence should state how good and bad points are defined by a quantile `\gamma` so that a point for which y < y* is a good one and y >= y* is a bad one, where y* is defined based on \gamma = p(y < y*).

bouthilx: Maybe: Good and bad points are defined as the `gamma` best points and `1 - gamma` worst points observed so far.

docs/src/user/algorithms.rst
```diff
@@ -186,6 +186,71 @@ converging trials that do not lead to best results at convergence (stragglers).
...
+
+Ratio to split the observed trials into good and bad distributions. Default is ``0.25``.
+
+``equal_weight``
+
+True to set equal weights for observed points. Default is ``False``.
+
+``seed``
+
+The weight given to the prior point of the input space. Default is ``1.0``.
```
bouthilx, line 250: Sounds like this should have a different name and seed description should go below? :thinking:

bouthilx: Oh, right, prior_weight based on configuration example above.

src/orion/algo/space.py
```diff
@@ -533,6 +538,11 @@ def get_prior_string(self):
         prior_string = super(Integer, self).get_prior_string()
         return prior_string[:-1] + ', discrete=True)'
 
+    @property
+    def prior_name(self):
+        """Return the name of the prior"""
+        return "int_uniform"
```
bouthilx, line 544: The discrete dimension can have different priors. It should be `'int_{}'.format(super(Integer, self).prior_name)`

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+            point = []
+            below_points, above_points = self.split_trials()
+            below_points = numpy.array([flatten_dims(point, self.space) for point in below_points])
+            above_points = numpy.array([flatten_dims(point, self.space) for point in above_points])
+            idx = 0
+            for dimension in self.space.values():
+
+                if dimension.type == 'real':
+                    shape = dimension.shape
+                    if not shape:
```
bouthilx, line 225: Maybe we could keep this for when we support shape > 1.

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+                else:
+                    raise ValueError("TPE now only support Real Dimension.")
+
+                idx += 1
+            point = regroup_dims(point, self.space)
+            samples.append(point)
+        return samples
+
+    def _sample_real(self, dimension, below_points, above_points):
+        """Return a real point based on the observed good and bad points"""
```
bouthilx, line 244:
```suggestion
        """Sample a real dimension value based on the observed good and bad points"""
```

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+        low, high = dimension.interval()
+        below_mus, below_sigmas, blow_weights = \
```
bouthilx, line 246:
```suggestion
        below_mus, below_sigmas, below_weights = \
```

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+            adaptive_parzen_estimator(below_points, low, high, self.prior_weight,
+                                      self.equal_weight, flat_num=25)
+        above_mus, above_sigmas, above_weights = \
+            adaptive_parzen_estimator(above_points, low, high, self.prior_weight,
+                                      self.equal_weight, flat_num=25)
+
+        gmm_sampler_below = GMMSampler(self, below_mus, below_sigmas, low, high, blow_weights)
```
bouthilx, line 253:
```suggestion
        gmm_sampler_below = GMMSampler(self, below_mus, below_sigmas, low, high, below_weights)
```

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+    def observe(self, points, results):
+        """Observe evaluation `results` corresponding to list of `points` in
+        space.
+
+        A simple random sampler though does not take anything into account.
+        """
+        super(TPE, self).observe(points, results)
+
+
+class GMMSampler():
```
bouthilx, line 285: I think it would be better to use [sklearn's Gaussian Mixture](https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html#sklearn.mixture.GaussianMixture). This way we could also easily swap it to kernel [density estimator (parzen window)](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KernelDensity.html#sklearn.neighbors.KernelDensity).

tests/functional/algos/test_algos.py
```diff
@@ -71,9 +71,9 @@ def test_simple(monkeypatch, config_file):
 
 @pytest.mark.usefixtures("clean_db")
 @pytest.mark.usefixtures("null_db_instances")
-@pytest.mark.parametrize('config_file', config_files)
-def test_random_stop(monkeypatch, config_file):
+def test_random_stop(monkeypatch):
     """Test a simple usage scenario."""
+    config_file = 'random_config.yaml'
```
bouthilx, line 76: Why did you remove the parametrization? This means only random search will be tested.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+def gmm_sampler(tpe):
+    """Return an instance of GMMSampler"""
+    mus = numpy.linspace(-3, 3, num=12, endpoint=False)
+    sigmas = [0.5] * 12
+
+    return GMMSampler(tpe, mus, sigmas, -3, 3)
+
+
+def test_compute_max_ei_point():
+    """Test that max ei point is computed correctly"""
```
bouthilx, line 40: Code is fine, but test would be more _severe_ if you would shuffle the data to test, otherwise answer is just last item and a buggy function could easily slip through unnoticed. But again, `compute_max_ei_point` is fine. :)

src/orion/algo/tpe.py
bouthilx: Forget about it. Ready more carefully the details on the Parzen estimator in the original paper I understand that we could hardly do this with implementations in sklearn. That's fine then.

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,351 @@
...
+    if size > 1:
+        order = numpy.argsort(mus)
+        sorted_mus = mus[order]
+        prior_mu_pos = numpy.searchsorted(sorted_mus, prior_mu)
+
+        weights = update_weights(size)
+
+        mixture_mus = numpy.zeros(size + 1)
+        mixture_mus[:prior_mu_pos] = sorted_mus[:prior_mu_pos]
+        mixture_mus[prior_mu_pos] = prior_mu
```
bouthilx, line 75: Why do you add the mean of the dimension? Why not stick to points?

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+    points = numpy.linspace(-3, 3, num=10)
+    below_likelis = numpy.linspace(0.5, 0.9, num=10)
+    above_likes = numpy.linspace(0.2, 0.5, num=10)
+    max_ei_point = compute_max_ei_point(points, below_likelis, above_likes)
+    assert max_ei_point == 3
+
+
+def test_adaptive_parzen_normal_estimator():
+    """Test adaptive kernel estimator"""
+    obs_mus = numpy.linspace(-1, 5, num=30, endpoint=False)
```
bouthilx, line 50: This test should be on shuffled data as well since we need to test that the function sorts it properly.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # priori weight
+    mus, sigmas, weights = adaptive_parzen_estimator(obs_mus, low, high, prior_weight=0.5,
+                                                     equal_weight=False, flat_num=25)
+    priori_pos = numpy.searchsorted(mus, 2)
+    assert weights[priori_pos] == 0.5 / 31
+    assert numpy.all(weights[31 - priori_pos:] == weights[-1])
+    assert numpy.all(sigmas == 6 / 10)
+
+    # sigma clip
```
bouthilx, line 74: Sigmas are a bit tricky with edges being different than middle ones. I would test them separately in another unit-test and verify the edges plus the middle ones. That would require (pseudo)-random intervals as well, not `linear()`.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+        assert len(gmm_sampler.pdfs) == 12
+
+    def test_sample(self, gmm_sampler):
+        """Test GMMSampler sample function"""
+        points = gmm_sampler.sample(25)
+        points = numpy.array(points)
+
+        assert len(points) <= 25
+        assert numpy.all(points >= -3)
+        assert numpy.all(points < 3)
```
bouthilx, line 96: We should also add a test_weight which samples many points and verify that the samples are properly spread according to weights. This could be done with very different mus and small sigmas.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def test_get_loglikelis(self, gmm_sampler):
+        """Test to get log likelis of points"""
+        points = numpy.random.random(25)
+        likelis = gmm_sampler.get_loglikelis(points)
+
+        assert len(likelis) == len(points)
```
bouthilx, line 103: Values should be tested with points for which we can simply compute the log likelihood manually.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+    def test_split_trials(self, tpe):
+        """Test observed trails can be split based on TPE gamma"""
+        space = Space()
+        dim1 = Real('yolo1', 'uniform', -3, 6)
+        space.register(dim1)
+
+        tpe.space = space
+        tpe.gamma = 0.25
+
+        points = numpy.linspace(-3, 3, num=10, endpoint=False)
```
bouthilx, line 137: Should be shuffled as well to test that tpe.split_trials sorts data properly.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,220 @@
...
+        space = Space()
+        dim = Real('yolo1', 'uniform', 0.9, shape=(2, 1))
+        space.register(dim)
+
+        with pytest.raises(ValueError) as ex:
+            TPE(space)
+
+        assert 'TPE now only supports 1D shape' in str(ex.value)
+
+    def test_suggest(self, tpe):
```
bouthilx, line 182: There should also be a `test_suggest_n_initial_points` and `test_suggest_n_ei_candidates` to make sure they are supported properly. Especially for `n_initial_points`.

bouthilx:
Great! We'll have a concise and simple implementation for TPE. :)

src/orion/algo/space.py
donglinjy: The intention of this was more like to say no matter what prior specified for integer, tpe will always assume it is uniform. But I think it should be better to go with your suggestion, except tpe will only accept int_uniform. Will add this into the documentation when I have discrete and categorical support done.

src/orion/algo/tpe.py
donglinjy: This is more like the purpose of the prior_weight, take the dimension interval into consideration but people can decide the weight of it.

src/orion/algo/tpe.py
donglinjy: did not quite get this one, we support shape value > 1, but only 1D shape.

src/orion/algo/tpe.py
donglinjy: good catch :)

tests/functional/algos/test_algos.py
donglinjy: Tpe should stop this case once we get discrete space supported later. now this case is not applied to tpe algorithm.

src/orion/algo/space.py
bouthilx: Yes, I understand why we need this, but the TPE will be fooled in thinking that the integer prior was uniform while the user is asking something else. Using the formatted prior name will lead to intended value error in TPE instantiation.

src/orion/algo/tpe.py
bouthilx: Ah! My bad, I misread that tpe only supported shape == 1 sor far but latter I was confused because your TPE implementation supports 1d arrays. Great! :)

tests/functional/algos/test_algos.py
bouthilx: ok. Can you please add a TODO not just to make sure we don't miss it in next PR.

src/orion/algo/tpe.py
bouthilx: Sorry, I still don't understand. `mixture_mus` is basically the list of points. Why add the mean to this list?

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,396 @@
...
+                points.append(new_point)
+
+        return points
+
+    def _sample_real_point(self, dimension, below_points, above_points):
+        """Sample one value for a real dimension based on the observed good and bad points"""
+        low, high = dimension.interval()
+        below_mus, below_sigmas, below_weights = \
+            adaptive_parzen_estimator(below_points, low, high, self.prior_weight,
+                                      self.equal_weight, flat_num=25)
```
bouthilx, line 292: Shouldn't flat_num be a hyper-parameter of TPE? Maybe we could merge equal_weight with flat_num; if flat_num is None or equal 0, we use equal weights, otherwise we use the ramp up with flat_num.

tests/unittests/algo/test_tpe.py
```diff
@@ -0,0 +1,394 @@
...
+                                                     equal_weight=True, flat_num=25)
+    assert numpy.all(weights == 1 / 31)
+    assert numpy.all(sigmas == 6 / 10)
+
+    # prior weight
+    mus, sigmas, weights = adaptive_parzen_estimator(obs_mus, low, high, prior_weight=0.5,
+                                                     equal_weight=False, flat_num=25)
+    prior_pos = numpy.searchsorted(mus, 2)
+    assert weights[prior_pos] == 0.5 / 31
+    assert numpy.all(weights[31 - prior_pos:] == weights[-1])
```
bouthilx, line 90: The linear part should be tested as well, otherwise this test would pass even if equal_weight=False would return the same than equal_weight=True.

bouthilx:
This implementation is like in optuna where hyperparameters are assumed independent. Implementing a joint prior distribution is much trickier and isn't described in details in the paper AFAIK. The implementation of hyperopt seems to rely heavily on pyll to handle this. I would propose we start with this independent version, then add Integer/Categorical like you intended to and then tackle the joint version. Meanwhile we could add an argument independant=True and when set to False we raise NotImplementedError.

tests/unittests/algo/test_tpe.py
donglinjy: good idea

src/orion/algo/tpe.py
donglinjy: I was thinking to keep this parameter as internal at this moment, unlike equal_weight, this parameter has more impact on the performance, want to hide it with best practice value mentioned in the paper.

donglinjy:
> This implementation is like in optuna where hyperparameters are assumed independent. Implementing a joint prior distribution is much trickier and isn't described in details in the paper AFAIK. The implementation of hyperopt seems to rely heavily on pyll to handle this. I would propose we start with this independent version, then add Integer/Categorical like you intended to and then tackle the joint version. Meanwhile we could add an argument independant=True and when set to False we raise NotImplementedError.

My understanding is that hyperopt depend on pyll to describe the relative/depended parameters and optuna actually also support sample_relative and sample_independent except optuna tpe only implement sample_independent. What I am thinking is, dependency could mean partial of the hyper-parameters to search instead of them all. So I think from Orion search space definition perspective, we need first support to define the relative or depended search space.

src/orion/algo/tpe.py
```diff
@@ -0,0 +1,396 @@
...
+                     it is only valid if `equal_weight` is False.
+    """
+    def update_weights(total_num):
+        """Generate weights for all components"""
+        if total_num < flat_num or equal_weight:
+            return numpy.ones(total_num)
+
+        ramp_weights = numpy.linspace(1.0 / total_num, 1.0, num=total_num - flat_num)
+        flat_weights = numpy.ones(flat_num)
+        return numpy.concatenate([ramp_weights, flat_weights])
```
bouthilx, line 65: From this paper: http://proceedings.mlr.press/v28/bergstra13.pdf

>  The first modification was to downweight trials as they age so that old results do not count for as much as more recent ones. We gave full weight to the most recent 25 trials and applied a linear ramp from 0 to 1.0 to older trials. This is a heuristic concession to TPE’s assumption of hyperparameter independence: as our search moves through space, we use temporal distance within the experiment as a surrogate for distance in the search space.

Seems what we are doing here is to ramp up from 1 / n up to 1.

tests/unittests/algo/test_tpe.py
```diff
@@ -85,9 +85,12 @@ def test_adaptive_parzen_normal_estimator_weight():
     # prior weight
     mus, sigmas, weights = adaptive_parzen_estimator(obs_mus, low, high, prior_weight=0.5,
                                                      equal_weight=False, flat_num=25)
+    print(weights)
```
bouthilx, line 88: Left-over print. :)

src/orion/algo/tpe.py
bouthilx: The value of 25 may very well be domain specific however. They did test on different ones (LFW, PubFig83 and subset of CIFAR10), but that's arguably small evidence.

bouthilx:
> My understanding is that hyperopt depend on pyll to describe the relative/depended parameters and optuna actually also support sample_relative and sample_independent except optuna tpe only implement sample_independent. What I am thinking is, dependency could mean partial of the hyper-parameters to search instead of them all. So I think from Orion search space definition perspective, we need first support to define the relative or depended search space.

I think you are right. From the same paper:  http://proceedings.mlr.press/v28/bergstra13.pdf

> The TPE algorithm is conspicuously deficient in optimizing each hyperparameter independently of the others. It is almost certainly the case that the optimal values of some hyperparameters depend on settings of others.

Let's keep it independent then. :)

src/orion/algo/tpe.py
donglinjy: yeah, I notice this too, but seems hyperopt did this too even the paper claim the other way.

bouthilx: Mismatch between paper and code in hyperband and now TPE again! :facepalm: I would be curious to compare both version, but now isn't a good time. Maybe when we start benchmarking the algos. Do you think it's worth adding a `paper_rampup` parameter so that we can test both versions later on? I think a pointer to the code would again be important to justify the discrepancy with the paper.

donglinjy: For pointer, since we now put the weight ramp function inside the parzen estimator which will call the ramp function, and we already has a pointer for parzen to hyperopt. I guess we can omit this one. 

For the papar_rampup, I guess we can, let's add this later together with the discrete support, as I assume there will be some other changes.

src/orion/algo/tpe.py
donglinjy: okay, let's then expose an additional parameter for tpe.

src/orion/algo/tpe.py
bouthilx: I believe it should be more explicit. We could re-use the same pointer but clearly state what we are referring to, that is, the use of an additional mid-space point and the ramp-up which is different than the paper, flat (and low) for trials before 25 last ones, and then linear up to 1.

docs/src/user/algorithms.rst
```diff
@@ -250,6 +251,11 @@ True to set equal weights for observed points. Default is ``False``.
 
 The weight given to the prior point of the input space. Default is ``1.0``.
 
+``full_weight_num``
+
+The number of the most recent trials which get the full weight where the others will be
+applied with a linear ramp from 0 to 1.0. It will only take effect if ``equal_weight``
+is ``False``. Default is ``25``.
```
bouthilx, line 258: This is consistent with the paper but not with the implementation, right?

src/orion/algo/tpe.py
donglinjy: okay, let's make it more explicitly when we have the paper_rampup weight update function, when we should move the current implementation outside parzen estimator and user can choose.

docs/src/user/algorithms.rst
donglinjy: I suddenly realize seems I misunderstand this ramp up when you said from 1/n to 1. Paper mentions ramp from 0 to 1.0, but the weight of course can not be 0, so get start from 1/n is not that different compared with the paper.

bouthilx: But it's (1/n, 1/n, ... 1/n,  growing to 1 through 25 steps), while paper says (0 growing to 1, 1, 1, ... 25 times).

PR #383, opened 2020-04-18
update asha and hyperband

donglinjy:
This PR include below updates:
1) fix a bug of ASHA when the num_brackets is larger than num_rungs.  Use num_rungs as num_brackets in such cases.

2) Add cardinality check for both asha and hyperband when the algorithms failed to suggest new samples. A warning message will be logged.

@bouthilx

### DIFF
--- a/src/orion/algo/asha.py
+++ b/src/orion/algo/asha.py
@@ -198,6 +198,25 @@ class ASHA(BaseAlgorithm):
                 logger.debug('Promoting')
                 return [candidate]
 
+        point = self._grow_point_for_bottom_rung()
+        if not point:
+            return None
+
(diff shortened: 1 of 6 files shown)


src/orion/algo/asha.py
```diff
@@ -194,18 +199,28 @@ def suggest(self, num=1):
                 return [candidate]
 
         if all(bracket.is_filled for bracket in self.brackets):
-            logger.debug('All brackets are filled.')
+            logger.warning('All brackets are filled.')
             return None
 
         for _attempt in range(100):
```
bouthilx, line 205: This function is getting complicated. I would be in favor of extracting all sampling logic (from line to 202 to 223 in new code) into another method like we do in Hyperband. What do you think?

PR #389, opened 2020-05-08
TPE discrete-categorical-loguniform space support

donglinjy:
Add discrete and categorical dimension into TPE support.

Now TPE will support uniform, uniform discrete and choices as prior. As for choices prior, the probabilities if any given will be ignored.


src/orion/algo/tpe.py
```diff
@@ -314,6 +339,51 @@ def _sample_real_point(self, dimension, below_points, above_points):
...
+        below_points = numpy.array(below_points).astype(int) - low
+        above_points = numpy.array(above_points).astype(int) - low
+
+        sampler_below = CategoricalSampler(self, below_points, choices)
+        candidate_points = sampler_below.sample(self.n_ei_candidates)
+
+        if list(candidate_points):
+            sampler_above = CategoricalSampler(self, above_points, choices)
+
+            lik_blow = sampler_below.get_loglikelis(candidate_points)
```
bouthilx, line 356:
```suggestion
            lik_below = sampler_below.get_loglikelis(candidate_points)
```

src/orion/algo/tpe.py
```diff
@@ -314,6 +339,51 @@ def _sample_real_point(self, dimension, below_points, above_points):
...
+            lik_above = sampler_above.get_loglikelis(candidate_points)
+
+            new_point = compute_max_ei_point(candidate_points, lik_blow, lik_above)
```
bouthilx, line 359:
```suggestion
            new_point = compute_max_ei_point(candidate_points, lik_below, lik_above)
```

src/orion/algo/tpe.py
```diff
@@ -314,6 +339,51 @@ def _sample_real_point(self, dimension, below_points, above_points):
...
+        below_points = [choices.index(point) for point in below_points]
+        above_points = [choices.index(point) for point in above_points]
+
+        sampler_below = CategoricalSampler(self, below_points, choices)
+        candidate_points = sampler_below.sample(self.n_ei_candidates)
+
+        if list(candidate_points):
+            sampler_above = CategoricalSampler(self, above_points, choices)
+
+            lik_blow = sampler_below.get_loglikelis(candidate_points)
```
bouthilx, line 378:
```suggestion
            lik_below = sampler_below.get_loglikelis(candidate_points)
```

src/orion/algo/tpe.py
```diff
@@ -314,6 +339,51 @@ def _sample_real_point(self, dimension, below_points, above_points):
...
+            lik_above = sampler_above.get_loglikelis(candidate_points)
+
+            new_point_index = compute_max_ei_point(candidate_points, lik_blow, lik_above)
```
bouthilx, line 381:
```suggestion
            new_point_index = compute_max_ei_point(candidate_points, lik_below, lik_above)
```

tests/unittests/algo/test_tpe.py
```diff
@@ -262,13 +370,13 @@ def test_set_state(self, tpe):
     def test_unsupported_space(self):
         """Test tpe only work for supported search space"""
         space = Space()
-        dim = Integer('yolo1', 'uniform', -2, 4)
+        dim = Fidelity('epoch', 1, 9, 3)
```
bouthilx, line 373: It could support easily Fidelity by just returning the max like random search does. `Integer('yolo1', 'normal', 0, 1)` for testing unsupported priors.

tests/unittests/algo/test_tpe.py
```diff
@@ -316,6 +424,69 @@ def test_split_trials(self, tpe):
...
+        space.register(dim2)
+
+        tpe = TPE(space)
+
+        obs_points = numpy.random.randint(-10, 10, 100)
+        below_points = [obs_points[:25]]
+        above_points = [obs_points[25:]]
+        points = tpe.sample_one_dimension(dim1, 1,
+                                          below_points, above_points, tpe._sample_int_point)
+        assert len(points) == 1
```
bouthilx, line 443: It would be good to test the values as well. That they are integers in the correct interval and correspond roughly to the best region given below_points and above_points.

tests/unittests/algo/test_tpe.py
```diff
@@ -325,20 +496,23 @@ def test_sample_real_dimension(self):
...
+        below_points = [points[:8], points[8:16]]
+        above_points = [points[16:24], points[24:]]
+        points = tpe.sample_one_dimension(dim2, 2,
+                                          below_points, above_points, tpe._sample_real_point)
         assert len(points) == 2
 
         tpe.n_ei_candidates = 0
-        points = tpe.sample_real_dimension(dim2, 2, below_points, above_points)
+        points = tpe.sample_one_dimension(dim2, 2,
+                                          below_points, above_points, tpe._sample_real_point)
```
bouthilx, line 515: I see that we didn't test for values for test_sample_real_dimensions. We could sample separately below and above with different intervals to artificially create the best regions and verify the sampling of the TPE.

donglinjy: It was same idea for int, categorical and real, as there are separate cases for gmm and categorical sampler. But sure, we can add some specific verification at these places.

tests/unittests/algo/test_tpe.py
donglinjy: It is pretty straight forward for random as all suggestions simply follow the same pattern, for tpe, we need to add separate paths to handle fidelity dimension and mention such support in doc. The only value for random or tpe to support fidelity I can think is to make it easier for user to switch from different search algorithms, or user can simply specify fixed fidelity value for random and tpe.

bouthilx: Yes the advantage of supporting fidelity dimension in algos like random search and TPE is really just to make it easier to switch from one algo to another using the same search space definition.

src/orion/algo/tpe.py
```diff
@@ -185,12 +185,13 @@ def __init__(self, space, seed=None,
...
 
         for dimension in self.space.values():
+
             if dimension.type not in ['real', 'integer', 'categorical']:
                 raise ValueError("TPE now only supports Real, Integer "
                                  "and Categorical Dimension.")
 
-            if dimension.prior_name not in ['uniform', 'int_uniform', 'choices']:
-                raise ValueError("TPE now only supports uniform, uniform discrete "
+            if dimension.prior_name not in ['uniform', 'reciprocal', 'int_uniform', 'choices']:
```
bouthilx, line 193: We could support int_reciprocal as well.

donglinjy: I saw this possiblity, but did not want to do this now for 2 reasons: 1)discrete int solution still need verify, 2) loguniform is a little bit tricky for the space range, and user needs be careful to specify the space, didn't want to make it more complicated, plus I think it is almost good enough to have uniform support for discrete it.

bouthilx: Agreed!

src/orion/algo/tpe.py
```diff
@@ -283,7 +280,8 @@ def suggest(self, num=1):
                                                        above_points[idx: idx + shape[0]],
                                                        self._sample_categorical_point)
                 else:
-                    raise NotImplementedError()
+                    # fidelity dimension
```
bouthilx, line 283: maybe `elif dimension.type == 'fidelity'` to be sure to catch unsupported dimensions that would have been missed in `__init__`

PR #398, opened 2020-06-02
Update TOX configuration

Thomsch:
# Description
These changes are refreshing `tox.ini` to increase the maintainability of the project and the good practices, and updating the linting packages.

# Changes
- `tox.ini` is now sorted by environment first. The misc package configs are moved at the end.
- The `flake8` configuration is moved to a standalone `.flake8` file  to improve linting compatibility with IDEs, and unclutter `tox.ini`.
- `dev-requirement.txt` now contains only the package `tox`. Since it's the only package required for developing. The other dependencies were duplicates of the environments dependencies specified in `tox.ini`.
- Versions for linters `pylint`, `flake8`, and `doc8` have been updated to their latest version*.
- New rules introduced by the update to the linters have been **disabled for now** **.
- Relaxed `basepython` property in `tox.ini` from `python3.6` to `python3` since the CI already specifies which minor python version to use.

# Checklist
## Documentation

## Quality

# Further comments
## * Linter version
We specify the major and minor version for the linter so our build system do not break the whole codebase when a linter adds a rule.

## ** Ignored new linting rules
The following rules are ignored for now; we have to schedule a maintenance task to fix them before accepting this PR.

 **Each rule was triggered at least once in the current code base.**

### flake8
- W504 (Line break after binary operator)
- W605 (Invalid escape sequence)
- F841 (Local variable never used)
- F632 (Use of == / != to compare constant literals)
- D401 (First line should be imperative mood)
- B008 (Do not perform function calls in argument defaults.)
- B009 (Do not call getattr with a constant attribute value, it is not any safer than normal property access)
- B011 (Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().)

### pylint
- C0415 (Used when an import statement is used anywhere other than the module top-level.)
- F0010 (Used when a `__init__.py` module file is missing.)
- R0205 (Used when a class inherit from object, which under python3 is implicit)
- R1705 (Used in order to highlight an unnecessary block of code following an if containing a return statement)
- R1711 (Emitted when a single "return" or "return None" statement is found at the end of function or method definition)
- R1720 (Used in order to highlight an unnecessary block of code following an if containing a raise statement)
- W0106 (Used when an expression that is not a function call is assigned to nothing)
- W0107 (Used when a "pass" statement that can be avoided is encountered.)
- W0127 (Emitted when we detect that a variable is assigned to itself)
- W0706 (Used when an except handler uses raise as its first or only operator. This is useless because it raises back the exception immediately)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,43 @@
+[flake8]
+count = True
+show-source = True
+doctests = True
+
+ignore =
+    # Missing docstring in __init__
+    D107
+    # blank-line after doc summaries (annoying for modules' doc)
+    D205
+    # conflicts with D211: No blank lines allowed before class docstring
+    D203
+    # do not enforce first-line-period at module docs
+    D400
+    # conflicts with E133: closing bracket is missing indentation
+    E123
+
+exclude =
+    .tox,
+    .git,
+    __pycache__,
+    docs,
+    config,
+    build,
+    dist,
+    *.pyc,
+    *.egg-info,
+    .cache,
+    .eggs,
+    src/orion/core/_version.py,
+    src/orion/core/utils/_appdirs.py
+
+# Line length
+max-line-length = 100
+
+# McCabe complexity checker
+max-complexity = 20
+
+# flake8-import-order: style
+import-order-style = google
+
+# flake8-import-order: local module name checker
+application-import-names = orion, versioneer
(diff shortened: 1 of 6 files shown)


bouthilx:
> Relaxed basepython property in tox.ini from python3.6 to python3 since the CI already specifies which minor python version to use.

Thomsch:
The new disabled rules will be re-enabled in a future PR.

PR #400, opened 2020-06-12
Plotting API Starter

Thomsch:
newplot

# Description
This pull request implement the architecture to support plotting via Orion's python API. It also implements a regret plot using the new architecture. The plots are generated using Plotly.

[See it in action](https://jsfiddle.net/rtv7u35j/show) (You may be prompted to start the fiddle. Confirm or look at the code [here](https://jsfiddle.net/rtv7u35j/))

# Changes
These changes are implemented in a new module `orion.plotting`. For ease of use, a convenience attribute inspired by `pandas` is added to `ExperimentClient` to create plots easily: `experiment.plot()` or `experiment.plot.regret()`.

# Checklist
## Tests

## Documentation

## Quality

# Further comments
The plotting backend relies on Plotly, which is separated from the plotting API if we want to add or change the plotting backend in the future.


src/orion/core/utils/tests.py
```diff
@@ -37,6 +40,74 @@ def default_datetime():
...
+    """Context manager for the creation of an ExperimentClient and storage init"""
+
+    if exp_config is None:
+        raise ValueError("Parameter 'exp_config' is missing")
+    if trial_config is None:
+        raise ValueError("Parameter 'trial_config' is missing")
+    if statuses is None:
+        statuses = ['new', 'interrupted', 'suspended', 'reserved', 'completed']
+
+    from orion.client.experiment import ExperimentClient
```
bouthilx, line 98: Can we move this import to top module?

src/orion/plotting/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""
```
bouthilx, line 1: Ideally we should reuse the same docstring format for all modules. Like
```python
"""
:mod:`orion.core.utils.tests` -- Utils for tests
================================================
.. module:: state
   :platform: Unix
   :synopsis: Helper functions for tests
"""
```

src/orion/plotting/_core.py
```diff
@@ -0,0 +1,85 @@
...
+    Raises
+    ------
+    ValueError
+        If no experiment is provided.
+
+    """
+    return backend.regret(experiment, order_by, verbose_hover, **kwargs)
+
+
+PLOT_METHODS = {'regret': regret}
```
bouthilx, line 43: I think it would be worth reusing the same factory architecture that is used for Algorithms, Databases and Strategies. A backend could have a global configuration, hence it could be useful to represent it as an object instead of a module.

src/orion/plotting/_core.py
```diff
@@ -0,0 +1,85 @@
...
+
+def regret(experiment, order_by='suggested', verbose_hover=False, **kwargs):
+    """
+    Make a plot to visualize the performance of the hyper-optimization process.
+
+    The x-axis contain the trials and the y-axis their respective best performance.
+
+    Parameters
+    ----------
+        experiment: ExperimentClient, Experiment or ExperimentView
```
bouthilx, line 12: Unless numpy docstring official format changed recently I think the body of Parameters should be only 1 indent.

src/orion/plotting/_core.py
```diff
@@ -0,0 +1,85 @@
...
+            The orion object containing the experiment data
+
+        order_by: str
+            Indicates how the trials should be ordered. Acceptable options are below.
+            See attributes of `Trial` for more details.
+
+            * 'suggested': Sort by trial suggested time (default).
```
bouthilx, line 19: Do you mean `'new'`? There is no `'suggested'` status.

src/orion/plotting/_core.py
```diff
@@ -0,0 +1,85 @@
...
+            * 'reserved': Sort by trial reserved time.
+            * 'completed': Sort by trial completed time.
+
+        verbose_hover: bool
+            Indicates whether to display the hyperparameter in hover tooltips. False by default.
```
bouthilx, line 24: Why not True by default?

src/orion/plotting/backend_plotly.py
```diff
@@ -0,0 +1,103 @@
+import pandas as pd
+import plotly.graph_objects as go
+
+
+def regret(experiment, order_by, verbose_hover, **kwargs):
```
bouthilx, line 5: No docstring?

src/orion/plotting/_core.py
```diff
@@ -0,0 +1,85 @@
+import orion.plotting.backend_plotly as backend
```
bouthilx, line 1: There is no module docstring.

Also `base.py` would be coherent with the rest of the code instead of `_core.py`.

src/orion/plotting/backend_plotly.py
```diff
@@ -0,0 +1,103 @@
+import pandas as pd
```
bouthilx, line 1: There is no module docstring.

src/orion/plotting/backend_plotly.py
```diff
@@ -0,0 +1,103 @@
...
+    data = [(trial.id, trial.status, trial.submit_time, trial.start_time,
+             trial.end_time, format_hyperparameters(trial.params), trial.objective.value) for trial in trials]
+
+    df = pd.DataFrame(data, columns=['id', 'status', ORDER_KEYS[0],
+                                     ORDER_KEYS[1], ORDER_KEYS[2], 'params', 'objective'])
+
+    df = df.sort_values(order_by)
+
+    df['best'] = df['objective'].cummin()
+    df['best_id'] = get_best_ids(df)
```
bouthilx, line 43: It would be good to extract the data preparation process. Experiment.to_frame() could even be a method would be useful in general.

src/orion/plotting/backend_plotly.py
```diff
@@ -0,0 +1,103 @@
...
+
+    y_axis_label = f"{trials[0].objective.type.capitalize()} '{trials[0].objective.name}'"
+    fig.update_layout(title=f"Regret for experiment '{experiment.name}'",
+                      xaxis_title=f"Trials ordered by {order_by} time",
+                      yaxis_title=y_axis_label)
+
+    return fig
+
+
+def get_best_ids(df):
```
bouthilx, line 66: I believe this should be doable with a single line using panda. I'd need to try out though.

src/orion/plotting/backend_plotly.py
```diff
@@ -0,0 +1,103 @@
...
+        return f"[{','.join(format_value(x) for x in value)}]"
+
+    except TypeError:
+        if isinstance(value, float):
+            return f"{value:.5G}"
+        else:
+            return value
+
+
+def format_hyperparameters(hyperparameters):
```
bouthilx, line 96: No docstring. It's a bit confusing to have template functions inside regret() and the others outside. I would be in favor of moving out the template fcts.

tests/unittests/plotting/test_plot_accessor.py
```diff
@@ -0,0 +1,77 @@
...
+
+def test_init_require_experiment():
+    with pytest.raises(ValueError):
+        PlotAccessor(None)
+
+
+def test_call_nonexistent_kind():
+    with create_experiment(config, trial_config, ['completed']) as (_, _, experiment):
+        pa = PlotAccessor(experiment)
+        with pytest.raises(KeyError):
```
bouthilx, line 52: I think ValueError would be more accurate. The error message should be clear for the user also.

tests/unittests/plotting/test_plot_accessor.py
```diff
@@ -0,0 +1,77 @@
...
+
+        check_regret_plot(plot)
+
+
+def test_call_to_regret():
+    with create_experiment(config, trial_config, ['completed']) as (_, _, experiment):
+        pa = PlotAccessor(experiment)
+        plot = pa.regret()
+
+        check_regret_plot(plot)
```
bouthilx, line 77: There would be test required for the data processing (like getting the best ids). If you extract it from regret, testing it will be easier as well.

tests/unittests/plotting/test_plotly_backend.py
```diff
@@ -0,0 +1,97 @@
...
+    assert not trace1.x
+
+    trace2 = plot.data[1]
+    assert trace2.type == "scatter"
+    assert trace2.name == "best-to-date"
+    assert trace2.mode == "lines"
+    assert len(trace2.y) == 1
+    assert not trace2.x
+
+
```
bouthilx, line 57: I think we could automate most of these tests across backends with pytest.mark.parametrize: https://docs.pytest.org/en/stable/parametrize.html

src/orion/core/utils/tests.py
```diff
@@ -37,6 +40,74 @@ def default_datetime():
...
+    """Generate Trials with different configurations"""
+
+    def _generate(obj, *args, value):
+        if obj is None:
+            return None
+
+        obj = copy.deepcopy(obj)
+        data = obj
+
+        for arg in args[:-1]:
```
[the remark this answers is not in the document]
Thomsch: You're right. This was legacy code in another test class that I moved here. I didn't see this simplification.

src/orion/core/utils/tests.py
Thomsch: Unfortunately no, it creates a circular dependency otherwise because `Experiment` has a hard dependency on the storage.

src/orion/plotting/_core.py
Thomsch: You're right. I don't know why this one is indented

src/orion/plotting/_core.py
Thomsch: My rationale was that we want something as streamlined as possible for tooltips, we would see the values of the hyperparameters in a table under the graph (another option that will be added in another PR).

src/orion/plotting/backend_plotly.py
Thomsch: Fixed in 8f4b1c4563c9ec86748c4201460aaf3c9b25bb8c

src/orion/plotting/backend_plotly.py
Thomsch: Fixed in 8f4b1c4563c9ec86748c4201460aaf3c9b25bb8c

src/orion/plotting/backend_plotly.py
Thomsch: I'm happy to extract the data preparation process. That's a great suggestion! Since the preparation is strictly related to the plot (i.e., some columns are irrelevant outside of the plot), I'm not going to export it outside.

`Experiment.to_frame()` would be useful indeed. However, wouldn't it be more polyvalent if we pass a dict of the experiment to build a dataframe in a general case?

src/orion/plotting/backend_plotly.py
Thomsch: That's what I thought too but I couldn't find a straightforward solution, unfortunately.

src/orion/plotting/backend_plotly.py
Thomsch: My rationale was that the template functions would not be used anywhere else because they are glorified macros. I wanted to avoid cluttering the file by not placing them with the helpers that had potential to be reused.

tests/unittests/plotting/test_plot_accessor.py
Thomsch: Good point!

tests/unittests/plotting/test_plot_accessor.py
Thomsch: I agree!

tests/unittests/plotting/test_plotly_backend.py
Thomsch: That would be possible if there is a common format we could check the plots against or make test adapters for each backend.

Thomsch: I think the adapter solution is the most elegant and allows us to define exactly what test information (e.g., number of elements in the second trace) each backend needs to provide.

The big con with the common format output method is that we're not sure we'll use it elsewhere in the code.

src/orion/plotting/_core.py
Thomsch: Let me know if you prefer it to be true by default

src/orion/plotting/_core.py
Thomsch: I'm not sure what changes it would entail for the current source code related to plotting.

Thomsch:
[API Visualization vertical slice](https://trello.com/c/UPJc3qJg/532-api-visualization-vertical-slice)

src/orion/plotting/_core.py
bouthilx: I agree with the simplest/leanest default option. I think the utility for the HP values is however large enough to make it a default.

src/orion/plotting/backend_plotly.py
bouthilx: It depends, if the data frame should contain the trials then the dictionary of the experiment would be insufficient. Also `some_experiment.to_frame()` would follow standard conversion methods in tools such as xarray. From what I see in `build_frame()` I would say these information (columns) are useful in general.

src/orion/plotting/backend_plotly.py
bouthilx: Alright, let's leave it like that for now.

src/orion/plotting/backend_plotly.py
bouthilx: On the other hand doing so scatters the templates around making it harder to edit them.

tests/unittests/plotting/test_plotly_backend.py
bouthilx: I'm not sure I understand what you mean. From what I see the current tests are very generic, that's why I though they could be parametrized across several backends. There would be additionally specific tests for each backend (like for the storage tests) that are not covered by the generic parametrized tests.

src/orion/plotting/_core.py
Thomsch: Decision: The architecture will be changed when the Concept class will be integrated (~0.2.0)

src/orion/plotting/backend_plotly.py
Thomsch: That's a good point. However, it's unlikely that we'll have to edit all the templates at the same time regularly, but it would be convenient to have tooltip related stuff together.

For now, since you are right that their location is more confusing than helping, I'll move them outside of the function 👍 .

Issue #392, opened 2020-05-24
Error message on missing fidelity dimension in ASHA is misleading ("ASHA cannot be used if space does contain a fidelity dimension.")


vmichals:
**Describe the bug**
The exception raised by ASHA when no fidelity dimension exists says "ASHA cannot be used if space does contain a fidelity dimension."
See [here](https://github.com/Epistimio/orion/blob/29c0dfc13637aa3add6a57ce585232d65afcc541/src/orion/algo/asha.py#L30)

**Expected behavior**
It should say "ASHA can only be used if there is at least one fidelity dimension."

**Steps to reproduce**
It's just a typo in the error message that I stumbled upon while looking for some details about the ASHA implementation.

** Possible solution**
"ASHA cannot be used if space does contain a fidelity dimension." -> "ASHA can only be used if there is at least one fidelity dimension."

Thomsch:
Whoops! Thank you for pointing this problem out!
We'll get to it very soon!

bouthilx:
![](https://github.trello.services/images/mini-trello-icon.png) [Error message on missing fidelity dimension in ASHA](https://trello.com/c/RbRsLyz5/567-error-message-on-missing-fidelity-dimension-in-asha)

bouthilx:
Closed by #402

PR #402, opened 2020-06-16
Round ASHA budget properly

### LINKED ISSUE #393: ASHA suggests points not contained in space
**Describe the bug**
When I define the fidelity dimension as fidelity(25,1000,base=2), ASHA proposes 24 for the bottom rung. Do min_resources, max_resources have to be powers of base?

**Expected behavior**
I expected the rungs to be adjusted in some way to match the min/max resources. I don't know the algorithm well, so I might be missing something.

**Steps to reproduce**
Define a search space with fidelity(25,1000,base=2) and use ASHA

**Environment**
 - OS: CentOS Linux release 7.7.1908 (Core)
 - Python version: e.g. 3.6.9
 - Oríon version: develop
 - Database: PickleDB

**Possible solution**
I think the rungs are computed [here](https://github.com/Epistimio/orion/blob/29c0dfc13637aa3add6a57ce585232d65afcc541/src/orion/algo/asha.py#L43)

There seems to be a check if the last element is greater than `max_resources` [here](https://github.com/Epistimio/orion/blob/29c0dfc13637aa3add6a57ce585232d65afcc541/src/orion/algo/asha.py#L52).
Maybe one could add a similar test for the first element not being lower than `min_resources`.

bouthilx:
# Description

It was currently rounded taking the floor value. This lead to budgets
which were below the actual fidelity dimension.

# Further comments

Fixed #393 & #392

### DIFF
--- a/src/orion/algo/asha.py
+++ b/src/orion/algo/asha.py
@@ -43,7 +43,8 @@ def compute_budgets(min_resources, max_resources, reduction_factor, num_rungs):
     budgets = numpy.logspace(
         numpy.log(min_resources) / numpy.log(reduction_factor),
         numpy.log(max_resources) / numpy.log(reduction_factor),
-        num_rungs, base=reduction_factor).astype(int)
+        num_rungs, base=reduction_factor)
+    budgets = (budgets + 0.5).astype(int)
 
     for i in range(num_rungs - 1):
         if budgets[i] >= budgets[i + 1]:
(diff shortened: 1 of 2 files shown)


bouthilx:
[ASHA suggests points not contained in space](https://trello.com/c/K2Ro4xkR/566-asha-suggests-points-not-contained-in-space)

bouthilx:
[Error message on missing fidelity dimension in ASHA](https://trello.com/c/RbRsLyz5/567-error-message-on-missing-fidelity-dimension-in-asha)

PR #403, opened 2020-06-16
Promote Oríon's agnosticity

Thomsch:
# Description
Adds a line to promote the ability of Oríon to work with any language and support for any configuration file.

# Checklist
## Tests

## Documentation

## Quality

### DIFF
--- a/docs/src/user/script.rst
+++ b/docs/src/user/script.rst
@@ -54,6 +54,8 @@ the local one passed to
 created. To access the particular working directory of a trial, see next sections
 :ref:`commandline_templates` and :ref:`env_vars`.
 
+.. _language_compatibility:
+
 Language compatibility
 ======================
 The command line works for scripts and programs in any language.
--- a/docs/src/user/script.rst
+++ b/docs/src/user/script.rst
@@ -61,6 +61,21 @@ Language compatibility
 The command line works for scripts and programs in any language.
 The only requirement is that the executed script returns a JSON string with the objective value.
 
+The format is
+
+.. code-block:: json
+
+   [
+      {
+         "name": "some-objective",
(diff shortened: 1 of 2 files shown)


bouthilx:
I feel we should give a pointer for this item as well otherwise it is likely difficult to understand what we mean. We don't have any clear documentation for the language agnostic part, probably something we should add, but for the configuration we have this: https://orion.readthedocs.io/en/latest/user/searchspace.html#configuration-file

Thomsch:
I agree! I wanted to put a link but couldn't find any relevant resources. Thank you for the link.
Maybe we could add a paragraph or two in https://orion.readthedocs.io/en/latest/install/gettingstarted.html#integration ?

bouthilx:
> Maybe we could add a paragraph or two in https://orion.readthedocs.io/en/latest/install/gettingstarted.html#integration ?

Good idea, that would be the right place I think! Would you want to do it in this PR or we create a card for later?

docs/src/user/script.rst
```diff
@@ -54,6 +54,11 @@ the local one passed to
 created. To access the particular working directory of a trial, see next sections
 :ref:`commandline_templates` and :ref:`env_vars`.
 
+Language compatibility
+======================
+The command line works for scripts and programs in any language.
+The only requirement is that the executed script returns a JSON string with the objective value.
+
```
bouthilx, line 61: The format is
```
[
    {
        'name': 'some-objective',
        'type': 'objective',
        'value': 1
    }
]
```

See https://orion.readthedocs.io/en/latest/code/client.html#orion.client.report_results for more details.

Thomsch: Very good!

Issue #395, opened 2020-05-25
ASHA promotes unfinished trials


vmichals:
**Describe the bug**
ASHA promotes trials that are not yet completed in the next lower rung.

**Expected behavior**
From my understanding ASHA should not promote trials that didn't report a result.

**Steps to reproduce**
Use the following minimal dummy experiment script and make it executable:
```python
#!/usr/bin/env python
from orion.client import report_objective
from numpy.random import randn
import time
time.sleep(120) # set this large enough to be able to start multiple workers
report_objective(randn())
# vim: set ts=4 sw=4 sts=4 expandtab:
```
Run multiple (for me it happened with 3-4) workers (`clusterssh` is very convenient for this) with
```bash
orion hunt -c planetgc_orion_config.yaml -n test_asha --worker-max-trials 5 test.py --epochs~'fidelity(1,64,base=2)' --learning-rate~'loguniform(.00001,.01)' --momentum~'uniform(.7,.95)'
```

Fetch trial details with the following python script:
```python
from orion.client import create_experiment
exp = create_experiment(name='test_asha')
print(f"{'STATUS':^10} {'START':^30} {'END':^30} {'FIDELITY':^8} {'PARAMS_HASH':^35}")
print('-'*120)
for trial in exp.fetch_trials():
    print(f"{trial.status:>10} {str(trial.start_time):>30} {str(trial.end_time):>30} {trial.params['/epochs']:>8} {trial.hash_params:>35}")
```

This produces the following output for me:
```
  STATUS               START                           END               FIDELITY             PARAMS_HASH            
------------------------------------------------------------------------------------------------------------------------
 completed     2020-05-25 04:19:12.680012     2020-05-25 04:19:43.317895        1    72227ac933931b34ff548bc8e76edb5a
 completed     2020-05-25 04:19:12.724151     2020-05-25 04:19:43.362414        1    e8e56812294709896f4d063bfeb2d665
 completed     2020-05-25 04:19:13.859707     2020-05-25 04:19:44.524475        2    72227ac933931b34ff548bc8e76edb5a
 completed     2020-05-25 04:19:13.923803     2020-05-25 04:19:44.561748        1    6c69424baa02529ed192c149f3710f61
 completed     2020-05-25 04:17:09.541189     2020-05-25 04:17:40.249103        1    46b411cdecce47ec49da2bcb8785361c
...
```
[...]

PR #404, opened 2020-06-17
Handle producer config in experiment section

bouthilx:
# Description
The user may (it happened, see #395 ) simply transfer the producer config
to experiment section, in which case it will ignored.

# Changes

Verify if producer is defined in experiment section of config and handle it.

Update docs according to new config format


Thomsch:
[Improve doc clarity on usage of strategy config item](https://trello.com/c/SlCT9uD3/542-improve-doc-clarity-on-usage-of-strategy-config-item)

PR #405, opened 2020-06-17
Update Plugins documentation section

bouthilx:
# Description

The doc was referring to the test algo gradient descent instead of
reusing the extensive documentation of the cookiecutter.

### DIFF
--- a/docs/src/plugins/algorithms.rst
+++ b/docs/src/plugins/algorithms.rst
@@ -2,37 +2,246 @@
 Developing algorithms
 *********************
 
-The absolute bare **minimum interface** an optimization algorithm needs to have
-is an ``observe`` method and a ``suggest`` method, so that:
+The documentation here explains how to create new algorithm plugins based on the
+`cookiecutter <https://github.com/Epistimio/cookiecutter-orion.algo>`_.
 
- 1. An algorithm can ``observe`` the **results** associated with the evaluation
-    of a list of points in the parameter space. Using the history of evaluated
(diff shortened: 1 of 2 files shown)


docs/src/plugins/algorithms.rst
```diff
@@ -2,37 +2,246 @@
...
+
+Implementation
+==============
+
+.. code-block:: python
+
+    requires = 'real'
+
+Some algorithms requires the search space to be real. You can specify this requirement by adding the
+attribute ``requires = 'real'`` at the base of the class definition.In this case, the algorithm
```
Thomsch, line 136:
```suggestion
attribute ``requires = 'real'`` at the base of the class definition. In this case, the algorithm
```

docs/src/plugins/algorithms.rst
```diff
@@ -2,37 +2,246 @@
...
+wrapper in Orion's core will convert the search space to real one before passing it to your
+algorithm. This way the user can define discrete or categorital dimensions while using algorithms
```
Thomsch, line 138:
```suggestion
algorithm. This way, the user can define discrete or categorital dimensions while using algorithms
```

Thomsch:
In ´docs/src/plugins/install.rst´, we mention that
> Note that active development is currently focused on the core. Therefore only one algorithm plugin is available for now.

Is it still true?

bouthilx:
Still true indeed. Hopefully it will soon not be true anymore. :)

Thomsch:
Ok! I wasn't sure if soon and reality had collided yet :)

Issue #328, opened 2020-02-03
Stacktrace when executing oríon without arguments


Thomsch:
When executing orion without arguments, I get a stacktrace:

```
~$ orion
Traceback (most recent call last):
  File "/home/thomsch/venv/pycharm/bin/orion", line 10, in <module>
    sys.exit(main())
  File "/home/thomsch/venv/pycharm/lib/python3.7/site-packages/orion/core/cli/__init__.py", line 39, in main
    orion_parser.execute(argv)
  File "/home/thomsch/venv/pycharm/lib/python3.7/site-packages/orion/core/cli/base.py", line 73, in execute
    args, function = self.parse(argv)
  File "/home/thomsch/venv/pycharm/lib/python3.7/site-packages/orion/core/cli/base.py", line 67, in parse
    function = args.pop('func')
KeyError: 'func'
``` 
The expected behavior is to execute a default behaviour (for example `ls` lists the files and folder in the current directory) or show a helpful message to the user, usually the content of `--help`.

In orion's context, I believe showing the help would be the most appropriate action to execute.

Thomsch:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle `orion` without arguments](https://trello.com/c/ofHMX5pi/390-handle-orion-without-arguments)

Thomsch:
This indesirable behaviour is also present for the following subcommands:
- `init_only`
- `hunt`
- `db`
- `insert`
- `info`

Thomsch:
![](https://github.trello.services/images/mini-trello-icon.png) [https://github.com/Epistimio/orion/issues/328](https://trello.com/c/5KPINjGf/554-https-githubcom-epistimio-orion-issues-328)

Thomsch:
![](https://github.trello.services/images/mini-trello-icon.png) [https://github.com/Epistimio/orion/issues/328](https://trello.com/c/5KPINjGf/554-https-githubcom-epistimio-orion-issues-328)

Thomsch:
![](https://github.trello.services/images/mini-trello-icon.png) [https://github.com/Epistimio/orion/issues/328](https://trello.com/c/5KPINjGf/554-https-githubcom-epistimio-orion-issues-328)

Thomsch:
Fixed in #408

PR #408, opened 2020-06-18
Print help when calling `orion` alone

bouthilx:
Why:

Orion was crashing with a confusing error message.

How:

Detect if no command is defined, if so print help message and exit.


Thomsch:
[Stacktrace when executing oríon without arguments](https://trello.com/c/5KPINjGf/554-stacktrace-when-executing-or%C3%ADon-without-arguments)

Thomsch:
Closes #328.

PR #410, opened 2020-06-18
Raise exception when no prior is provided

Thomsch:
# Description
When a user execute `orion hunt` without any prior the application raises an unrelated exception without any explanation to the user.

# Changes
The number of priors is verified during the parsing and inform the user that the command requires at least one prior.

# Checklist
## Tests

## Documentation

## Quality

### DIFF
--- a/src/orion/core/cli/hunt.py
+++ b/src/orion/core/cli/hunt.py
@@ -59,10 +59,6 @@ def main(args):
     args['leafs'] = []
     # TODO: simplify when parameter parsing is refactored
     experiment = experiment_builder.build_from_args(args)
-
-    if len(experiment.space) == 0:
-        raise ValueError("No prior found. Please include at least one.")
-
     config = experiment_builder.get_cmd_config(args)
     worker_config = orion.core.config.worker.to_dict()
     if config.get('worker'):
--- a/src/orion/core/cli/hunt.py
+++ b/src/orion/core/cli/hunt.py
@@ -59,6 +59,10 @@ def main(args):
     args['leafs'] = []
     # TODO: simplify when parameter parsing is refactored
     experiment = experiment_builder.build_from_args(args)
+
+    if len(experiment.space) == 0:
+        raise ValueError("No prior found. Please include at least one.")
+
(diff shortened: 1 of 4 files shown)


Thomsch:
[`hunt` without parameters to optimize](https://trello.com/c/tyhc62i9/394-hunt-without-parameters-to-optimize)

src/orion/core/cli/hunt.py
```diff
@@ -59,6 +59,10 @@ def main(args):
     args['leafs'] = []
     # TODO: simplify when parameter parsing is refactored
     experiment = experiment_builder.build_from_args(args)
+
+    if len(experiment.space) == 0:
```
bouthilx, line 63: This problem would affect `init_only` as well. We could include this test in experiment_builder.build instead.

Thomsch: Ok! Wasn't really sure where to test. I will change this. This will allow me to also display the help instead of an exception :)

Thomsch:
How can I specify that an empty space is only an error in `hunt` and `init_only` environments?

src/orion/core/io/experiment_builder.py
```diff
@@ -204,6 +204,9 @@ def build(name, version=None, branching=None, **config):
         raise NoConfigurationError(
             'Experiment {} does not exist in DB and space was not defined.'.format(name))
 
+    if len(config['space']) == 0:
+        raise ValueError("No prior found. Please include at least one.")
```
bouthilx, line 208: We would need to update the documentation here:
https://github.com/Epistimio/orion/blob/develop/src/orion/client/__init__.py#L145

bouthilx: Also, ValueError is not detected by CLI but NoConfigurationError is (printing user-friendly error message vs full stack-trace). I would be in favor of merging the two are reword the error message if necessary. It's essentially the same error.

Thomsch: I opted to change the exception to `NoConfigurationError`, this way the message is still precise enough when shown to the user.

PR #420, opened 2020-07-02
Drop support for orion < 0.1.6

bouthilx:
Why:

Versions < 0.1.6 where available on pypi's orion.core which was deleted to avoid confusion.


bouthilx:
[Fix backward compatibility tests](https://trello.com/c/8O6yoYw7/591-fix-backward-compatibility-tests)

PR #422, opened 2020-07-08
Update macos image

bouthilx:
Why:

The setup for macos is horribly slow on travis. Updating to xcode9.4
image makes python 3.6 available and allows skipping multiple steps of
install.


bouthilx:
Seems to be a lost cause. We need to update brew to support mongodb service, but updating brew takes more than 15mins. There is little we can optimize. :'(

PR #423, opened 2020-07-08
REST Service API

Thomsch:
# Description
Fetching information from Oríon is currently available through the command line and the Python API. This pull request adds a new way to query information in the form of a REST API to interface with systems such as web applications. Specifically, this API was built in preparation for the Oríon web dashboard. 

This pull request adds the basic architecture and one endpoint to serve a list of experiments.

The API specification is available [here](https://docs.google.com/document/d/1kl6vlj4KqRW-yTCW_-5r4vxp5IhbkevsE-9zR78Zjfs/edit?usp=sharing)

# Changes
This pull request adds the basic architecture and one endpoint to serve a list of experiments.

- A new command `serve` is introduced. This command aims to be the point of entry for any future service-oriented features such as started a web app, or a REST service. 
- The `/experiments` endpoint which serves the list of the experiments

The REST service is composed of a WSGI app and a host container. The WSGI app is implemented using [Falcon](https://falcon.readthedocs.io/)(chosen over Flask for its resource object-oriented architecture, speed, modularity, and REST business domain). The host container is [gunicorn](https://gunicorn.org/) because of its simplicity, and customizability.

# Checklist
## Tests

## Documentation

## Quality

# Further comments
The information below is already available in the [specification](https://docs.google.com/document/d/1kl6vlj4KqRW-yTCW_-5r4vxp5IhbkevsE-9zR78Zjfs/edit?usp=sharing) but is summarized here for convenience.

Endpoint | PR
---------- | ---
`/` | **#423**
`/experiments` | **#423**
`/experiments/:name` | **#423**
`/trials/:exp_name` | #A
`/trials/:exp_name/:id` | #B
`/plots/:kind` | #C


bouthilx:
Great start! I think it would be great to list in the description of the PR all API calls that would be eventually supported and mark those that are planned to be implemented in this particular PR.

tests/functional/serving/test_root_resource.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+"""Perform functional tests for the REST endpoint `/`"""
+
+
+def test_orion_version(client):
```
bouthilx, line 5: Do you think merging these 3 tests in one would be simpler and more efficient?

Thomsch: Sure :). I was doing TDD so it was handy to have one of each, especially since there is no variations to test but one test also works 👍

docs/src/user/rest_api.rst
```diff
@@ -0,0 +1,16 @@
+The REST API provides another way to retrieve information about the experiments. The REST server
+serves requests to obtain information about trials and experiments.
+
+The API serves the following endpoints.
+
+* experiments/
+* experiments/:name
+* trails/:exp_name
```
bouthilx, line 8:
```suggestion
* trials/:exp_name
```

docs/src/user/rest_api.rst
```diff
@@ -0,0 +1,16 @@
...
+* trials/:exp_name/:id
+* plots/:kind
+
+
+The REST server is started by the command line ``$ orion serve``. The service is hosted through
+`gunicorn <https://gunicorn.org/>`. The database and host details are configured through the option
+``--config <file>``. Storage options are expressed in the configuration file in the same fashion as
+for other commands. Gunicorn options are placed under the top-level node ``gunicorn``.
```
bouthilx, line 16: It may be good to have a configuration file example.

src/orion/serving/experiments_resource.py
```diff
@@ -0,0 +1,160 @@
...
+            resp.body = json.dumps(response, indent=4)
+
+        except NoConfigurationError:
+            response = {
+                "message": f"Experiment '{name}' does not exist"
+            }
+            resp.body = json.dumps(response, indent=4)
+            resp.status = falcon.HTTP_404
+
+        except ValueError as exception:
```
bouthilx, line 82: What are the expected ValueError?

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def test_latest_versions(self, client):
+        """Tests that the API return the latest versions of each experiment"""
+        expected = [
+            {'name': 'a', 'version': 2},
+            {'name': 'b', 'version': 1}
+        ]
+
+        _add_experiment(name='a', version=1, _id=1)
+        _add_experiment(name='a', version=2, _id=2)
```
bouthilx, line 83: What if version 2 comes before version 1 in the db? Adding 3 versions in order 1, 3, 2 in the database would make sure the response does not simply return the first or last version in order of entry.

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+        Tests that a 404 response is returned when the experiment
+        doesn't exist in the database
+        """
+        response = client.simulate_get('/experiments/a')
+
+        assert response.status == "404 Not Found"
+        assert response.json['message']
+        assert response.json['message'] == "Experiment 'a' does not exist"
+
+        _add_experiment(name='a', version=1, _id=1)
```
bouthilx, line 106: When add a and test for b? You want to test when there is at least one exp in the DB to make sure the request would fail properly?

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+        assert len(response.json['user'])
+        assert response.json['orionVersion'] == "x.y.z"
+
+        _assert_config(response.json['config'])
+        _assert_best_trial(response.json['bestTrial'])
+
+    def test_default_is_latest_version(self, client):
+        """Tests that the latest experiment is returned when no version parameter exists"""
+        _add_experiment(name='a', version=1, _id=1)
+        _add_experiment(name='a', version=2, _id=2)
```
bouthilx, line 136: Same comment here for version number.

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+
+        assert response.status == "200 OK"
+        assert response.json['version'] == 2
+
+    def test_unknown_parameter(self, client):
+        """
+        Tests that if an unknown parameter is specified in
+        the query string, an error is returned
+        """
+        _add_experiment(name='a', version=1, _id=1)
```
bouthilx, line 159: Do we need experiment 'a' in db for this test?

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+        response = client.simulate_get('/experiments/a?unknown=true')
+
+        assert response.status == "400 Bad Request"
+        assert response.json == \
+               "Parameter 'unknown' is not supported. Expected parameter 'version'."
+
+
+def _add_experiment(**kwargs):
+    """Adds experiment to the dummy orion instance"""
+    base_experiment.update(kwargs)
```
bouthilx, line 170: This update may cause nasty side-effects in the future if we are to reuse these methods more globally for tests. We should at least make a deep copy and update the local copy.

tests/unittests/serving/test_experiments_endpoint.py
```diff
@@ -0,0 +1,209 @@
...
+    get_storage().create_experiment(base_experiment)
+
+
+def _add_trial(**kwargs):
+    """Add trials to the dummy orion instance"""
+    base_trial.update(kwargs)
```
bouthilx, line 176: Same comment here regarding the update()

src/orion/serving/experiments_resource.py
Thomsch: It is raised by `_verify_experiment_parameters()` if a parameter given in the query string doesn't exist. I agree it's a bit too general to catch here. In this case, do you think it's better overall the raise a custom exception and catch it, or to simply make `_verify_experiment_parameters()` return a bool and return immediately with the error code if a parameter is unsupported.

tests/unittests/serving/test_experiments_endpoint.py
Thomsch: Exactly. I'm testing against an empty database and on a "full" database to make sure the behavior is consistent.

tests/unittests/serving/test_experiments_endpoint.py
Thomsch: Yes. Because I want to distinguish between having only an unsupported parameter versus having an unsupported parameter with a non-existing experiment. I updated the test to remove the ambiguity and test explicitly the latter case.

tests/unittests/serving/test_experiments_endpoint.py
Thomsch: Sounds good for now! I'll move these when testing the endpoint trials/:experiment so they will be easier to change when reused more globally.

src/orion/serving/experiments_resource.py
bouthilx: I would favor the bool approach in this case. It's quite local, so custom error would not be very reusable.

tests/functional/serving/test_experiments_endpoint.py
```diff
@@ -75,13 +76,14 @@ def test_send_name_and_versions(self, client):
     def test_latest_versions(self, client):
         """Tests that the API return the latest versions of each experiment"""
         expected = [
-            {'name': 'a', 'version': 2},
+            {'name': 'a', 'version': 3},
             {'name': 'b', 'version': 1}
         ]
 
         _add_experiment(name='a', version=1, _id=1)
+        _add_experiment(name='a', version=3, _id=3)
```
bouthilx, line 84: I would swap the ids to avoid having the DB reshuffle the data according to ID. So (version=3, _id=2), (version=2, _id=3).

tests/functional/serving/test_experiments_endpoint.py
```diff
@@ -133,12 +135,13 @@ def test_experiment_specification(self, client):
     def test_default_is_latest_version(self, client):
         """Tests that the latest experiment is returned when no version parameter exists"""
         _add_experiment(name='a', version=1, _id=1)
+        _add_experiment(name='a', version=3, _id=3)
```
bouthilx, line 138: Same here for ids.

src/orion/serving/experiments_resource.py
Thomsch: Sounds good. I agree. I wasn't sure what was the most pythonic way to solve this.

Thomsch:
I replaced the checking mechanism for parameters. In the next PR, I refactor code with trials to have a smoother source code.
Basically, one separate function in its own module will take the query parameters and verify them.

src/orion/serving/experiments_resource.py
```diff
@@ -43,9 +43,16 @@ def on_get_experiment(self, req: Request, resp: Response, name: str):
...
+            response = {
+                "message": error_message
+            }
+            resp.status = falcon.HTTP_400
+            resp.body = json.dumps(response)
+            return
 
+        try:
             version = int(req.params['version']) if 'version' in req.params else None
             experiment = experiment_builder.build_view(name, version)
```
bouthilx, line 58: It may be worth moving the response creation outside of the try-except. There is nothing we want to catch there anyway, right?

Thomsch: I agree. I already did this refactoring in the branch for the trial endpoint. It will be updated then.

bouthilx: That's for another PR right?

PR #424, opened 2020-07-08
Ci/update coverage

bouthilx:
Why:

Configuration was outdated and was not ignoring the files under tests/

How:

Use pytest-cov instead of coverage and move to .coveragerc instead of codecov.yaml.


Thomsch:
[Fix code coverage](https://trello.com/c/6pBBVHOs/594-fix-code-coverage)

bouthilx:
I will work on this in another PR when I am back from vacations.

PR #426, opened 2020-07-15
Analysis module

Thomsch:
# Description
The pull request creates a new module `orion.analysis` to host hyperparameter optimization tools such as regret, lpi, and so on. Having a dedicated module for these operations is beneficial in the long-run when more analytic capabilities will be added and that those will be used in other contexts than plotting (i.e, as an API).

# Changes
Currently, the only analysis of this kind is the computation of the regret done in the homonymous plotting function. The regret analysis has been moved to the new `orion.analysis` module.

# Checklist
## Tests

## Quality


src/orion/analysis/regret.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Returns
+    -------
+    A copy of the original dataframe with two new columns containing respectively the best value
+    so far and its trial id.
+    """
+    df = pd.DataFrame(trials, copy=True)
+    if df.empty:
+        return df
+
```
bouthilx, line 34: Maybe we should test whether the provided tuple of names have the right format (two names) and raise a user-friendly error if it's the case.

tests/unittests/analysis/test_regret.py
```diff
@@ -0,0 +1,126 @@
...
+
+    assert 'something' in result.columns
+    assert 'else' in result.columns
+
+
+def test_regret_sequential():
+    """Tests that cumulative best objectives are linked to their respective ids"""
+    data = pd.DataFrame(data={
+        'id': ['a', 'b', 'c', 'd'],
+        'objective': [10, 12, 8, 9],
```
bouthilx, line 81: We should test also in case where there is many equals. I believe current implementation would override the ids. I would be in favor of keeping previous id in case of equality.

Thomsch: Fixed in 7798278

bouthilx:
[Analysis Module](https://trello.com/c/POdQtRfU/609-analysis-module)

PR #429, opened 2020-07-16
Quit if code changed between two trials execution

bouthilx:
Why:

We test the code version at beginning of experiment but it could change
during an execution, leading to trials ran with different code versions.

How:

Validate before the execution of the trial that code version did not
change. If so, raise BranchingEvent().

Note:

There may be race conditions between the validation and the execution of the trial.
The only solution for this would be to test the code version at
beginning of trial execution but that would require reimplementation for
all languages of users' scripts...


bouthilx:
[If code changes between two trials this goes undetected](https://trello.com/c/csiTiYcF/277-if-code-changes-between-two-trials-this-goes-undetected)

PR #433, opened 2020-07-31
Specialized properties for Trials

Thomsch:
# Description
Trial supports results of type 'objective', 'constraint', 'gradient', and 'statistic'; it's not possible to fetch more than one result. For the results of type 'gradient' and 'objective' it's acceptable since multi-objective optimization is not yet supported. However, multiple results of type 'constraint' and 'statistic' can already be supported.

# Changes
This pull request adds properties to fetch the results of type 'constraint' and 'statistic'.

# Checklist
## Tests

## Documentation

## Quality

### DIFF
--- a/tests/unittests/core/test_trial.py
+++ b/tests/unittests/core/test_trial.py
@@ -123,7 +123,7 @@ class TestTrial(object):
     def test_str_trial(self, exp_config):
         """Test representation of `Trial`."""
         t = Trial(**exp_config[1][1])
-        assert str(t) == "Trial(experiment='supernaedo2-dendi', status='completed', "\
+        assert str(t) == "Trial(experiment='supernaedo2-dendi', status='completed', " \
                          "params=/decoding_layer:lstm_with_attention,/encoding_layer:gru)"
 
     def test_str_value(self, exp_config):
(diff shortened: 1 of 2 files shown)


Thomsch:
[Ajouter des getters pour les résultats de type ``constraint`` et ``gradient`` dans un ``Trial``](https://trello.com/c/eREXesTs/634-ajouter-des-getters-pour-les-r%C3%A9sultats-de-type-constraint-et-gradient-dans-un-trial)

tests/unittests/core/test_trial.py
```diff
@@ -208,6 +208,72 @@ def test_gradient_property(self, exp_config):
         assert t.gradient.value == [5, 3]
         tmp = exp_config[1][1]['results'].pop()
 
+    def test_constraints_property(self, exp_config):
+        """Tests the property for accessing constraints"""
+        # 0 result of type 'constraints' exist
+        tmp = exp_config[1][1]['results'].pop(1)
+        t = Trial(**exp_config[1][1])
```
bouthilx, line 215: Unless it makes fail something that I'm not aware of, you could use a blank Trial to test the constraints only. It would be better if we can avoid depending on the global exp_config.

tests/unittests/core/test_trial.py
```diff
@@ -208,6 +208,72 @@ def test_gradient_property(self, exp_config):
...
+        trial = Trial(**exp_config[1][1])
+
+        assert expected == trial.constraints
+
+        # > 1 results of type 'constraint' exist
+        expected = [
+            Trial.Result(name='contra', type='constraint', value=1.2),
+            Trial.Result(name='a', type='constraint', value=5),
+            Trial.Result(name='b', type='constraint', value=20)
+        ]
```
bouthilx, line 233: You should add a result that is not a constraint to test that they are filtered out properly.

tests/unittests/core/test_trial.py
```diff
@@ -208,6 +208,72 @@ def test_gradient_property(self, exp_config):
...
+
+        trial = Trial(**exp_config[1][1])
+
+        assert expected == trial.statistics
+
+        # > 1 results of type 'statistic' exist
+        expected = [
+            Trial.Result(name='a', type='statistic', value=5),
+            Trial.Result(name='b', type='statistic', value=20)
+        ]
```
bouthilx, line 268: Same here

tests/unittests/core/test_trial.py
bouthilx: Or maybe there are results already coming from exp_config? In this case I would be in favor of hard-coding the additional results here and using a blank trial if possible.

Thomsch: Yeah, there is already results in this variable. It will be explicit when I replace _exp_config_ by trials.

tests/unittests/core/test_trial.py
Thomsch: I agree. I did it like that originally to mimics the rest of the module.

tests/unittests/core/test_trial.py
```diff
@@ -156,65 +156,143 @@ def test_invalid_result(self, exp_config):
...
-    def test_objective_property(self, exp_config):
+    def test_objective_property(self):
         """Check property `Trial.objective`."""
+        base_trial = {
+            'results': []
+        }
+        base_trial['results'].append({'name': 'a', 'type': 'gradient', 'value': 0.5})
+
+        # 0 objective in `results` list
+        trial = Trial(**base_trial)
```
bouthilx, line 167: May be worth simplifying? Dunno... :man_shrugging: 
```suggestion
        trial = Trial(results=[{'name': 'a', 'type': 'gradient', 'value': 0.5}])
```

bouthilx: Ah, append later on. Forget it.

PR #436, opened 2020-08-07
Deprecate init_only and add hunt --init-only

bouthilx:
Why:

The command init_only was confusing for many users and --init-only as an
argument to `hunt` is more intuitive. Users tended to look for an option
of hunt to have init only instead of looking for another argument.

Note:

The regression tests for backward compatibility relies heavily on
init_only and will need to be adjusted to keep using it for previous
version but use `hunt --init-only` starting with next release. This
modification should be done in release v0.1.10, otherwise tests will
fail in develop until v0.1.9 is released.


src/orion/core/worker/trial.py
```diff
@@ -283,11 +283,6 @@ def objective(self):
         """
         return self._fetch_one_result_of_type('objective')
 
-    @property
```
Thomsch, line 286: This code change is already fixed in the current develop. We might want to remove it to avoid unnecessary merge conflicts?

tests/functional/backward_compatibility/test_versions.py
```diff
@@ -99,11 +99,13 @@ def get_virtualenv_dir(version):
 
 def fill_from_cmdline_api(orion_script, version):
     """Add experiments and trials using the commandline API"""
+    # TODO: Adapt when removing init_only after deprecation phase
```
Thomsch, line 102: We might want to add a version target for these TODO so they don't get lost in time and are easy to reference from an action item.

tests/unittests/core/worker/test_trial.py
```diff
@@ -288,39 +286,6 @@ def test_statistics_property(self):
 
         assert expected == trial.statistics
 
-    def test_statistics_property(self, exp_config):
```
Thomsch, line 291: Same comment as before. These code changes are already in develop.

bouthilx:
[Faire de `init_only` une option de `orion hunt`](https://trello.com/c/OnS6BiUM/579-faire-de-initonly-une-option-de-orion-hunt)

PR #439, opened 2020-08-14
Add orion.client.get_experiment()

Thomsch:
# Description
`orion.client` is the public-facing API to work with experiments. However, there is currently no intuitive way to retrieve an experiment for the only viewing past results. Currently, the options are to either use `orion.client.create_experiment()` which is semantically opposed to what the user wants to do or use `orion.core.io.experiment_builder.build_view()` which is located in internal modules.

# Changes
This PR adds an intuitive method to retrieve existing experiments through `orion.client.get_experiment`. An instance of `orion.core.worker.experiment.ExperimentView` is returned to ensure a read-only experience.

# Checklist
## Tests

## Documentation

## Quality

### DIFF
--- a/src/orion/client/__init__.py
+++ b/src/orion/client/__init__.py
@@ -206,7 +206,8 @@ def get_experiment(name, version=None):
 
     Raises
     ------
-    :class:`orion.core.utils.exceptions.NoConfigurationError` when no experiment is found.
+    `orion.core.utils.exceptions.NoConfigurationError`
+        The experiment is not in the database provided by the user.
     """
     return experiment_builder.build_view(name, version)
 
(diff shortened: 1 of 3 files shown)


src/orion/client/__init__.py
```diff
@@ -187,6 +188,29 @@ def create_experiment(
     return ExperimentClient(experiment, producer, heartbeat)
 
 
+def get_experiment(name: str, version: int = None) -> ExperimentView:
```
bouthilx, line 191: It should additionally support storage argument like in create_experiment. With `setup_storage` like [here](
https://github.com/Epistimio/orion/pull/439/files#diff-bf2b327fcde4615d08710f7e9c57188dR162).

bouthilx, line 191: Should we limit typing to new modules for code style coherence? :)

bouthilx:
[Add orion.client.get_experiment to provide a view to users](https://trello.com/c/IkrVeyj4/630-add-orionclientgetexperiment-to-provide-a-view-to-users)

src/orion/client/__init__.py
Thomsch: Good catch!

PR #447, opened 2020-08-26
Return onehot interval instead of categorical

bouthilx:
Why:

When the categorical dimension is transformed to one-hot, the interval
returned should be in the space of the one-hot otherwise the algorithm
requesting a real space cannot handle the categories.

How:

Add an interval method to OneHot Transformer.
TransformerDimension will look for an interval method, otherwise use the
default interval of original dimension.
Note that CompositeTransformer push the interval() method recursively if
necessary, so that a one-hot encoding that is part of a composite
transformer can still provide its interval().


bouthilx:
[Return onehot interval instead of categorical](https://trello.com/c/0pKzwKtX/665-return-onehot-interval-instead-of-categorical)

PR #448, opened 2020-09-25
changed variable name in docs to reflect the correct name

mirkobronzi:
# Description
The correct name for this orion working directoy variable seems to be ORION_WORKING_DIR.
In the docs instead there is ORION_WORKING_DIRECTORY.


bouthilx:
Thanks! :)

PR #452, opened 2020-10-29
Support non existing user script conf files 

bouthilx:
# Description

If the user pass script configs to experiments, the monitoring tools will fail to fetch the corresponding experiments if the script configs cannot be found. The information necessary for monitoring is already available in the database, so we can ignore the files if they cannot be found.

# Changes

- Extend orion cmdline parser to handle command lines with non existing configuration files
- Fix bug in flatten() to handle dictionaries containing double references.
- Add force_update option in backward.populate_space so that build_view can avoid loading the config files.

## Tests

## Documentation

## Quality


bouthilx:
[If config file is not present orion status fails?](https://trello.com/c/5f44KyNJ/680-if-config-file-is-not-present-orion-status-fails)

PR #453, opened 2020-10-29
Support load yaml config for dict items

bouthilx:
Why:

Some options (like algorithm configuration) have a dict type. The
load_yaml was failing on such configuration because flatten would break
the structure of the configuration, looking at the subkeys of the item's
dictionary as if they were options in the global configuration.

How:

Turn the load_yaml into a recursive function that is called on each
subconfig. This way dictionary items (like algo conf) are not unpacked
and thus processed properly.

### DIFF
--- a/src/orion/core/io/config.py
+++ b/src/orion/core/io/config.py
@@ -18,8 +18,6 @@ import os
 
 import yaml
 
-from orion.core.utils.flatten import flatten
-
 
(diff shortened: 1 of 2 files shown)


bouthilx:
[config.load_yaml does not support dict](https://trello.com/c/J4ezI6ho/678-configloadyaml-does-not-support-dict)

PR #455, opened 2020-11-03
Fix fetch_lost_trials

bouthilx:
Why:

The fetch of lost trials should look for trials with heartbeat older
than x * heartbeat (with x > 1) otherwise we will often catch trials
that are being run (ex: heartbeat late by a microsecond). It used to be
2 * heartbeat. For some reason it was now 1 * heartbeat...

How:

Set it back to 5 * heartbeat to be sure, and rework the unittest to
verify that trials lost but with heartbeat that are not 5 * heartbeat
late yet are not fetched.


bouthilx:
[Fix fetch_lost_trials](https://trello.com/c/PrSW5IAp/683-fix-fetchlosttrials)

bouthilx:
Travis did not connect with github, but the build passed: https://travis-ci.com/github/bouthilx/orion/builds/198267482

Issue #415, opened 2020-06-28
Algorithm Benchmark in Orion


donglinjy:
**Why need benchmark**

Algorithms benchmark could be useful in some scenarios:
- Understand the capabilities of difference algorithms so people can choose the proper algorithm in different user scenarios.
- It could be a useful tool to help developer to evaluate the correctness and performance of new search algorithms.

**How it can be done**

To run benchmark, we need define the goal (Assessment Scenarios) of the benchmark and tasks (whose hyper-parameters will be optimized) that will be used.

Some possible Assessment Scenarios can be:
- Time to result
- Average performance
- Search space dimension size (the number of hyperparameters to search)
- Search space dimension Type (Rang vs Categorical)
- Parallel Execution advantage (does more worker means better performance?)
- Search Algorithm parameters sensitivity (easy to use)

And Tasks to run can be divided into 3 categories:
- Synthetic functions, e.g. Branin, rosenbrock
- Emulator/Simulator, e.g. PROFET, HPOlib
- Real verification tasks, e.g. MNIST, cifar10 

From high level view, we can run benchmark with below mechanism. 
<img width="550" alt="image" src="https://user-images.githubusercontent.com/10850941/85941309-bdbea200-b954-11ea-9de0-843ce147bb60.png">

And this can be the rough code sketch.
```
def create_benchmark():
    """
    - create a benchmark object based on user input
    - run the benchmark in backend(thread) and return the benchmark object
    - or run the benchmark in sequence and ask user to call the benchmark process()
    :return:
    """
    pass


class Benchmark():
    def __init__(self):
        """
        - build assess and task pairs
        """
        pass

    def process(self):
        """
        - for each assess-task, build Assess object and run assess execute
        - support multiple thread based a configured thread_num ?
        :return:
        """
        pass

    def status(self):
        """
        - get progress/details of each assess-task
        :return:
        """
        pass

    def analysis(self):
        """
        - display the result of each assess
        :return:
        """
        pass

    def register(self):
        """
        register benchmark object into db
        :return:
        """
        pass

class BaseAssess():

    def __init__(self, task):
        """
        - build assess object
        - build task object (from db[existing], or from config[new])
        """
        pass

    def execute(self):
        """
        - run the tasks
        - there may be needs to run the task multiple times (such as when assess average performance)
        :return:
        """
        pass

    def status(self):
        """
        - get the overall status of the assess, like how many tasks to run and the status of each task(experiment)
        :return:
        """
        pass

    def result(self):
[...]

PR #513, opened 2020-12-10
Add benchmark support into orion

donglinjy:
Search Algorithms benchmark support in Orion.

https://github.com/Epistimio/orion/issues/415

# Checklist

## Tests

## Documentation

## Quality


src/orion/storage/base.py
```diff
@@ -71,6 +71,14 @@ class BaseStorageProtocol(metaclass=AbstractSingletonType):
 
     """
 
+    def create_benchmark(self, config):
+        """Insert a new experiment inside the database"""
```
bouthilx, line 75:
```suggestion
        """Insert a new benchmark inside the database"""
```

src/orion/storage/base.py
```diff
@@ -71,6 +71,14 @@ class BaseStorageProtocol(metaclass=AbstractSingletonType):
...
+        raise NotImplementedError()
+
+    def fetch_benchmark(self, query, selection=None):
+        """Fetch all experiments that match the query"""
```
bouthilx, line 79:
```suggestion
        """Fetch all benchmarks that match the query"""
```

src/orion/storage/legacy.py
```diff
@@ -119,6 +119,14 @@ def _setup_db(self):
         self._db.ensure_index("trials", "start_time")
         self._db.ensure_index("trials", [("end_time", Database.DESCENDING)])
 
+    def create_benchmark(self, config):
+        """Insert a new experiment inside the database"""
```
bouthilx, line 123:
```suggestion
        """Insert a new benchmark inside the database"""
```

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                total_trials += status["trials"]
+                benchmark_status.append(column)
+
+        if not silent:
+            print(
+                "Benchmark name: {}, Experiments: {}/{}, Submitted trials: {}".format(
+                    self.name, complete_exp_num, total_exp_num, total_trials
+                )
+            )
+            try:
```
bouthilx, line 93: It may be worth moving this try-except in a utils function, something like ```pretty_table```.

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                print(table)
+
+        return benchmark_status
+
+    def analysis(self):
+        """Return all the assessment figures"""
+        figures = []
+        for study in self.studies:
+            figure = study.display()
+            # figure.show()
```
bouthilx, line 124:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                )
+                print(table)
+            print("Total Experiments: {}".format(len(experiment_table)))
+
+        return experiment_table
+
+    # pylint: disable=invalid-name
+    @property
+    def id(self):
+        """Id of the experiment in the database if configured.
```
bouthilx, line 174:
```suggestion
        """Id of the benchmark in the database if configured.
```

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Value is `None` if the experiment is not configured.
```
bouthilx, line 176:
```suggestion
        Value is `None` if the benchmark is not configured.
```

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+            for task in tasks:
+                str_tasks.append(task.configuration)
+            str_target["task"] = str_tasks
+        targets.append(str_target)
+        config["targets"] = targets
+
+        if self.id is not None:
+            config["_id"] = self.id
+
+        # print('config:', config)
```
bouthilx, line 208:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+        return copy.deepcopy(config)
+
+
+class Study:
+    """
+    Benchmark definition
```
bouthilx, line 214:
```suggestion
```

Maybe explain shortly what a study is in contrast to general benchmark.

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                )
+                self.experiments_info.append((task_index, experiment))
+
+    def execute(self):
+        """Execute all the experiments of the study"""
+        blackbox_fun = self.task.get_blackbox_function()
+        max_trials = self.task.max_trials
+
+        for _, experiment in self.experiments_info:
+            # TODO: it is a blocking call
```
bouthilx, line 272: I'm not sure it should be done differently. If the execution of the benchmark is to be consistent with the execution of the experiments (orion hunt), then scaling up should require running the benchmark in many workers in parallel.

src/orion/benchmark/base.py
```diff
@@ -0,0 +1,116 @@
...
+       Configurable parameters of the task, a particular task
+       implementation can have its own parameters.
+    """
+
+    def __init__(self, max_trials, **kwargs):
+        self.trials_num = max_trials
+        self._param_names = list(kwargs.keys())
+
+    @abstractmethod
+    def get_blackbox_function(self):
```
bouthilx, line 85: Do you see an advantage of returning a function instead of defining it as ```__call__``` so that we can directly work on the task object?

src/orion/benchmark/base.py
```diff
@@ -0,0 +1,116 @@
...
+    task_num : int
+       Number of experiment the assessment ask to run the corresponding task
+    kwargs : dict
+       Configurable parameters of the assessment, a particular assessment
+       implementation can have its own parameters.
+    """
+
+    def __init__(self, task_num, **kwargs):
+        self.task_number = task_num
+        self._param_names = list(kwargs.keys())
```
bouthilx, line 30: Why do you keep the param names only and not the values?

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,144 @@
...
+"""
+import datetime
+import importlib
+
+from orion.benchmark import Benchmark, Study
+from orion.core.utils.exceptions import NoConfigurationError
+from orion.storage.base import get_storage, setup_storage
+
+
+def get_or_create_benchmark(name, algorithms=None, targets=None):
```
bouthilx, line 20: Storage should be configurable like here: https://github.com/Epistimio/orion/blob/develop/src/orion/client/__init__.py#L185

src/orion/benchmark/__init__.py
donglinjy: I was thinking about 2 consideration when I added this `TODO` comment.
1) Run the experiment in a non-blocking way, so in the same notebook, we can query the state of benchmark instead of waiting for all them done.
2) Another consideration is to support run multiple experiments in parallel, which is low priority at this moment I think. This is related to what you mentioned, instead of only one way to run the benchmark code multiple times, we can also achieve some level parallel capability with one one.

src/orion/benchmark/base.py
donglinjy: `__call__` looks a more straightforward way, but I am thinking how to guarantee the sub-task implementation will always get this `__call__`. What's your suggestion?

src/orion/benchmark/base.py
donglinjy: good point. I can actually simply just keep the `kwargs`.  I guess repeating changing code make you forget the original purpose :)

src/orion/benchmark/benchmark_client.py
donglinjy: yeah, we can. 

I was thinking to make it using global configure and add a python api version interface for such kind configure, just like below. 
```
orion.init(storage=xxx, ...)
```
Because we now have benchmark and experiment client both require similar configure, and there may be more in the future, instead of asking different/same values in each call, we can make user initialize it at the beginning, as for one notebook or py code session, they should share the same configure. And if there is requirement to change some configure in the middle, we can provide additional interfaces.


But for sure, before we have this init interface, let me add the storage configure into benchmark call first. :)

src/orion/benchmark/base.py
donglinjy: another possible way is that we add `__call__` into the base task and replace this `get_black_box_function` with a `call`, and in the `__call__`, we call the sub-class `call()`.

bouthilx: Can we make the `__call__` an abstract method? Otherwise I think we should make it raise `NotImplemented` error so that subclasses that fails to implement it fails in an obvious way.

src/orion/benchmark/base.py
bouthilx: Also shouldn't the kwargs (both keys and values) be part of the configuration (`def configuration` below)?

src/orion/benchmark/benchmark_client.py
bouthilx: Providing the init in the client is a good idea. I would however keep the storage config argument in create_experiment and create_benchmark so that users aren't forced to call init() when unnecessary. It simplifies the process for users. Actually, `setup_storage` kind of plays this role already, we would only need to add it an to the client helper functions.

docs/src/user/benchmark.rst
```diff
@@ -0,0 +1,30 @@
+**********
+Benchmark
+**********
+
+You can benchmark the performance of search algorithms with different tasks at different
+assessment levels.
+
+
+Benchmark can be created as below, refer :doc:`/code/benchmark/benchmark_client`
```
bouthilx, line 9:
```suggestion
Benchmark can be created as below, refer to :doc:`/code/benchmark/benchmark_client`
```

docs/src/user/benchmark.rst
```diff
@@ -0,0 +1,30 @@
+**********
+Benchmark
+**********
```
bouthilx, line 3: You should point to the jupiter notebook example in this doc.

src/orion/benchmark/assessment/averagerank.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+import pandas as pd
+import plotly.express as px
+
+from orion.benchmark.base import BaseAssess
+
+
+class AverageRank(BaseAssess):
+    """
+    For each algorithm, run fixed number of Experiment, average the rank of trials for
```
bouthilx, line 24: Sorry, I am unable to understand the description here. Could you please try to rephrase it? :sweat_smile:

src/orion/benchmark/assessment/averagerank.py
```diff
@@ -0,0 +1,116 @@
...
+    For the performance of trials in an Experiment, instead using the actual trial objective value,
+    here we use the best objective value in the same Experiment until the particular trial.
+    """
+
+    def __init__(self, task_num=1):
+        super(AverageRank, self).__init__(task_num=task_num)
+
+    def plot_figures(self, task, experiments):
+        """
+        Generate a `plotly.graph_objects.Figure`
```
bouthilx, line 35: You should explain here what the plots are. Regret curves?

src/orion/benchmark/assessment/averagerank.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        task_algorithm_exp = defaultdict(list)
+
+        for task_index, exp in experiments:
+            algorithm_name = list(exp.configuration["algorithms"].keys())[0]
+
+            trials = list(
+                filter(lambda trial: trial.status == "completed", exp.fetch_trials())
+            )
+            exp_trails = self._build_exp_trails(trials)
```
bouthilx, line 51: Looks like you are building list of best objectives found in submit time order. You could reuse the `regret` function here instead: https://github.com/Epistimio/orion/blob/develop/src/orion/analysis/regret.py#L14. I think working on a panda frame will make it easier for the work needed in `_display_plot`.

bouthilx: Also, I think it may be better if we extract the plotting functions and move them to the plotting backing. It would be great if we could generalize the regret function [here](https://github.com/Epistimio/orion/blob/develop/src/orion/plotting/backend_plotly.py#L139) to plot multiple experiments together.

src/orion/benchmark/assessment/averageresult.py
```diff
@@ -0,0 +1,94 @@
...
+    For each algorithm, run fixed number of Experiment, average the performance of trials
+    for the same algorithm at the same trial sequence order.
+    For the performance of trials in an Experiment, instead using the actual trial objective
+    value, here we use the best objective value in the same Experiment until the particular trial.
+    """
+
+    def __init__(self, task_num=1):
+        super(AverageResult, self).__init__(task_num=task_num)
+
+    def plot_figures(self, task, experiments):
```
bouthilx, line 32: Same comment applies here with respect to plotting and panda frame.

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,144 @@
...
+    algorithms = db_config["algorithms"]
+
+    obj_targets = []
+    str_targets = db_config["targets"]
+    for target in str_targets:
+        obj_target = {}
+
+        assessments = target["assess"]
+        obj_assessments = []
+        for assessment in assessments:
```
bouthilx, line 78: That's a lot of things to process that we should probably be able to avoid by using the Factory instead.

src/orion/benchmark/task/branin.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+import math
+
+import numpy
+
+from orion.benchmark.base import BaseTask
+
+
+class Branin(BaseTask):
+    """Branin function as benchmark task"""
```
bouthilx, line 20: Is there any citation we should use for this function?

bouthilx: Same comment applies to other tasks.

src/orion/benchmark/base.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        Return the black box function to optimize, the function will expect hyper-parameters to
+        search and return objective values of trial with the hyper-parameters.
+        """
+        pass
+
+    @property
+    def max_trials(self):
+        """Return the max number of trials to run for the"""
+        return self.trials_num
```
bouthilx, line 95: Why do we pass number of trials to the task? I think it's more related to the assessment.

tests/unittests/benchmark/test_assessments.py
```diff
@@ -0,0 +1,192 @@
...
+            "orion-benchmark-assessment-averagerank-AverageRank": {"task_num": 1}
+        }
+
+        ar2 = AverageRank(task_num=5)
+        assert ar2.task_num == 5
+        assert ar2.configuration == {
+            "orion-benchmark-assessment-averagerank-AverageRank": {"task_num": 5}
+        }
+
+    def test_plot_figures(self):
```
bouthilx, line 64: Another reason to move plotting to the plotting module is that it already covers many corner cases in the tests.

tests/unittests/benchmark/test_assessments.py
```diff
@@ -0,0 +1,192 @@
...
+        ):
+            plot = ar1.plot_figures("task_name", [(0, experiment)])
+
+        assert type(plot) is plotly.graph_objects.Figure
+
+    def test_figure_layout(self):
+        """Test assessment plot format"""
+        ar1 = AverageRank()
+
+        gen_exps = []
```
bouthilx, line 81: Could we move the code here for the creation of the experiments outside of this tests so that it can be reused for other assessment test of figure layout.

tests/unittests/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,387 @@
...
+
+            assert experiment is not None
+
+    def test_status(self, study, algorithms, generate_experiment_trials):
+        """Test to get status of a study"""
+        algo_num = len(algorithms)
+        task_num, gen_exps, gen_trials = generate_experiment_trials
+
+        with OrionState(experiments=gen_exps, trials=gen_trials):
+            experiments = []
```
bouthilx, line 316: Looks like the block below appears in many tests. I would be nice to take it out and reuse it.

src/orion/benchmark/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                )
+                print(table)
+
+        return benchmark_status
+
+    def analysis(self):
+        """Return all the assessment figures"""
+        figures = []
+        for study in self.studies:
+            figure = study.display()
```
bouthilx, line 123: The names are a bit confusing I think. Experiment has analysis, study has display and assessment has plot_figures. As I understand it they all are providing figures for the analysis? I would reuse the same name across the different class.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -0,0 +1,142 @@
...
+                        ],
+                    }
+                ],
+            }
+            assert bm1.configuration == cfg
+
+            assert bm1.configuration == bm2.configuration
+
+    def test_create_with_only_name(self):
+        """Test creation with a not existed benchmark name"""
```
bouthilx, line 89:
```suggestion
        """Test creation with a non-existing benchmark name"""
```

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -0,0 +1,142 @@
...
+                    algorithms=algorithms,
+                    targets=[
+                        {"assess": [DummyAssess], "task": [RosenBrock(25, dim=3)]}
+                    ],
+                )
+            print(str(exc.value))
+
+            assert "type object '{}' has no attribute ".format("DummyAssess") in str(
+                exc.value
+            )
```
bouthilx, line 142: The function `_resolve_db_config` looks tricky. I would test it carefully.

src/orion/benchmark/base.py
donglinjy: I guess we can. But abstract `__call__` seems a little bit weird for me. That's why I suggest some way else in the last comment.

src/orion/benchmark/base.py
donglinjy: yes, they should, they are actually already included. Only difference was it did not use kwargs directly which will be changed.

src/orion/benchmark/benchmark_client.py
donglinjy: yeah, and even for the config argument, maybe can also make it as a sharable config object (not just storage I think) and used at experiment/benchmark phase.

docs/src/user/benchmark.rst
donglinjy: have not tried this before, but how can be put a git repo link in the doc so it will not affected by the `branch`, or just always point to `develop`?

src/orion/benchmark/assessment/averagerank.py
donglinjy: Let me see if pandas can make it easy :)

At this moment, one reason for the existence of the assessment is to draw different charts for different assess purpose, if we do this some plot can be shared, we can put it into the some utils code.

src/orion/benchmark/benchmark_client.py
donglinjy: Can you be more specific about `Factory`? do you mean the `Factory design pattern` or similar stuff in the orion like below, or something else?
https://github.com/Epistimio/orion/blob/develop/src/orion/core/utils/__init__.py#L31

src/orion/benchmark/__init__.py
donglinjy: Good idea. I was trying to differentiate at the beginning as they are display at different levels, but seems make it more confusing.

src/orion/benchmark/base.py
bouthilx: I think it's reasonable to force implementation of ` __call__` in children. Do you find it weird because it's an operator function?

src/orion/benchmark/base.py
donglinjy: The current benchmark constructor interface allows one assessment pairs with multiple tasks,  and each task can have different max trials definition.  Assessment at this moment only decide how many times each task should run, and may be whether run in parallel or not, which is not included yet.

src/orion/benchmark/base.py
donglinjy: yeah, it is the python magic methods, for me, it is more like some reserved methods. 

Some possible reference like, TensorFlow is asking `call()` and Pytorch is asking `forward` in the sub class for similar purpose.

https://www.tensorflow.org/tutorials/customization/custom_layers#implementing_custom_layers
https://pytorch.org/tutorials/beginner/examples_nn/two_layer_net_module.html

src/orion/benchmark/assessment/averagerank.py
donglinjy: Not sure if it will have a known name like `Regret curve`.  My idea is that the assessment (or purpose) will be the plot name.

tests/unittests/benchmark/test_benchmark_client.py
donglinjy: by tricky,  do you mean the import module based on the string configure from db part?

src/orion/benchmark/base.py
bouthilx: Sounds good, let's go with `call()` as you suggested.

src/orion/benchmark/benchmark_client.py
bouthilx: What do you mean by sharable config object? Sharing across what?

docs/src/user/benchmark.rst
bouthilx: I would point to develop. At the time of release, we replace with master.

src/orion/benchmark/assessment/averagerank.py
bouthilx: Ok, but we should explain in a short (or couple of short) sentence(s) what the plot is a bout. A bit like we would describe a figure in a paper.

src/orion/benchmark/benchmark_client.py
bouthilx: Yeah I meant the Factory in Oríon, like it is being used for the algorithm: https://github.com/Epistimio/orion/blob/develop/src/orion/algo/base.py#L295

src/orion/benchmark/base.py
bouthilx: I see. In this regard I would say It seems more specific to assessment-task pairs than task specifically. Maybe defining assessment-task pairs would be over-engineering however. Since task class is very simple it may be best to keep it in for now then.

tests/unittests/benchmark/test_benchmark_client.py
bouthilx: Sorry the comment wasn't clear at all. I mean that the comprehension of the function requires lots of efforts and therefore bugs could easily be introduced without us noticing it if the function is only tested superficially.

src/orion/benchmark/assessment/averagerank.py
bouthilx: New version for reference:

```diff
- For each algorithm, run fixed number of Experiment, average the rank of trials for
- the same algorithm at the same trial sequence order.
- For the performance of trials in an Experiment, instead using the actual trial objective value,
- here we use the best objective value in the same Experiment until the particular trial.
+ Evaluate the average performance (objective value) between different search algorithms from
+ the rank perspective at different time steps (trial number).
+ The performance (objective value) used for a trial will the best result until the trial.
```

I think it is much more clear, I only had to read it once to understand. :) Thanks!

src/orion/benchmark/benchmark_client.py
donglinjy: some thing like this:
orion_config = {“storage": {xxx}, "max_broken": 5, "xxx": "yyy"}

create_benchmark(xxx..., config=orion_config)
create_experiment(xxx..., config=orion_config)

bouthilx: Oh ok. Yes. One way would be to set the values in orion.core.config, which is shared. We could make a helper function to help with this.

src/orion/core/utils/__init__.py
```diff
@@ -110,18 +110,16 @@ def __call__(cls, of_type, *args, **kwargs):
...
-        classes = {clz.__name__.lower(): clz for clz in cls.types}
-
-        if not classes.get(of_type.lower()):
+        if of_type.lower() not in cls.typenames:
             cls.types = list(get_all_subclasses(cls.__base__))
             cls.types = [
                 class_ for class_ in cls.types if class_.__name__ != cls.__name__
             ]
-            classes = {clz.__name__.lower(): clz for clz in cls.types}
+            cls.typenames = list(map(lambda x: x.__name__.lower(), cls.types))
```
bouthilx, line 118: Do you repeat this in __call__ to enable detection of newly created subclasses after creation of the factory? Seems fine, but I would factor out this method to avoid duplication with __init__. Maybe we should also extend cls.types instead of overwriting it.

bouthilx: Also we should add unittests to verify this new behavior.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -223,3 +223,96 @@ def test_create_with_invalid_targets(self, benchmark_algorithms):
...
+        }
+
+        with OrionState(benchmarks=cfg_task):
+            with pytest.raises(NotImplementedError) as exc:
+                get_or_create_benchmark("bm00002")
+            assert "Could not find implementation of BaseTask" in str(exc.value)
+
+    def test_create_with_not_exist_targets_parameters(self, benchmark_algorithms):
+        """Test creation with not existing assessment parameters"""
+        cfg = {
```
bouthilx, line 274: To avoid duplications I would define these configs at the top of the module and modify the small parts as necessary in each tests. To avoid side-effects you can either define them as pytest.fixture or do a deep.copy at the beginning of the test.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -223,3 +223,96 @@ def test_create_with_invalid_targets(self, benchmark_algorithms):
...
+                        "CarromTable": {"max_trials": 20},
+                    },
+                }
+            ],
+        }
+
+        with OrionState(benchmarks=cfg):
+            with pytest.raises(TypeError) as exc:
+                get_or_create_benchmark("bm00002")
+            assert "__init__() got an unexpected keyword argument 'idontexist'" in str(
```
bouthilx, line 294: Would the error tell what classes this __init__ is for?

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if not algorithms or not targets:
+        raise NoConfigurationError(
+            "Benchmark {} does not exist in DB, "
+            "algorithms and targets space was not defined.".format(name)
+        )
+
+    benchmark = _create_benchmark(name, algorithms, targets)
+
+    if benchmark_id is None:
```
bouthilx, line 68: Just to be sure I didn't miss anything. If we pass a benchmark config with a name, and the benchmark is already registered in DB but the passed config is different, the one from the config would have precedence and would be the one loaded and executed. Is that right? I think it would be important to check any differences and at least raise a warning in such case.

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,145 @@
...
+        # persist benchmark into db
+        _register_benchmark(benchmark)
```
bouthilx, line 70: In case of race conditions, this will raise a DuplicateKeyError and make the worker crash. One way to handle this is to retry recursively like we do here: https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/experiment_builder.py#L187-L198.

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,145 @@
...
+    study.setup_experiments()
+
+    return study
+
+
+def _fetch_benchmark(name):
+
+    if name:
+        configs = get_storage().fetch_benchmark({"name": name})
+    else:
```
bouthilx, line 131: Is there a use case for `_fetch_benchmark(None)`? I can't see any based on get_or_create_benchmark.

src/orion/benchmark/benchmark_client.py
```diff
@@ -0,0 +1,145 @@
...
+import datetime
+
+from orion.benchmark import Benchmark, Study
+from orion.benchmark.assessment.base import BenchmarkAssessment
+from orion.benchmark.task.base import BenchmarkTask
+from orion.core.utils.exceptions import NoConfigurationError
+from orion.storage.base import get_storage, setup_storage
+
+
+def get_or_create_benchmark(
```
bouthilx, line 21: We can polish this in another PR, but it would be great to have a consistent interface with the ExperimentClient. We now have [`build_experiment`](https://github.com/Epistimio/orion/blob/develop/src/orion/client/__init__.py#L44) to create executable experiments and [`get_experiment`](https://github.com/Epistimio/orion/blob/develop/src/orion/client/__init__.py#L246) to load a readable/writable experiment.

tests/unittests/benchmark/conftest.py
```diff
@@ -0,0 +1,98 @@
...
+        "end_time": None,
+        "heartbeat": None,
+        "results": [],
+        "params": [],
+    }
+    return trial_config
+
+
+@pytest.fixture
+def generate_experiment_trials(
```
bouthilx, line 74: I think this should be added to [`orion.testing`](https://github.com/Epistimio/orion/blob/develop/src/orion/testing/__init__.py).

tests/unittests/benchmark/test_assessments.py
```diff
@@ -0,0 +1,155 @@
...
+            _,
+            experiment,
+            _,
+        ):
+            plot = ar1.analysis("task_name", [(0, experiment)])
+
+        assert type(plot) is plotly.graph_objects.Figure
+
+    @pytest.mark.usefixtures("version_XYZ")
+    def test_figure_layout(
```
bouthilx, line 114: You could reuse `assert_regrets_plot` from the plotting unittests. This way if we make any changes we only need to adjust `assert_regrets_plot`.

tests/unittests/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,316 @@
...
+import orion.core.io.experiment_builder as experiment_builder
+from orion.benchmark import Benchmark, Study
+from orion.benchmark.assessment import AverageRank, AverageResult
+from orion.benchmark.task import CarromTable, RosenBrock
+from orion.client.experiment import ExperimentClient
+from orion.core.worker.experiment import Experiment
+from orion.testing import OrionState
+
+
+def build_study_experiments(algo_num, task_num):
```
bouthilx, line 17: This should also be in orion.testing.

tests/unittests/core/test_utils.py
```diff
@@ -49,3 +49,8 @@ class MyFactory(Base, metaclass=Factory):
     assert "Could not find implementation of Base, type = 'random'" in str(
         exc_info.value
     )
+
+    class Random(Base):
+        pass
+
+    assert type(MyFactory(of_type="random")) is Random
```
bouthilx, line 56: Oh, it is tested! Great :)

src/orion/core/utils/__init__.py
donglinjy: yes.  As the factory could be created at the very beginning when importing some modules before other subclasses are actually defined. Let me see how to make the part code be reusable.  As for overwriting or extending, I think it is the same here, because at the moment if some sub-classes defined before is not in cls.typenames after rewriting, means in the memory we can not find those sub-classes any more, and later creation will cause failure.

tests/unittests/benchmark/test_benchmark_client.py
donglinjy: no, there is not currently. Unless we catch the  `TypeError` in the Factory `__call__` and raise a more specific message.

src/orion/benchmark/benchmark_client.py
donglinjy: Yes, there is no branching support like experiment and the existing db config will be honored not what different algorithms or targets for the second call.  I think we can raise a generic warning and print the db config when we detect there is difference.

src/orion/benchmark/benchmark_client.py
donglinjy: I was planning `get_benchmarks ` interface for people to retrieve benchmark for just the status/view purpose, this interface will give `None` to `_fetch_benchmark` so all benchmarks will be retrieved. 

I can remove it but seems leave it won't make harm, and the `get all benchmarks` may be still necessary to added later.

src/orion/benchmark/benchmark_client.py
donglinjy: okay.  We can think about this a little bit later. I was actually trying to avoid the too much concepts like `experiment` has ( `we get experimentclient, experimentbulder, experimetview and experiment `). It may be necessary and understandable for `experiment` as it is the key code of Orion :)

tests/unittests/benchmark/test_assessments.py
donglinjy: do we need to move such `assert_regrets_plot` to some other places, or just import them directly in benchmark test code?

src/orion/core/utils/__init__.py
bouthilx: Yeah, if the code to create the list of types is factored out and reused then overwriting should not be an issue since it's doing the same thing.

tests/unittests/benchmark/test_benchmark_client.py
bouthilx: We can see the class name anyway in the full stack trace so it's fine.

bouthilx: We can see the class name anyway in the full stack trace so it's fine.

src/orion/benchmark/benchmark_client.py
bouthilx: Yeah we should not aim to support an equivalent of the EVC for the benchmarks. I agree with a generic warning and db config print.

src/orion/benchmark/benchmark_client.py
bouthilx: Sounds good.

src/orion/benchmark/benchmark_client.py
bouthilx: I agree there is a lot, too much, for experiment. Do you think `get_experiment/benchmark` and `build_experiment/benchmark` could be enough? As you say, `get` is for reading the states or perhaps minor db edits like updating trial status, and `build` would be for fully loading experiment/benchmark object in order to execute them. They main difference is that `get` only load the data while `build` needs to make sure we are running the same config.

tests/unittests/benchmark/test_assessments.py
bouthilx: Oh right, they cannot be imported as they are. You could move the asserts in `orion.testing.plotting.backend_plotly`.

tests/unittests/benchmark/test_benchmark.py
donglinjy: do we have a criterion for `orion.testing`, compared with what did with `conftest` directly in the test code?

bouthilx: We use conftest for fixtures and orion.testing for functions that are widely useful. orion.testing is also useful to reuse testing code in external plugins. Since the creation of the OrionState class to ease testing when using the database, we try to favor functions to populate the DB rather than fixtures. This way we can also parametrize the function that creates the DB content. It's more tedious to do with fixtures. For the rest, fixtures work well.

src/orion/benchmark/benchmark_client.py
donglinjy: Agree, we can polish and make it consistent later, when we add the get all benchmarks interface maybe :)

bouthilx: Sounds good :)

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -224,95 +209,42 @@ def test_create_with_invalid_targets(self, benchmark_algorithms):
...
-                        "AverageRank": {"task_num": 2},
-                    },
-                    "task": {
-                        "RosenBrock": {"dim": 3, "max_trials": 25},
-                        "CarromTable": {"max_trials": 20},
-                    },
-                }
-            ],
+
+        benchmark_config["targets"][0]["assess"]["AverageResult"] = {
```
bouthilx, line 207: It would be better to deepcopy here.

src/orion/benchmark/benchmark_client.py
```diff
@@ -65,9 +73,24 @@ def get_or_create_benchmark(
...
+
     if benchmark_id is None:
-        # persist benchmark into db
-        _register_benchmark(benchmark)
+        logger.debug("Benchmark not found in DB. Now attempting registration in DB.")
+        try:
+            _register_benchmark(benchmark)
+            logger.debug("Benchmark successfully registered in DB.")
+        except DuplicateKeyError:
+            print("exception....")
```
bouthilx, line 83: This `exception` print may be misleading and scary for users. I think we can remove it. The logging below could be `info` instead of `debug`.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -164,46 +129,54 @@ def test_create_with_only_name(self):
 
             assert "Benchmark {} does not exist in DB".format(name) in str(exc.value)
 
-    def test_create_with_invalid_algorithms(self):
+    def test_create_with_different_configure(self, benchmark_config_py):
```
bouthilx, line 132: The logging should be captured and tested with caplog to make sure the warning was logged.

src/orion/testing/__init__.py
```diff
@@ -71,6 +71,53 @@ def _generate(obj, *args, value):
     return new_trials
 
 
+def generate_experiments_trials(
```
bouthilx, line 74:
```suggestion
def generate_benchmark_experiments_trials(
```

src/orion/testing/__init__.py
```diff
@@ -71,6 +71,53 @@ def _generate(obj, *args, value):
...
+    benchmark_algorithms, experiment_config, trial_config, task_number, max_trial
+):
+    """Return a list of experiments and trials"""
```
bouthilx, line 77:
```suggestion
    """Return a list of experiments and trials for a benchmark"""
```

bouthilx (approved):
:tada: :smiley:

PR #539, opened 2021-01-20
Add grid search algorithm

### LINKED ISSUE #508: Add basic Grid Search

bouthilx:
[Fixes #508]

### DIFF
--- a/src/orion/algo/gridsearch.py
+++ b/src/orion/algo/gridsearch.py
@@ -32,8 +32,10 @@ def grid(dim, num):
         return fidelity_grid(dim, num)
     else:
         raise TypeError(
-            "Grid Search only supports `loguniform`, `uniform` and `choices`: "
-            "`{}`".format(dim.prior_name)
+            "Grid Search only supports `real`, `integer`, `categorical` and `fidelity`: "
+            f"`{dim.type}`\n"
+            "For more information on dimension types, see "
+            "https://orion.readthedocs.io/en/stable/user/searchspace.html"
         )
 
 
--- a/src/orion/algo/gridsearch.py
+++ b/src/orion/algo/gridsearch.py
@@ -1,0 +1,225 @@
+# -*- coding: utf-8 -*-
+"""
+:mod:`orion.algo.gridsearch` -- Grid Search
+===========================================
+
+.. module:: gridsearch
+    :platform: Unix
+    :synopsis: Grid Search Algorithm
+
+"""
+import itertools
+import logging
+
+import numpy
+
+from orion.algo.base import BaseAlgorithm
+from orion.algo.space import Categorical, Fidelity, Integer, Real
+
+log = logging.getLogger(__name__)
+
+
+def grid(dim, num):
+    """Build a one-dim grid of num points"""
+
+    if dim.type == "categorical":
+        return categorical_grid(dim, num)
+    elif dim.type == "integer":
+        return discrete_grid(dim, num)
+    elif dim.type == "real":
+        return real_grid(dim, num)
+    elif dim.type == "fidelity":
+        return fidelity_grid(dim, num)
+    else:
+        raise TypeError(
+            "Grid Search only supports `loguniform`, `uniform` and `choices`: "
+            "`{}`".format(dim.prior_name)
+        )
+
+
+def fidelity_grid(dim, num):
+    """Build fidelity grid, that is, only top value"""
+    return [dim.interval()[1]]
+
+
+def categorical_grid(dim, num):
+    """Build categorical grid, that is, all categories"""
+    if len(dim.categories) != num:
+        log.warning(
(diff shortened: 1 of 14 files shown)


src/orion/algo/gridsearch.py
```diff
@@ -0,0 +1,225 @@
...
+
+        while n_trials > max_trials:
+            coordinates = []
+            capped_values = []
+            for name, dim in space.items():
+                capped_value = max(n_values[name] - adjust, 1)
+                capped_values.append(capped_value)
+                coordinates.append(list(grid(dim, capped_value)))
+
+            if all(value <= 1 for value in capped_values):
```
donglinjy, line 159: looks like capped_value for all dimension will always be the same at this place.  So we want `raise` if any dimension `n_values` is adjusted to 1 or all the dimension is adjust?

tests/unittests/algo/test_gridsearch.py
```diff
@@ -0,0 +1,139 @@
...
+    space.register(dim1)
+    space.register(dim2)
+    space.register(dim3)
+    space.register(dim4)
+
+    with pytest.raises(ValueError) as exc:
+        GridSearch.build_grid(space, {k: 5 for k in space}, 10)
+
+    assert exc.match(
+        "Cannot build a grid smaller than 10. "
```
donglinjy, line 137: Same as above, for the space, there could be a grid with size `8`, where first 2 dimension with only one candidate. Is it valid?

src/orion/algo/gridsearch.py
```diff
@@ -0,0 +1,225 @@
...
+    if dim.prior_name.endswith("reciprocal"):
+        a, b = dim.interval()
+        return list(_log_grid(a, b, num))
+    elif dim.prior_name.endswith("uniform"):
+        a, b = dim.interval()
+        return list(_lin_grid(a, b, num))
+    else:
+        raise TypeError(
+            "Grid Search only supports `loguniform`, `uniform` and `choices`: "
+            "`{}`".format(dim.prior_name)
```
donglinjy, line 90: Seems same error message as the one inside the `def grid(dim, num):`

src/orion/algo/gridsearch.py
bouthilx: `n_values[name]` can vary from one dimension to another. We want to keep adjusting until grid is >= max_trials or until there is no more adjustment possible (all dims have only 1 point each). If this is still larger than max_trials, raise ValueError.

bouthilx: `n_values[name]` can vary from one dimension to another. We want to keep adjusting until grid is >= max_trials or until there is no more adjustment possible (all dims have only 1 point each). If this is still larger than max_trials, raise ValueError.

tests/unittests/algo/test_gridsearch.py
bouthilx: The source of the (intentional) problem is the 2 categorical dimensions with 5 choices each. The grid always picks all choices, no matter what `n_values` is. This leads to a smallest possible grid of 1 * 1 * 5 * 5 > 10.

src/orion/algo/gridsearch.py
bouthilx: Oh indeed, I should change the error message for grid(). Thanks!

tests/unittests/algo/test_gridsearch.py
donglinjy: oh, I made a mistake, thought `Categorical("dim3", "abcde")` had 2 values, but it had 5.

src/orion/algo/gridsearch.py
donglinjy: How about a space like below. if `max_trials` is 10, when `n_values` = 2 for all dims, the grid size is 16, but in a middle of adjustment such as `n_values[dim1]=1, n_values[dim2]=2, n_values[dim3]=2, n_values[dim4]=2`, the grid size is 8, which looks like a valid gride.

```
    dim1 = Real("dim1", "uniform", 0, 1)
    dim2 = Integer("dim2", "uniform", 0, 10)
    dim3 = Categorical("dim3", "ab")
    dim4 = Categorical("dim4", "cd")
```

bouthilx: Yeah we could do the adjustment in a more efficient way, adjusting by 1 for all incrementally is very basic. Lazy me is tempted to say current version is good enough however. :) Do you think leaving this improvement for later would be harmful for the usability of the algorithm? 

Or perhaps I could just add an `if` test in the for-loop to break earlier if we reach n_trials < max_trials.

donglinjy: I think current implementation is okay, just want to confirm this is what we want to do. :)

donglinjy (approved):

PR #567, opened 2021-03-03
Deterministic algorithm benchmark

donglinjy:
Why:
Some algorithm like `gridsearch` is `deterministic` when the space is defined, so it make no sense to same task multiple time for the average purpose. 

Fix:
For the input algorithms of benchmark, 
`get_or_create_benchmark(..., algorithms=[...], ...)`

Instead of keeping it exactly the same as experiment input
`algorithms=["random", {"gridsearch": {"n_values": 50}}]`

We change it as below so when some algorithm is `deterministic`, put `True` here.
```
algorithms = [
            {"algorithm": "random"},
            {"algorithm": {"gridsearch": {"n_values": 50}}, "deterministic": True},
        ]
```


src/orion/benchmark/__init__.py
```diff
@@ -21,7 +21,13 @@ class Benchmark:
     name: str
         Name of the benchmark
     algorithms: list, optional
-        Algorithms used for benchmark, each algorithm can be a string or dict.
+        Algorithms used for benchmark, each algorithm definition will be a dict with two keys.
+
+        algorithm: str or dict
```
bouthilx, line 26: Is `str` still supported? That would be preferable.

bouthilx: oh, so yes, but we need to pass [{'algorithm': 'algo'}] all the time. That seems very cumbersome. :worried:

bouthilx:
The main problem for having `deterministic` defined in the algorithm class is that Assessment/Study are not working on the instantiated algorithms so they cannot access it, is that it? If we could access at least the algo class based on the algorithm name we could access the class.deterministic property.

src/orion/benchmark/__init__.py
bouthilx: Yes, supporting str as well would be better.

bouthilx:
> Although, algorithm level deterministic may not be always accurate, a random seed could make random algorithm reproducible actually.

Good point! I did not think if this at all. So maybe defining if an algorithm is deterministic in the object itself is slightly more complicated.

Perhaps your solution is the simplest one. There is one issue however with supporting str and dict. We should support the dict in the form of `{'algo_name': config}` as well as `{'algorithms': {'algo_name': config}, 'deterministic': True}`.

bouthilx:
> Although, algorithm level deterministic may not be always accurate, a random seed could make random algorithm reproducible actually.

Good point! I did not think if this at all. So maybe defining if an algorithm is deterministic in the object itself is slightly more complicated.

Perhaps your solution is the simplest one. There is one issue however with supporting str and dict. We should support the dict in the form of `{'algo_name': config}` as well as `{'algorithms': {'algo_name': config}, 'deterministic': True}`.

donglinjy:
@bouthilx we now support all the algorithms input format.
- ["random"]
- ["random" :{"seed": 1}]
- [{"algorithm": "random"}]
- [{"algorithm": {"random" :{"seed": 1}}}]
- [{"algorithm": "random", "deterministic": True}]

docs/src/user/benchmark.rst
```diff
@@ -15,7 +15,7 @@ for how to create and :doc:`/code/benchmark` for how to use benchmark.
   from orion.benchmark.task import RosenBrock, EggHolder, CarromTable
 
   benchmark = get_or_create_benchmark(name='benchmark',
-          algorithms=['random', 'tpe'],
+          algorithms=[{'algorithm': 'random'}, {'algorithm': 'tpe'}],
```
bouthilx, line 18:
```suggestion
          algorithms=['random', 'tpe'],
```
I would use the simplest form that is supported for the documentation.

src/orion/benchmark/__init__.py
```diff
@@ -21,7 +21,18 @@ class Benchmark:
     name: str
         Name of the benchmark
     algorithms: list, optional
-        Algorithms used for benchmark, each algorithm can be a string or dict.
+        Algorithms used for benchmark, with supported formats:
```
bouthilx, line 24: I would give examples of lists here to make it more clear.

src/orion/benchmark/__init__.py
```diff
@@ -195,17 +206,66 @@ class Study:
...
+        algorithm: one algorithm in the `Study` algorithms list.
+        """
+
+        def __init__(self, algorithm):
+            parameters = None
+            deterministic = False
+
+            if isinstance(algorithm, dict):
+                if len(algorithm) > 1 or algorithm.get("algorithm"):
+                    print(algorithm)
```
bouthilx, line 232:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -195,17 +206,66 @@ class Study:
...
+                    deterministic = algorithm.get("deterministic", False)
+                    experiment_algorithm = algorithm["algorithm"]
+
+                    if isinstance(experiment_algorithm, dict):
+                        name, parameters = list(experiment_algorithm.items())[0]
+                    else:
+                        name = experiment_algorithm
+                else:
+                    print(list(algorithm.items())[0])
```
bouthilx, line 241:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -195,17 +206,66 @@ class Study:
...
+                    name, parameters = list(algorithm.items())[0]
+                    print(name, parameters)
```
bouthilx, line 243:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -195,17 +206,66 @@ class Study:
...
+            else:
+                print(algorithm)
```
bouthilx, line 245:
```suggestion
```

src/orion/benchmark/__init__.py
```diff
@@ -224,6 +291,13 @@ def setup_experiments(self):
 
             for algo_index, algorithm in enumerate(self.algorithms):
 
+                # deterministic = algorithm.get("deterministic", False)
```
bouthilx, line 294: Could remove the comments here.

src/orion/benchmark/__init__.py
```diff
@@ -294,13 +369,16 @@ def experiments(self):
 
     def __repr__(self):
         """Represent the object as a string."""
-        algorithms_list = list()
+        algorithms_list = [algorithm.name for algorithm in self.algorithms]
+        """
```
bouthilx, line 373: Could the comment below be removed?

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -169,10 +169,28 @@ def test_create_with_invalid_algorithms(self, benchmark_config_py):
...
+        with OrionState():
+            config = copy.deepcopy(benchmark_config_py)
+            config["algorithms"] = algorithms
+            bm = get_or_create_benchmark(**config)
+
+            for study in bm.studies:
+                for status in study.status():
+                    algo = status["algorithm"]
+                    if algo == "gridsearch":
+                        assert status["experiments"] == 1
```
bouthilx, line 192: I would add an else statement verifying that the number of experiments for other algorithms is 2

bouthilx:
Alright! There is only few minor things left and then it is ready for merge. Thanks!

bouthilx (approved):
LGTM, thanks! :)

PR #581, opened 2021-03-29
Fix for coverage issue

### LINKED ISSUE #569: Fix CodeCov
Code coverage is not correctly measured. In a last attempt to only measure coverage of src/orion (and not tests/), it seems that it now measures tests/ only. :disappointed:

abergeron:
# Description
Fix #569.  Now codecov should report coverage on all files within src/ and not report anything about the tests.

There may be some lingering reports about `tests/functional/backward_compatibility/test_version.py`, because this was the file that had coverage previously, but those should disappear with time.

# Changes
tox now installs the package in development mode, to allow coverage to keep track of files.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,6 +21,7 @@ PLATFORM =
 
 [testenv]
 description = Run tests with coverage with pytest under current Python env
+usedevelop = true
 setenv = COVERAGE_FILE=.coverage.{envname}
 passenv = CI
 deps =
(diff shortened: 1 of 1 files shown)


abergeron:
This should be ready for review now.

bouthilx:
That's all that was needed? :open_mouth: 

Is `--source=src` different than the option [here](https://github.com/Epistimio/orion/blob/develop/codecov.yml#L18)?

abergeron:
Yes.

`--source=src` is more or less the same thing, yes.  It might be optional.

bouthilx:
Ok so the only part required to solve the issue was `usedevelop = true` for tox?

bouthilx (approved):
Quite fast and simple. :)

abergeron:
I just flashed that there is one key difference between `--source=src` and `src` in the codecov config.

`--source=src` will report empty coverage on files which are not even imported, whereas `src` in the codecov config will just exclude all coverage reports that are not in the specified path.

bouthilx:
Oh, quite different! We should keep both then, especially --source=src.

PR #582, opened 2021-03-30
[WIP] Remove MongoDB dependency from core tests

breuleux:
# Description

Progress towards #571 

# Changes

The tests have been fixed in the following files:

- [x] tests/functional/branching/test_branching.py
- [x] tests/functional/client/test_cli_client.py
- [x] tests/functional/commands/test_db_rm.py
- [x] tests/functional/commands/test_db_set.py
- [x] tests/functional/commands/test_info_command.py
- [x] tests/functional/commands/test_insert_command.py
- [x] tests/functional/commands/test_list_command.py
- [x] tests/functional/commands/test_setup_command.py
- [x] tests/functional/commands/test_status_command.py
- [x] tests/functional/demo/test_demo.py
- [x] tests/functional/parsing/test_parsing_base.py
- [x] tests/functional/parsing/test_parsing_hunt.py
- [x] tests/functional/parsing/test_parsing_insert.py
- [x] tests/unittests/core/cli/test_checks.py
- [x] tests/unittests/core/evc/test_resolutions.py
- [x] tests/unittests/core/evc/test_conflicts.py
- [x] tests/unittests/core/worker/test_consumer.py
- [x] tests/unittests/core/worker/test_experiment.py
- [x] tests/unittests/core/worker/test_producer.py
- [x] tests/unittests/core/worker/test_trial.py
- [x] tests/unittests/core/worker/test_trial_pacemaker.py
- [x] tests/unittests/core/worker/test_trials_history.py
- [x] tests/unittests/core/test_branch_config.py
- [x] tests/unittests/core/test_insert.py

I added the `storage` and `pdatabase` fixtures in `conftest.py` to provide the empty PickledDB storage (since `setup_pickleddb_database` does not yield it).

* `test_conflicts.py::TestExperimentNameConflict.test_conflict_exp_w_child_as_parent` still fails, possibly because of `_id` (was a string in MongoDB, not in PickledDB).
* `test_producer.py` is giving me problems, namely `orion.core.utils.exceptions.NoConfigurationError: Experiment supernaedo2-dendi does not exist in DB and space was not defined`.

PR #585, opened 2021-03-30
Change the default database to PickledDB

### LINKED ISSUE #572: Set PickledDB as default Storage Backend
Note: `docs/src/install/database.rst` should be reworked based on new default backend. Running a grep for `mongodb` in docs would be a good idea also to make sure we don't miss any other docs that need an update.

To be clear, this task involves mostly work on the documentation. In the code there is probably only `src/orion/core/__init__.py` that needs to be modified.

abergeron:
# Description
Fixes #572 

# Checklist
## Tests

## Documentation

## Quality

### DIFF
--- a/docs/src/install/database.rst
+++ b/docs/src/install/database.rst
@@ -77,10 +77,10 @@ stage.
    $ orion db test
 
    Check for a configuration inside the default paths...
-       {'name': 'orion', 'type': 'pickleddb', 'host': '[...]/orion/orion_db.pkl', 'port': 27017}
+       {'name': 'orion', 'type': 'pickleddb', 'host': '', 'port': 27017}
    Success
    Check if configuration file has valid database configuration...
-       {'name': 'orion', 'type': 'pickleddb', 'host': '[...]/orion/orion_db.pkl', 'port': 27017}
+       {'name': 'orion', 'type': 'pickleddb', 'host': '', 'port': 27017}
    Success
 
    [...]
@@ -94,7 +94,7 @@ This is with MongoDB since there are more options to play with.
    $ orion db test --config local.yaml
 
    Check for a configuration inside the global paths...
-       {'name': 'test', 'type': 'pickleddb', 'host': '[...]/orion/orion_db.pkl', 'port': 27017}
+       {'name': 'test', 'type': 'pickleddb', 'host': '', 'port': 27017}
    Success
    Check if configuration file has valid database configuration...
        {'type': 'mongodb', 'host': 'localhost'}
@@ -111,7 +111,7 @@ that will be used and then prints the instance created to confirm the database t
 
    [...]
 
-   Using configuration: {'name': 'orion', 'type': 'pickleddb', 'host': '[...]/orion/orion_db.pkl', 'port': 27017}
+   Using configuration: {'name': 'orion', 'type': 'pickleddb', 'host': '', 'port': 27017}
    Check if database of specified type can be created... Success
    DB instance <orion.core.io.database.pickleddb.PickledDB object at 0x7f86d70067f0>
 
--- a/docs/src/install/database.rst
+++ b/docs/src/install/database.rst
@@ -28,9 +28,10 @@ This will create a yaml file of the following format.
 
    .. code-block:: yaml
 
-      database:
-        type: 'pickleddb'
-        host: '/path/to/a/file.pkl'
+      storage:
+        database:
+          type: 'pickleddb'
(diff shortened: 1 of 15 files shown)


abergeron:
It seems that the MongoDB setup failed, but I didn't change anything there.  I can't manually re-run the tests.

Should I push an empty commit to restart the tests?

abergeron:
This doesn't pass the tests currently apparently, but I would probably wait for #582 before I fix it.

docs/src/install/database.rst
```diff
@@ -79,7 +76,7 @@ stage.
    $ orion db test
 
    Check for a configuration inside the default paths...
-       {'type': 'mongodb', 'name': 'mydb', 'host': 'localhost'}
+       {'type': 'pickleddb', 'name': 'orion', 'host': 'orion_db.pkl'}
```
bouthilx, line 79: You can drop `'name'` for the pickleddb.

docs/src/install/database.rst
```diff
@@ -101,15 +98,15 @@ Alternatively, here's an example including all three configuration methods.
    $ orion db test --config local.yaml
```
bouthilx, line 98: Here the text should be adjusted. Perhaps make it an example using mongodb, because there isn't much options to set on pickleddb.

docs/src/user/config.rst
```diff
@@ -88,10 +88,10 @@ Full Example of Global Configuration
 .. code-block:: yaml
 
     database:
-        host: localhost
+        host: orion_db.pkl
         name: orion
```
bouthilx, line 92: name and port can be dropped.

docs/src/user/config.rst
```diff
@@ -174,7 +174,7 @@ host
 ~~~~
 
 :Type: str
-:Default: 127.0.1.1
+:Default: orion_db.pkl
```
bouthilx, line 177: The default is https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/database/pickleddb.py#L25

src/orion/core/__init__.py
```diff
@@ -117,7 +112,7 @@ def define_database_config(config):
     database_config.add_option(
         "host",
         option_type=str,
-        default=default_host,
+        default="orion_db.pkl",
```
bouthilx, line 115: Default should be https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/database/pickleddb.py#L25

docs/src/install/database.rst
abergeron: Actually I would add 'port' since it would match what orion outputs.

I agree that they are irrelevant for pickleddb, but it might confuse new users to see unexpected output.

docs/src/user/config.rst
abergeron: Maybe but trying to import that file in the config function leads to an import loop and fails.

src/orion/core/__init__.py
abergeron: Maybe but trying to import that file in the config function leads to an import loop and fails.

abergeron: But I found out that I can just reach into DIRS locally and use that to build the default path so that works.

abergeron:
I've made the changes locally, but I'm waiting on the merge of #582 to push them since I've rebased this branch on top of that.

src/orion/core/__init__.py
bouthilx: Actually you can forget about this default. I prefer Olivier's idea to set it to None by default and leave it to the DB class to set it to DEFAULT if it is None. Could you please do this in this PR?

abergeron: Ok

abergeron: It actually doesn't work and sets the default db to the string "None" if I put `None` there.

bouthilx: Yes that's what I meant by 'leave it to the DB class to set it to DEFAULT if it is None. Could you please do this in this PR?'

You would need to adjust the code here: https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/database/pickleddb.py#L100

abergeron: The 'None' string was due to Configuration, but I've adjusted to make the default an empty string and the database classes will replace that with their built-in defaults.

bouthilx:
@abergeron The PR was merged. This is ready for rebase.

bouthilx (approved):

Issue #583, opened 2021-03-30
Python API hangs when experiment leaves without releasing the trials.


bouthilx:
When a trial is reserved with ExpelimentClient.reverse(trial) (or within ExperimentClient.suggest), it must be released otherwise Oríon will hang. Ideally, the released trial should have its status set to broken during `atexit` as implemented [here](https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L34), but for some reason this part of the code seems to hang. We would need to find out why and fix this.

bouthilx:
Fixed by #587

PR #587, opened 2021-03-31
[WIP] Add context-manager-enabled Trial subclass.

obilaniu:
# Description
There are hangs in the tests when `release(trial)` is not invoked on a reserved trial (#583).

After discussion, it is agreed that the `atexit()` handling strategy should be abandoned in favour of turning `Trial` into a context manager, usable with the Python `with`-statement. We begin doing so here.

## Tests

## Documentation

## Quality

### DIFF
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -533,10 +533,11 @@ class ExperimentClient:
 
             raise e
 
-        if trial is not None:
+        if trial is None:
+            return trial
+        else:
             self._maintain_reservation(trial)
-
-        return None if trial is None else TrialCM(self, trial)
+            return TrialCM(self, trial)
 
     def observe(self, trial, results):
         """Observe trial results
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -23,7 +23,7 @@ from orion.core.utils.exceptions import (
     WaitingForTrials,
 )
 from orion.core.utils.flatten import flatten, unflatten
-from orion.core.worker.trial import Trial
+from orion.core.worker.trial import Trial, TrialCM
 from orion.core.worker.trial_pacemaker import TrialPacemaker
 from orion.plotting.base import PlotAccessor
 from orion.storage.base import FailedUpdate
@@ -536,7 +536,7 @@ class ExperimentClient:
         if trial is not None:
             self._maintain_reservation(trial)
 
-        return trial
+        return None if trial is None else TrialCM(self, trial)
 
     def observe(self, trial, results):
         """Observe trial results
--- a/src/orion/core/worker/trial.py
+++ b/src/orion/core/worker/trial.py
@@ -445,3 +445,34 @@ class Trial:
         return hashlib.md5(
             (params + experiment_repr + lie_repr).encode("utf-8")
         ).hexdigest()
+
+
+class TrialCM:
+    __slots__ = ("_cm_experiment", "_cm_trial")
+
+    def __init__(self, experiment, trial):
+        self._cm_experiment = experiment
+        self._cm_trial = trial
+
+    def __getattribute__(self, name):
+        if name in {"_cm_experiment", "_cm_trial", "__enter__", "__exit__"}:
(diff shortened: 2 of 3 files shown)


src/orion/client/experiment.py
```diff
@@ -528,7 +528,7 @@ def suggest(self):
         if trial is not None:
             self._maintain_reservation(trial)
 
-        return trial
+        return None if trial is None else TrialCM(self, trial)
```
bouthilx, line 531: This could be merged with the if above.

src/orion/core/worker/trial.py
```diff
@@ -445,3 +445,34 @@ def compute_trial_hash(
...
+            object.__setattr__(self, name, value)
+
+    def __enter__(self):
+        return self._cm_trial
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            if exc_type is None:
+                self._cm_experiment.release(self._cm_trial)
+            else:
```
bouthilx, line 475: Not for any exceptions, KeyboardInterrupt should be set to 'interrupted'.

tests/unittests/client/test_experiment_client.py
```diff
@@ -476,10 +476,12 @@ class TestBroken:
     def test_broken_trial(self):
         """Test that broken trials are detected"""
         with create_experiment(config, base_trial) as (cfg, experiment, client):
-            trial = client.suggest()
-            assert trial.status == "reserved"
-
-            atexit._run_exitfuncs()
```
bouthilx, line 482: All other atexit in the tests need to be removed and replace with the with syntax.

bouthilx:
We will need additional tests. 

1. If trial is set to completed with experiment.observe() and then there is a raise within the `with`-block, then the status should stay at observe, not be set to broken.
2. Same thing if trial is released in the `with`-block and there is a raise before leaving it.
3. TrialCM must be tested to make sure attributes can be accessed and set properly, including to_dict()

src/orion/core/worker/trial.py
```diff
@@ -445,3 +445,36 @@ def compute_trial_hash(
...
+        if name not in {"_cm_experiment", "_cm_trial"}:
+            setattr(self._cm_trial, name, value)
+        else:
+            object.__setattr__(self, name, value)
+
+    def __enter__(self):
+        return self._cm_trial
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
```
bouthilx, line 472: If the trial status is not `reserved` locally, the release may _corrupt_ the database. For instance if `experiment.observe(trial)` is called before the release, if will change status from 'completed' to 'broken'. To be clear, this is not due to this new implementation but rather was an existing bug that I just found out.

I would suggest adding `was` argument to `set_trial_status` in `storage` ([here for example](https://github.com/Epistimio/orion/blob/develop/src/orion/storage/base.py#L330) to specify what is supposed to be the current status. By default it would be set to None and would use trial.status in the `where` query as is currently the case.

In `ExperimentClient.release`, `set_trial_status(trial, status)` would be changed to 
`set_trial_status(trial, status, was='reserved')`

This would require no changes from this PR implementation of `__exit__`

src/orion/core/worker/trial.py
```diff
@@ -470,9 +470,11 @@ def __enter__(self):
...
-            if exc_type is None:
-                self._cm_experiment.release(self._cm_trial)
-            else:
+            if exc_type is KeyboardInterrupt:
+                self._cm_experiment.release(self._cm_trial, "interrupted")
+            elif exc_type is not None:
                 self._cm_experiment.release(self._cm_trial, "broken")
+            else:
+                self._cm_experiment.release(self._cm_trial)
         except:
```
bouthilx, line 479: I think it would be better to only silence runtime errors here. It may be good to log a warning when catching Runtime errors as well. I say this error in particular because it is the only likely error that may occur because of status issues during `__exit__` (https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L448).

bouthilx (approved):
LGTM thanks! :)

Issue #574, opened 2021-03-17
Add `orion plot` command line.


bouthilx:
The plots can only be rendered using the python API. This is especially cumbersome if the experiments were run with the command line API. The `orion plot` command line would be very convenient even when running experiments with the python API.

PR #589, opened 2021-04-01
for the "orion plot" subcommand

gyom:
For [orion plot](https://github.com/Epistimio/orion/issues/574). This has been tested manually and documented, but no tests written.


src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+def add_subparser(parser):
+    """Add the subparser that needs to be used for this command"""
+    plot_parser = parser.add_parser("plot", help=DESCRIPTION, description=DESCRIPTION)
+
+    cli.get_basic_args_group(plot_parser)
+
+    plot_parser.add_argument(
+        "kind",
+        type=str,
+        help="kind of plot to generate. "
```
bouthilx, line 36: It would be good to move [this]( https://github.com/Epistimio/orion/blob/develop/tests/unittests/plotting/test_plot_accessor.py) inside orion/plotting/base.py and pass this list here to argument `choices`.

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+        " Pick one among ['lpi', 'partial_dependencies', 'parallel_coordinates', 'regret']"
+    )
+
+    plot_parser.add_argument(
+        "-t",
+        "--type",
+        type=str,
+        default="png",
+        help="type of plot to return. "
```
bouthilx, line 45: Same here, it would be good to create a list of choices so that argparse can take care of validating the input.

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+        " Will override the `type` argument."
+        " (default is {exp.name}_{kind}.{type})"
+    )
+
+    plot_parser.add_argument(
+        "--scale",
+        type=float,
+        default=None,
+        help="more pixels, but same proportions of the plot. "
+        " Reference is 1.0."
```
bouthilx, line 65: I guess greater than 1 means more high definition?

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    plot_parser.set_defaults(func=main)
+
+    return plot_parser
+
+
+def main(args):
+    """Starts an application that will generate a plot."""
+    config = experiment_builder.get_cmd_config(args)
```
bouthilx, line 76: You don't need this line

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Note : If you specify no argument at all, the default behavior
+    #        is to plot the lpi as <experiment_name>.png .
+
+
+    # What does this do?    
+    #experiment_builder.setup_storage(config.get("storage"))
```
bouthilx, line 83: This would create the storage object, but it is dealt with at line 85 anyway, so no need for line  83.

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+    
+    experiment = experiment_builder.build_from_args(args)
+
+    assert args['kind'] in ['lpi', 'partial_dependencies', 'parallel_coordinates', 'regret']
```
bouthilx, line 87: You could drop this when passing choices to argparse

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+    func_plotting = getattr(orion.plotting.base, args['kind'])
+
+    output_plot = func_plotting(experiment)
```
bouthilx, line 90:
```suggestion
    output_plot = experiment.plot(kind=args['kind'])
```

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+
+    valid_types = ['png', 'jpg', 'jpeg', 'webp', 'svg', 'pdf', 'html', 'json']
+    # 'png', 'jpg', 'jpeg', 'webp', 'svg', 'pdf' handled by fig.write_image
+    # 'html' handled by fig.write_html
+    # 'json' handled by just writing out the file
+
+    if args['output']:
+        # `output` was given, and we'll override `type`
+        args['type'] = args['output'].split(".")[-1]
+        assert args['type'] in valid_types, (
```
bouthilx, line 100: This also could be dropped by passing valid choices to argparse.

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+            "Overriding the `type` field with something from `output`, "
+            f"but we got the invalid value : {args['type']} .")
+    else:
+        # `type` was given, and we'll pick a `output` based on the name
+        assert args['type'] in valid_types, (
```
bouthilx, line 105: Idem

src/orion/core/cli/plot.py
```diff
@@ -0,0 +1,125 @@
...
+            f"Invalid `type` {args['type']} .")
+        # Using `experiment.name` instead of args['name'] because it leaves
+        # orion the possibility of having inferred the name in other ways.
+        args['output'] = f"{experiment.name}_{args['kind']}.{args['type']}"
```
bouthilx, line 109:
```suggestion
        args['output'] = f"{experiment.name}-v{experiment.version}_{args['kind']}.{args['type']}"
```
Sorry I forgot about this when we discussed. We should use the version as well.

src/orion/core/cli/plot.py
gyom: Yes. I think it's natural to think that scale 2.0 means you double the pixels. I didn't think for a moment that it could be ambiguous, but I guess there's some ambiguity. Do you think it needs clarification?

src/orion/core/cli/plot.py
gyom: Indeed. I will remove it.

src/orion/core/cli/plot.py
gyom: Indeed. I will remove it.

src/orion/core/cli/plot.py
gyom: Good point. I didn't know that was an option. I'll do it for both 'type' and 'kind'.

src/orion/core/cli/plot.py
bouthilx: Maybe not. :)

src/orion/core/cli/plot.py
gyom: Done.

bouthilx:
Continued in #593

PR #591, opened 2021-04-06
Handle cases when trial.status is not reserved anymore locally.

bouthilx:
Why:

Some trial status updates require the trial to have a specific status
(like releasing the trial reservation). In such case, `set_trial_status`
should not verify that status in database in current trial.status, but
rather that the trial status should be a precise one.

How:

Add argument `was` which will override current trial status if given.

### DIFF
--- a/src/orion/storage/legacy.py
+++ b/src/orion/storage/legacy.py
@@ -294,10 +294,8 @@ class Legacy(BaseStorageProtocol):
 
     def set_trial_status(self, trial, status, heartbeat=None, was=None):
         """See :func:`orion.storage.base.BaseStorageProtocol.set_trial_status`"""
-        if heartbeat is None:
-            heartbeat = datetime.datetime.utcnow()
-        if was is None:
-            was = trial.status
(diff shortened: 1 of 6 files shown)


src/orion/storage/legacy.py
```diff
@@ -292,18 +292,19 @@ def push_trial_results(self, trial):
...
 
-    def set_trial_status(self, trial, status, heartbeat=None):
+    def set_trial_status(self, trial, status, heartbeat=None, was=None):
         """See :func:`orion.storage.base.BaseStorageProtocol.set_trial_status`"""
         if heartbeat is None:
             heartbeat = datetime.datetime.utcnow()
-
-        update = dict(status=status, heartbeat=heartbeat, experiment=trial.experiment)
+        if was is None:
+            was = trial.status
```
obilaniu, line 300: `was = was or trial.status`

obilaniu: Pas litéralement équivalent (ça convertit aussi `False`, `[]`, `{}` et `""`) mais très Pythonique.

obilaniu (approved):

Issue #568, opened 2021-03-08
Parallel (asynchronous) trial execution at single worker


donglinjy:
`orion hunt` at one worker will consume trials in sequential and block until one trial is done.

With python api, `experimentclient workon` will run function in the same way, although `experimentclient suggest/observe` can be used separately and leaves user to decide how the trial can be executed, either in sequential or parallel.

PR #601, opened 2021-04-20
Orion python api in parallel

donglinjy:
# Description
Support Orion to run in parallel at local or remote cluster (dask, ray, etc.) through `joblib` for python interfaces.

fyi https://github.com/Epistimio/orion/issues/568

# Usage
~~Workon function not support yet:~~

Experiment client workon method:
```
experiment = create_experiment(
   name='foo4',
   space=dict(x='uniform(-50,50)'))

experiment.workon(foo, max_trials=5, n_jobs=2)

Or 

with joblib.parallel_backend('ray'):
    experiment.workon(foo, max_trials=5, n_jobs=2)

```

# Checklist

## Tests

## Documentation

## Quality


src/orion/client/experiment.py
```diff
@@ -617,19 +622,26 @@ def workon(self, fct, max_trials=infinity, **kwargs):
...
-                return trials
-            kwargs.update(flatten(trial.params))
-            results = fct(**unflatten(kwargs))
-            self.observe(trial, results=results)
-            trials += 1
-
-        return trials
+        with Parallel(n_jobs=n_jobs, verbose=100) as parallel:
+            trials = parallel(
+                delayed(self._optimize)(fct, kwargs) for _ in range(max_trials)
```
bouthilx, line 627: I think we could avoid lots of overhead by creating `n_jobs` instead and optimizing sequentially in each. 

```suggestion
                delayed(self._optimize)(fct, kwargs) for _ in range(n_jobs)
```

src/orion/client/experiment.py
```diff
@@ -617,19 +622,26 @@ def workon(self, fct, max_trials=infinity, **kwargs):
...
+        # this is required for process-based or remote backend
+        setup_storage(storage=self.storage_config)
+
+        trial = self.suggest()
+        if trial is None:
+            log.warning("Algorithm could not sample new points")
+            return 0
+        kwargs.update(flatten(trial.params))
+        results = fct(**unflatten(kwargs))
+        self.observe(trial, results=results)
```
bouthilx, line 643: Following comment above, turning this into a sequential optimization (basically moving code from `workon` to here, with small modifications)
```suggestion
        trials = 0
        trials_done = len(self.fetch_completed_trials())
        while not self.is_done and  len(self.fetch_completed_trials()) - trials_done < max_trials:
            trial = self.suggest()
            if trial is None:
                log.warning("Algorithm could not sample new points")
                return trials
            kwargs.update(flatten(trial.params))
            results = fct(**unflatten(kwargs))
            self.observe(trial, results=results)
            trials += 1
        return trials
```

bouthilx, line 643: Following comment above, turning this into a sequential optimization (basically moving code from `workon` to here, with small modifications)
```suggestion
        trials = 0
        trials_done = len(self.fetch_completed_trials())
        while not self.is_done and  len(self.fetch_completed_trials()) - trials_done < max_trials:
            trial = self.suggest()
            if trial is None:
                log.warning("Algorithm could not sample new points")
                return trials
            kwargs.update(flatten(trial.params))
            results = fct(**unflatten(kwargs))
            self.observe(trial, results=results)
            trials += 1
        return trials
```

src/orion/client/__init__.py
```diff
@@ -279,7 +279,14 @@ def get_experiment(name, version=None, mode="r", storage=None):
 
 
 def workon(
```
bouthilx, line 281: Since the parallelism is implemented in ExpelimentClient.workon I would avoid modifying this function for now to keep things simple. We can focus on ExpelimentClient.workon for now.

src/orion/core/worker/experiment.py
```diff
@@ -118,10 +117,13 @@ def __init__(self, name, version=None, mode="r"):
         self.algorithms = None
         self.working_dir = None
         self.producer = {}
-        self._storage = get_storage()
```
bouthilx, line 121: Is this to avoid pickling issues?

donglinjy: yes

src/orion/client/__init__.py
donglinjy: do you mean no need to care the parallelism support for this function, and leave ExpelimentClient.workon always with n_jobs=1 from here?

src/orion/client/experiment.py
donglinjy: Seems a good idea. But then maybe instead of `n_jobs`, `n_workers` is more accurate.

bouthilx: Oh yes, I totally agree!

src/orion/client/__init__.py
bouthilx: Yeah, for now.

src/orion/core/worker/experiment.py
donglinjy: With more testing, simply making `_storage` as property method can introduce issues:
- For the `workon` function, since it calls `update_singletons` inside it, make `get_storage()` outside the call will have different storage instance and leading the failure of some method calls (related to db of course) of the returned `ExperimentClient`.
- Similarly, although the `ExperimentClient.workon` does not `update_singletons` , but orion api provides such options, if user build `ExperimentClient` with one storage configure, call its `workon` method, and then `update_singletons`, will make user fail if they simply call `ExperimentClient` db query related apis.

I made some change as below, it may be able to solve above 2 issues, but will introduce a new issue, if user call `ExperimentClient.workon` with `n_workers > 2` first time, it will be okay, but if after this first call, user call `ExperimentClient.workon` db related apis, such as `fetch_trials`, and then call `ExperimentClient.workon` with `n_workers > 2` again, there will be un-picklable issue since ExperimentClient will now have a storage instance which may not un-picklable, like mongodb.
```
        self._istorage = None

    @property
    def _storage(self):
        if not self._istorage:
            self._istorage = get_storage()
        return self._istorage
```

To solve this, I am now thinking to **clear** the storage instance property before `joblib parallel` and **reset** it back after that inside `ExperimentClient.workon`. But it is a little bit tricky, especially when there is error between the **clear** and **reset**. What do you think?

bouthilx: Could we implement instead a method to pickled the experiment which ignores the storage? We could implement [__getstate__](https://docs.python.org/3/library/pickle.html#object.__getstate__) for instance.

src/orion/client/experiment.py
```diff
@@ -622,26 +621,34 @@ def workon(self, fct, max_trials=infinity, n_jobs=None, **kwargs):
...
-        trial = self.suggest()
-        if trial is None:
-            log.warning("Algorithm could not sample new points")
-            return 0
-        kwargs.update(flatten(trial.params))
-        results = fct(**unflatten(kwargs))
-        self.observe(trial, results=results)
-        return 1
+        trials = 0
+        while not self.is_done and trials < max_trials:
```
bouthilx, line 643: `trials < max_trials` should never be reached right? If `max_trials` is smaller than `exp.max_trials`, than `exp.max_trials` is set to `max_trials`. So here `max_trials` can only be equal or larger than `exp.max_trials`. Since `exp.is_done` depends on `num_trials < exp.max_trials`, then `exp.is_done` will always be True whenever `trials < max_trials` is True.

tests/unittests/client/test_experiment_client.py
```diff
@@ -969,3 +974,36 @@ def foo(a, b):
             assert len(params)
             assert "x" in params["a"]
             assert "y" in params["b"]
+
+    def test_parallel(self, monkeypatch):
```
bouthilx, line 978: It's good to have a test with a mocked call like this, but I think there should also be a test where the optimization is actually done to make sure race conditions are handled properly.

src/orion/core/worker/experiment.py
donglinjy: This can be an option. Let me give it a try.

tests/unittests/client/test_client.py
```diff
@@ -503,6 +503,30 @@ def foo(x):
...
+    def test_workon_with_parallel_backend(self):
+        """Test there is no impact of joblib parallel for workon function"""
+
+        def foo(x):
+            return [dict(name="result", type="objective", value=x * 2)]
+
+        import joblib
+
+        with joblib.parallel_backend("loky"):
+            experiment = workon(
```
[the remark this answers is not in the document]
donglinjy: no, it does not, the test case is just ensure that there is no impact/error when user does this.

src/orion/client/experiment.py
donglinjy: It is suppose to be, and this is the same code as the old loop.  `self.is_done` should already cover `trials < max_trials`.

bouthilx: Yeah the difference now is that exp.max_trial is set to `max_trial` if `max_trial < exp.max_trial`. So I think we could remove `trials < max_trials` in the while clause.

src/orion/client/experiment.py
```diff
@@ -617,9 +622,28 @@ def workon(self, fct, max_trials=infinity, **kwargs):
...
+
+        with Parallel(n_jobs=n_workers) as parallel:
+            trials = parallel(
+                delayed(self._optimize)(fct, max_trials, **kwargs)
+                for _ in range(n_workers)
+            )
+
+        return sum(trials)
+
+    # pylint: disable=protected-access
```
bouthilx, line 638:
```suggestion
```

src/orion/client/experiment.py
```diff
@@ -617,9 +622,28 @@ def workon(self, fct, max_trials=infinity, **kwargs):
...
+    def _optimize(self, fct, max_trials, **kwargs):
+        # this is required for process-based or remote backend
+        setup_storage(storage=self.storage_config)
+        self._experiment._storage = get_storage()
```
bouthilx, line 642:
```suggestion
        self._experiment._storage = get_storage()  # pylint: disable=protected-access
```

src/orion/client/experiment.py
```diff
@@ -617,9 +622,28 @@ def workon(self, fct, max_trials=infinity, **kwargs):
         """
         self._check_if_executable()
 
+        # reduce chance of race condition for trials creation
```
bouthilx, line 625:
```suggestion
        # Use worker's max_trials inside `exp.is_done` to reduce chance of race condition for trials creation
```

tests/functional/algos/test_algos.py
```diff
@@ -247,3 +247,64 @@ def test_with_evc(algorithm):
...
         assert param.name == "x"
         assert param.type == "real"
+
+
+@pytest.mark.parametrize(
+    "algorithm", algorithm_configs.values(), ids=list(algorithm_configs.keys())
+)
+def test_parallel_workers(algorithm, tmp_path):
+    """Test parallel execution with joblib"""
+    pickle_host = str(tmp_path / "db.pkl")
```
bouthilx, line 257: It would be better to use OrionState to make sure we handle cleanup properly. Using `with OrionState(options...) as storage_state:` you can access storage config with `storage_state.storage_config` and pass it to `create_experiment`.

tests/functional/algos/test_algos.py
```diff
@@ -247,3 +247,64 @@ def test_with_evc(algorithm):
...
+
+    exp.workon(rosenbrock, max_trials=30, n_workers=3)
+
+    assert exp.configuration["algorithms"] == algorithm
+
+    trials = exp.fetch_trials(with_evc_tree=False)
+    assert len(trials) >= 20
+
+    trials_with_evc = exp.fetch_trials(with_evc_tree=True)
+    assert len(trials_with_evc) >= 30
```
bouthilx, line 288:
```suggestion
    assert 30 <=len(trials_with_evc) <= 30 + N_WORKERS
```

tests/functional/algos/test_algos.py
```diff
@@ -247,3 +247,64 @@ def test_with_evc(algorithm):
...
+    storage = {"type": "legacy", "database": {"type": "PickledDB", "host": pickle_host}}
+    base_exp = create_experiment(
+        name=name,
+        space=space_with_fidelity,
+        algorithms=algorithm_configs["random"],
+        storage=storage,
+    )
+    base_exp.workon(rosenbrock, max_trials=10)
+
+    exp = create_experiment(
```
bouthilx, line 270: We can avoid tests with evc when testing multiple workers. Unless you see a potential issue with EVC that is specific to workon with n_workers > 1?

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -63,6 +63,20 @@ def test_create_benchmark_no_storage(self, benchmark_config_py):
             assert isinstance(storage._db, PickledDB)
             assert storage._db.host == host
 
+    def test_create_benchmark_with_storage(self, tmp_path, benchmark_config_py):
+        """Test benchmark instance has the storage configurations"""
+        conf_file = str(tmp_path / "db.pkl")
+
+        config = copy.deepcopy(benchmark_config_py)
```
bouthilx, line 70: It would be best to use OrionState here as well.

tests/unittests/client/test_experiment_client.py
```diff
@@ -969,3 +974,35 @@ def foo(a, b):
...
+
+        with OrionState(
+            experiments=[config], trials=generate_trials(base_trial, [], config)
+        ) as cfg:
+
+            experiment = experiment_builder.build(name=config["name"])
+            if cfg.trials:
+                experiment._id = cfg.trials[0]["experiment"]
+            client = ExperimentClient(
+                experiment, Producer(experiment), storage=_get_default_test_storage()
```
bouthilx, line 996:
```suggestion
                experiment, Producer(experiment), storage=cfg.storage_config
```

tests/unittests/client/test_experiment_client.py
```diff
@@ -969,3 +974,35 @@ def foo(a, b):
...
+
+        def foo(x):
+            return [dict(name="result", type="objective", value=x * 2)]
+
+        def optimize(*args, **kwargs):
+            optimize.count += 1
+            return 1
+
+        with OrionState(
+            experiments=[config], trials=generate_trials(base_trial, [], config)
```
bouthilx, line 989: Why do you pass trials? Could this test be done with no trials?

tests/unittests/client/test_experiment_client.py
```diff
@@ -969,3 +974,35 @@ def foo(a, b):
...
+        ) as cfg:
+
+            experiment = experiment_builder.build(name=config["name"])
+            if cfg.trials:
+                experiment._id = cfg.trials[0]["experiment"]
```
bouthilx, line 994: Why do you need to override the `id`?

tests/unittests/client/test_experiment_client.py
bouthilx: Maybe you should use something like this instead?
```
        with create_experiment(
            exp_config=config, trial_config={}, statuses=[]
        ) as (cfg, experiment, client):
```
So that your client is already created.

tests/functional/algos/test_algos.py
donglinjy: You know what, I actually did not notice the issue of EVC at the beginning, I did this just because I think with or without n_workers >1, the behavior should be the same. But just when I re-ran the case, it failed, the root cause is in last commit, I only restore the storage for the current experiment node after de-serialize the object, but did not restore the storage for parent nodes, this is really weird. Any way, after I fixed the issue, it works.

tests/unittests/client/test_experiment_client.py
donglinjy: Did not use this before because need set storage for `experimentclient` explicitly at the beginning for `mulitple-process`, but later I changed to use 'mulitple-threads', this is good enough though.

tests/unittests/client/test_experiment_client.py
donglinjy: Legacy code, forgot to remove :)

tests/functional/algos/test_algos.py
donglinjy: Change to verify the `completed_trials` as `trials_with_evc` including all the state trials.

tests/functional/algos/test_algos.py
donglinjy: I tried this, but it did not work when put the code into `with OrionState`. After fixed the evc issue this time, still fail when putting code into `with OrionState(storage=storage):`, only the first random algo succeeds, all the other algorithms will fail because some un-consistent storage issue in `OrionState context` and in the `_optimize method` although they have exactly the same storage configure.

donglinjy: It will good if you can also give it a try. Or maybe we can take a look of this potential issue later.

donglinjy: It will be good if you can also give it a look/try of this behavior Or maybe we can take a look of this potential issue later? Seems it should not be a functional issue.

src/orion/core/worker/experiment.py
```diff
@@ -150,6 +150,17 @@ def __setstate__(self, state):
         for key, value in state.items():
             setattr(self, key, value)
 
+    def _restore_storage(self):
+        self._storage = get_storage()
```
bouthilx, line 154: I think you should add a comment here explaining why we need to do this. (the weird evc issue with pickling)

tests/functional/algos/test_algos.py
bouthilx: Hmm, I would be concerned that users stumble on a similar issue. I would be in favor of investigating the issue and fixing it inside this PR.

tests/functional/algos/test_algos.py
bouthilx: haha I am glad this test caught it! EVC will never stop to surprise us. :sweat_smile: This was probably due to the serialization of the nodes in the EVC tree. Dropping the other nodes would have done the job if I am right. But your solution looks simpler. :)

bouthilx: So we should keep the tests with EVC indeed.

tests/unittests/client/test_experiment_client.py
bouthilx: And for the trials, could you pass no trials?

donglinjy: There is no trials pass-in now I think.

bouthilx: Indeed!

tests/functional/algos/test_algos.py
bouthilx: I think this is the only issue left. Would you like me to take a look?

donglinjy: It will be helpful if you can take a look. thanks

bouthilx: I found the issue. The problem is that storage_config is None if no custom configuration for the storage is passed. The worker will end up creating a storage based on global configuration, but the local worker may already have a storage initiated with a custom configuration, only this configuration was not passed to `build_experiment`. I think it may be easier to get around this issue by making the DB picklable instead of changing the experiment. I will give it a try.

bouthilx: I made a PR on your branch with the fix: https://github.com/donglinjy/orion/pull/3

bouthilx:
@donglinjy I think it is ready for merge. Do you agree?

PR #603, opened 2021-04-26
Add preliminary Warm-Start support

lebrice:
# Description
Adds a MultiTaskAlgo wrapper, that allows the 'warm-starting' of the algorithm, when it isn't explicitly warm-starteable.
Adds the AbstractKnowledgeBase class as an interface for the KnowledgeBase which is currently not part of Orion.

# Checklist

## Tests

## Documentation

## Quality

# Further comments


src/orion/algo/base.py
```diff
@@ -105,6 +105,7 @@ def __init__(self, space, **kwargs):
             kwargs,
         )
         self._trials_info = {}  # Stores Unique Trial -> Result
+        self._warm_start_trials = {}  # Stores unique warm-star trials and their results
```
bouthilx, line 108:
```suggestion
        self._warm_start_trials = {}  # Stores unique warm-start trials and their results
```

src/orion/algo/random.py
```diff
@@ -64,3 +64,16 @@ def suggest(self, num):
...
 
         return points
+
+
+    @contextmanager
+    def warm_start_mode(self):
+        # NOTE: The Random algorithm just discards any points observed while in
+        # 'warm-start mode':
+        backup = self._trials_info
+        self._trials_info = self._trials_info.copy()
```
bouthilx, line 74: We could avoid the copy here since they won't be used anyway.

src/orion/algo/base.py
```diff
@@ -319,12 +320,30 @@ def has_observed(self, point):
...
+
+        The idea behing this is that we don't want the algorithm to modify its state the
+        same way it would if it were observing regular points. For example, the number
+        of "used" trials shouldn't increase, etc.
+
+        New Algorithms or Algo wrappers can implement this method to control how the
+        state of the algo is affected by observing trials from other tasks than the
+        current (target) task.
+        """
+        yield
```
bouthilx, line 345: Should it raise NotImplementedError or running an algorithm with no implementation here could work? Could the default implementation be the temporary swapping of trials_info?

src/orion/benchmark/assessment/warm_start_efficiency.py
```diff
@@ -0,0 +1,253 @@
+from orion.benchmark.assessment.base import BaseAssess
```
bouthilx, line 1: `$ tox -e run-isort` :)

src/orion/benchmark/assessment/warm_start_efficiency.py
```diff
@@ -0,0 +1,253 @@
...
+from typing import Callable
+import pandas as pd
+import orion.analysis
+
+WarmStartExperimentsTuple = Tuple[ExperimentClient, ExperimentClient, ExperimentClient]
+
+
+class WarmStartEfficiency(BaseAssess):
+    """
+    TODO: Evaluate the 'warm start efficiency' (objective value) for each search algorithm
```
bouthilx, line 23: That's for another assessment object? Should be in an issue for another PR I guess?

bouthilx: Well, actually, would you mind moving the warm-starting benchmark stuff to another PR so that we can first focus on warm-starting itself?

src/orion/client/__init__.py
```diff
@@ -55,6 +55,7 @@ def build_experiment(
     heartbeat=None,
     working_dir=None,
     debug=False,
+    knowledge_base: "KnowledgeBase" = None,
```
bouthilx, line 58:
```suggestion
    knowledge_base = None,
```
Unless the rest in typed. :grin:

bouthilx: Also, need to be added in the docstring.

src/orion/client/experiment.py
```diff
@@ -581,6 +581,14 @@ def observe(self, trial, results):
 
         """
         self._check_if_executable()
+        if not isinstance(results, list):
+            results = [results]
+        adapted_results = []
+        for result in results:
```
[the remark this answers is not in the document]
bouthilx: Also, need to be tested in tests/unittests/client/test_experiment.py

src/orion/core/io/experiment_builder.py
```diff
@@ -521,8 +529,17 @@ def _instantiate_algo(space, max_trials, config=None, ignore_unavailable=False):
         config = orion.core.config.experiment.algorithms
 
     try:
-        algo = PrimaryAlgo(space, config)
-        algo.algorithm.max_trials = max_trials
+        if knowledge_base is not None:
+            from orion.core.worker.multi_task_algo import MultiTaskAlgo
```
bouthilx, line 533: Can be moved to top of module or there is a circular import problem?

src/orion/core/io/experiment_builder.py
```diff
@@ -521,8 +529,17 @@ def _instantiate_algo(space, max_trials, config=None, ignore_unavailable=False):
...
+            algo = MultiTaskAlgo(
+                space=space, algorithm_config=config, knowledge_base=knowledge_base
+            )
+        else:
+            algo = PrimaryAlgo(space, config)
+        algo.unwrapped.max_trials = max_trials
+        # from orion.core.worker.multi_task_algo import MultiTaskAlgo
```
bouthilx, line 540: Can be removed?

src/orion/core/utils/backward.py
```diff
@@ -141,7 +141,10 @@ def get_algo_requirements(algorithm):
         )
 
         return requirements
-
+    if isinstance(algorithm, str):
```
bouthilx, line 144: This should not happen. If there is an issue with the algo plugin discoverability it would need to be fixed at the level of the factory. This fix should be removed so that we can see again the error and fix it in Factory.

src/orion/core/worker/experiment.py
```diff
@@ -282,7 +282,7 @@ def update_completed_trial(self, trial, results_file=None):
         trial.end_time = datetime.datetime.utcnow()
         self._storage.retrieve_result(trial)
         # push trial results updates the entire trial status included
-        log.info("Completed trials with results: %s", trial.results)
+        log.debug("Completed trials with results: %s", trial.results)
```
bouthilx, line 285: Why debug instead of info?

src/orion/core/worker/algo_wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        the wrapped algorithm's.
+        """
+        return self.algorithm._trials_info  # type: ignore
+
+    @_trials_info.setter
+    def _trials_info(self, value: Dict[str, Trial]) -> None:
+        if self.algorithm is None:
+            # this is being set as part of '__init__' of the base class: ignore.
+            # This __trials_info_set attribute should be False.
+            assert not getattr(self, "__trials_info_set", False)
```
bouthilx, line 188: Should be a `if <>: raise RuntimeError`

src/orion/core/worker/algo_wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        NOTE: Adding this property just to make sure that if an algo wrapper doesn't
+        override a BaseAlgorithm method (for example when new methods / properties get
+        added to BaseAlgorithm but aren't yet overridden in AlgoWrapper), then things
+        don't accidentally get changed in the Wrapper's `_trials_info` rather than in
+        the wrapped algorithm's.
+        """
+        return self.algorithm._trials_info  # type: ignore
+
+    @_trials_info.setter
+    def _trials_info(self, value: Dict[str, Trial]) -> None:
```
bouthilx, line 184: What do we assume here for the construction of the algorithm? Algo is built and assigned to self.algorithm, and then the trials_info is created and assigned to `self.unwrapped`? Isn't the `trials_info` created and assigned during the exection of root's algo `__init__`?

src/orion/core/worker/knowledge_base.py
```diff
@@ -0,0 +1,73 @@
+""" Interface for the Knowledge Base, which is not currently in the Orion codebase.
```
bouthilx, line 1: That will need to be changed before we merge. :)

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+from orion.core.worker.primary_algo import PrimaryAlgo
+from orion.core.worker.trial import Trial
+from orion.storage.base import Storage
+
+try:
+    from functools import singledispatchmethod
+except ImportError:
+    from singledispatchmethod import singledispatchmethod
+
+from .algo_wrapper import AlgoWrapper, Point, Results
```
bouthilx, line 27: We avoid local imports if possible to make it more explicit.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+log = getLogger(__file__)
+
+from .knowledge_base import AbstractKnowledgeBase
```
bouthilx, line 32: Same here.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+import pkg_resources
+from pkg_resources import EntryPoint
+from orion.algo.base import OptimizationAlgorithm
+from functools import lru_cache
+from functools import singledispatch
+
+
+@lru_cache()
+def get_all_algos() -> Dict[str, Type[OptimizationAlgorithm]]:
```
bouthilx, line 74: We should not need this.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+def _(algo: AlgoWrapper) -> bool:
+    # NOTE: Not directing directly to algo.unwrapped, because there might eventually be
+    # a Wrapper that makes an algo WarmStarteable. Recurse down the chain of wrappers
+    # instead.
+    return is_warmstarteable(algo.algorithm)
+
+
+@is_warmstarteable.register(str)
+def _(algo: str) -> bool:
+    available_algos = get_all_algos()
```
bouthilx, line 131: Ah, that's why you implemented get_all_algos. This function should be part of the Factory in that case.

bouthilx: In any case, we should not spend time on this as the Factory will be soon removed in favor of the concept class. Let's wait for this part. The PR won't be merge before the Concept class is released in v0.2.0.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        knowledge_base: AbstractKnowledgeBase,
+    ):
+        log.debug(
+            f"Creating a MultiTaskAlgo wrapper: space {space}, algo config: {algorithm_config}"
+        )
+        self.knowledge_base = knowledge_base
+        self._enabled: bool
+        if is_warmstarteable(algorithm_config):
+            log.debug(
+                "Chosen algorithm is warm-starteable, disabling this multi-task wrapper."
```
bouthilx, line 166: Oh just noticed here. I believe the correct spelling is _warm-startable_. It should be fixed everywhere else as well.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        else:
+            log.debug(
+                "Chosen algorithm is NOT warm-starteable, enabling this multi-task wrapper."
+            )
+            self._enabled = True
+
+        # Figure out the number of potential tasks using the KB.
+        # TODO: The current Experiment might already be registered in the KB?
+        if knowledge_base.n_stored_experiments < 1:
+            # TODO: Should probably turn the 'warm-start' portion of this 'wrapper' off,
```
bouthilx, line 181: I think it is fine with just the warning. If there is an efficiency error the user can take out the knowledge base from the config. Less coding and less testing on our part. :)

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+            # because it will probably reduce the effectiveness of the algo to have to
+            # optimize with an uninformative 'task_id' dimension.
+            warnings.warn(
+                RuntimeWarning(
+                    f"Making the Algo 'multi-task', even though there are only"
+                    f"{knowledge_base.n_stored_experiments} experiments in the "
+                    f"Knowledge base. This might reduce the efficiency of the algo."
+                )
+            )
+        # TODO: Should we have this dimension be larger than the current number of
```
bouthilx, line 191: We could not know how much larger it should be. I think this is fine for now while we do warm-starting at the beginning only. When we work on on-line multi-task we will need to revise this anyway. I think you can remove the todo without creating an issue for this.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        # experiments in the KB, in case some get added in the future?
+        max_task_id = self.knowledge_base.n_stored_experiments + 1
+        task_label_dimension = Categorical(
+            "task_id", list(range(0, max_task_id)), default_value=0,
+        )
+        space_without_task_ids = Space(space.copy())
+        space_with_task_ids = Space(space.copy())
+        space_with_task_ids.register(task_label_dimension)
+        assert "task_id" in space_with_task_ids
```
bouthilx, line 200: This should be tested in unittests instead.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        assert "task_id" not in space_without_task_ids
+
+        self.algorithm = PrimaryAlgo(
+            space=space_with_task_ids, algorithm_config=algorithm_config
+        )
+        self._space = space_without_task_ids
+        self.current_task_id = 0
+        assert "task_id" not in self._space
```
bouthilx, line 208: Same here

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+                    f"Making the Algo 'multi-task', even though there are only"
+                    f"{knowledge_base.n_stored_experiments} experiments in the "
+                    f"Knowledge base. This might reduce the efficiency of the algo."
+                )
+            )
+        # TODO: Should we have this dimension be larger than the current number of
+        # experiments in the KB, in case some get added in the future?
+        max_task_id = self.knowledge_base.n_stored_experiments + 1
+        task_label_dimension = Categorical(
+            "task_id", list(range(0, max_task_id)), default_value=0,
```
bouthilx, line 195: This implies that no hyperparameter may be named `task_id`. I think it's fine but it should be clearly stated in the main docstring of the class.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        space_without_task_ids = Space(space.copy())
+        space_with_task_ids = Space(space.copy())
+        space_with_task_ids.register(task_label_dimension)
+        assert "task_id" in space_with_task_ids
+        assert "task_id" not in space_without_task_ids
+
+        self.algorithm = PrimaryAlgo(
+            space=space_with_task_ids, algorithm_config=algorithm_config
+        )
+        self._space = space_without_task_ids
```
bouthilx, line 206: I would be in favor of giving a more explicit name otherwise it becomes quite confusing.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        self.current_task_id = 0
+        assert "task_id" not in self._space
+        assert "task_id" not in self.space
+        # TODO: For some optimizers this exact key won't be found, there is a View(OneHotEncode(...))
```
bouthilx, line 210: You mean for algos that transforms the categorical dims into one-hots?

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        # assert "task_id" in self.algorithm.space
+        self.transformed_space = self.space
+
+    def suggest(self, num: int = 1) -> List[Point]:
+        """Suggest a `num` of new sets of parameters.
+
+        :param num: how many sets to be suggested.
+
+        .. note:: New parameters must be compliant with the problem's domain
+           `orion.algo.space.Space`.
```
bouthilx, line 221: including the tasks dimension?

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        """
+        points = self.algorithm.suggest(num=num)
+        if not self._enabled:
+            return points
+
+        # Remove the task ids from these points:
+        # NOTE: These predicted task ids aren't currently used.
```
bouthilx, line 228: This note should be added to the method's docstring.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        points_without_task_ids = [self._remove_task_id(point) for point in points]
+        for point in points_without_task_ids:
+            if point not in self.space:
+                raise ValueError(
+                    textwrap.dedent(
+                        f"""\
+                        Point is not contained in space:
+                        Point: {point}
+                        Space: {self.space}
+                        """
```
bouthilx, line 239: Can use

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        points = self.algorithm.suggest(num=num)
+        if not self._enabled:
+            return points
+
+        # Remove the task ids from these points:
+        # NOTE: These predicted task ids aren't currently used.
+
+        points_without_task_ids = [self._remove_task_id(point) for point in points]
+        for point in points_without_task_ids:
+            if point not in self.space:
```
bouthilx, line 232: Should be in `self._space` no? This point does not contain the task id.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        space.
+
+        .. seealso:: `orion.algo.base.BaseAlgorithm.observe`
+        """
+        # Add the task ids to the points, because we assume that the points passed to
+        # `observe` are from the current task.
+        if self._enabled:
+            # Add the task ids to the points.
+            points = [
+                self._add_task_id(point, self.current_task_id) for point in points
```
bouthilx, line 257: This method _add_task_id is not implemented by default? Should this method also ignore the passed task_id if the point already contains it? I assume observe will be called with points containing the task_id during warm-starting.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        take advantage of these points.
+
+        Parameters
+        ----------
+        warm_start_trials : Dict[Mapping, List[Trial]]
+            Dictionary mapping from ExperimentInfo objects (dataclasses containing the
+            experiment config) to the list of Trials associated with that experiment.
+        """
+        # Being asked to warm-start the algorithm.
+        if is_warmstarteable(self.algorithm):
```
bouthilx, line 277: It's a bit weird to have this function defined externally. I understand it is to support calling it with `str`, but maybe we could fix this once we get rid of the factor. A note for myself in a week or 2. :)

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        for i, (experiment_info, trials) in enumerate(warm_start_trials.items()):
+            # exp_space = experiment_info.space
+            # from warmstart.utils.api_config import hparam_class_from_orion_space_dict
+            # exp_hparam_class = hparam_class_from_orion_space_dict(exp_space)
+            log.debug(f"Experiment {experiment_info.name} has {len(trials)} trials.")
+            n_incompatible_trials = 0
+
+            for trial in trials:
+                try:
+                    # TODO: Use something smarter than `trial_to_tuple`, since it
```
bouthilx, line 302: I think the filtering and adjustment (if needed) should be done outside of the algorithm. Filtering could be done by the knowledge base query, and adjustment could be done by something like the recycler you implemented for the warm-starting project or like the EVC adapters.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+                f"total, out of which {n_compatible_trials} were found to be "
+                f"compatible with the target experiment."
+            )
+
+        if not compatible_trials:
+            log.info("No compatible trials detected.")
+            return
+
+        # Only keep trials that are new.
+        # TODO: Check that the task dim doesnt interfere with this:
```
bouthilx, line 342: Shouldn't be warm-started only once? Why would we check for that?

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+                    if infer_trial_id(point) not in self.unwrapped._warm_start_trials
+                ]
+            )
+        )
+        if not new_trials:
+            log.info("No new new warm-starting trials detected.")
+            return
+
+        with self.algorithm.warm_start_mode():
+            new_points_with_task_ids = [
```
bouthilx, line 359: Perhaps there should be a method to convert lists of points as well.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+                self._add_task_id(point, task_id)
+                for point, task_id in zip(new_points, new_task_ids)
+            ]
+            new_trials_with_task_ids = [
```
bouthilx, line 363: Before release v0.2.0 we will drop points and use trials instead in observe/suggest, so here we could only use trials and forget about points.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+                self._add_task_id(trial, task_id)
+                for trial, task_id in zip(new_trials, new_task_ids)
+            ]
+            # for new_trial, new_trial_with_task_id in zip(new_trials, new_trials_with_task_ids):
+            #     assert False, (new_trial, new_trial_with_task_id)
+
+            log.info(f"About to observe {len(new_points)} new warm-starting points!")
+            # TODO: Not quite sure I understand what the 'results' is supposed to be!
+            # results = [trial.objective for trial in new_trials]
```
bouthilx, line 372: It's formatted with `orion.core.utils.format_trials.get_trial_results()`.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+    def score(self, point: Union[Tuple, np.ndarray]) -> float:
+        """Allow algorithm to evaluate `point` based on a prediction about
+        this parameter set's performance. Return a subjective measure of expected
+        performance.
+
+        By default, return the same score for any parameter (no preference).
+        """
+        assert point in self.space
+        if self._enabled:
+            point = self._add_task_id(point, self.current_task_id)
```
bouthilx, line 388: Maybe self._add_task_id could take care of `if self._enabled`?

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+    @property
+    def n_suggested(self):
+        """Number of trials suggested by the algorithm **in the target task**"""
+        return len(self.trials_from_target_task)
+
+    @property
+    def n_observed(self):
+        """Number of completed trials observed by the algorithm"""
+        return len(self.trials_from_target_task)
```
bouthilx, line 413: It should depend on the objective of the trials. Are they completed?

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+    def set_state(self, state_dict):
+        """Reset the state of the algorithm based on the given state_dict
+
+        :param state_dict: Dictionary representing state of an algorithm
+        """
+        trials_info = state_dict.get("_trials_info")
+        self.unwrapped._trials_info = trials_info
```
bouthilx, line 421: There can be more than `trials_info`. We could delegate just doing `self.algorithm.set_state(state_dict['wrapped'])`.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+    @property
+    def trials_from_target_task(self) -> Dict[str, Tuple[Tuple[float, ...], Results]]:
+        return {
+            k: (point, results)
+            for k, (point, results) in self.unwrapped._trials_info.items()
+            if self.get_task_id(point) == 0
```
bouthilx, line 428: Could be a good idea to set a global current_task_id instead of hardcoding it to 0 everywhere.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+    @property
+    def is_done(self):
+        """Return True, if an algorithm holds that there can be no further improvement.
+        By default, the cardinality of the specified search space will be used to check
+        if all possible sets of parameters has been tried.
+        """
+        if not self._enabled:
+            return self.algorithm.is_done
+        # NOTE: DEBUGGING:
```
bouthilx, line 447: The part below should not be reimplemented because this will differ from one algorithm to another. There should be a way to access the trials of current task only in the root algorithm so that is_done is still functional when using warm-starting.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+            return True
+
+        max_trials = getattr(self, "max_trials", float("inf"))
+        if trials_from_target_task >= max_trials:
+            return True
+
+        return False
+
+    @contextmanager
+    def in_task(self, task_id: int):
```
bouthilx, line 472: I think we could avoid implementing this.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        """ Contextmanager that temporarily changes the value of `self.current_task_id`
+        to `task_id` and restores the original value after exiting the with block.
+        """
+        previous_task_id = self.current_task_id
+        self.current_task_id = task_id
+        yield
+        self.current_task_id = previous_task_id
+
+    @singledispatchmethod
+    def _add_task_id(self, point: Any, task_id: int) -> Any:
```
bouthilx, line 482: We can use trial version only as points will be removed shortly.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+    @_add_task_id.register
+    def _add_task_id_to_trial(self, trial: Trial, task_id: int) -> Trial:
+        # NOTE: Need to do a bit of gymnastics here because the points are expected
+        # to be tuples. In order to add the task id at the right index, we convert
+        # them like so:
+        # Trial -> dict (.params) -> dict (add task_id) -> Trial.
+        params = trial.params.copy()
+        params["task_id"] = task_id
+        # TODO: Should we copy over the status and everything else?
+        trial_with_task_id = dict_to_trial(params, self.algorithm.space)
```
bouthilx, line 506: But we lose all other info in the task while doing so. An alternative would be

```python
trial_with_task_id = copy.deepcopy(trial)
task_id_index = list(space.keys()).index('task_id')
trial_with_task_id._params.insert(task_id_index, Trial.Param(name='task_id', value=value, type='categorical'))
```

Note that this operation is done on the original trial, without any transformation. Transformations should be applied after the task id is added (somewhere outside of _add_task_id_to_trial).

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        for attribute in trial.__slots__:
+            value = getattr(trial, attribute)
+            if "params" not in attribute:
+                setattr(trial_with_task_id, attribute, value)
+        return trial_with_task_id
+
+    @singledispatchmethod
+    def _remove_task_id(self, point: Any) -> Any:
+        raise NotImplementedError(point)
```
bouthilx, line 515: Same here, we could keep the trial version only.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        trial = tuple_to_trial(point, self.algorithm.space)
+        trial_without_task_id = self._remove_task_id(trial)
+        point_without_task_id = trial_to_tuple(trial_without_task_id, self.space)
+        return point_without_task_id
+
+    @_remove_task_id.register(Trial)
+    def _remove_task_id_from_trial(self, trial: Trial) -> Trial:
+        params = trial.params.copy()
+        _ = params.pop("task_id")
+        trial_without_task_id = dict_to_trial(params, self.space)
```
bouthilx, line 528: Same here, we lose all trial info. Alternative
```python
trial_with_task_id = copy.deepcopy(trial)
task_id_index = list(space.keys()).index('task_id')
del trial_with_task_id._params[task_id_index]
```

bouthilx: Again, this is on the original trial, not the transformed one.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+        return trial_without_task_id
+
+    @singledispatchmethod
+    def get_task_id(self, point: Any) -> int:
+        raise NotImplementedError(point)
```
bouthilx, line 533: Idem, trial version only.

src/orion/core/worker/multi_task_algo.py
```diff
@@ -0,0 +1,551 @@
...
+
+    @get_task_id.register
+    def get_task_id_from_trial(self, trial: Trial) -> int:
+        return trial.params["task_id"] != 0
```
bouthilx, line 537: It returns a bool?

src/orion/core/worker/primary_algo.py
```diff
@@ -152,22 +153,6 @@ def judge(self, point, measurements):
             self.transformed_space.transform(point), measurements
         )
 
-    @property
```
bouthilx, line 155: Why are they removed?

src/orion/core/worker/producer.py
```diff
@@ -17,6 +18,7 @@
 from orion.core.utils.exceptions import SampleTimeout, WaitingForTrials
 from orion.core.worker.trial import Trial
 from orion.core.worker.trials_history import TrialsHistory
+# from orion.core.worker.knowledge_base import AbstractKnowledgeBase
```
bouthilx, line 21: Can be removed?

src/orion/algo/random.py
lebrice: Not exactly: One of either `backup` or `self._trials_info` needs to be copied.

src/orion/algo/base.py
lebrice: Good question! It's better IMO to keep this a no-op as default. I think it's fine to assume that an algo can work with the trials from other tasks too.
The swapping of `trials_info` only happens for the Random algorithm, not for the other algorithms.

src/orion/benchmark/assessment/warm_start_efficiency.py
lebrice: Good idea, however I wouldn't know how to create the PR for the warm-start benchmark while this one isn't integrated, since it depends on it. Should I create that PR against this branch now? or against develop after this is merged?

src/orion/core/worker/knowledge_base.py
lebrice: Why though? Shouldn't adding the concrete implementation be done in another PR?

src/orion/core/worker/multi_task_algo.py
lebrice: Yes exactly. Sorry there are quite a few straggling todos here I need to take a look at.

src/orion/core/worker/primary_algo.py
lebrice: I think it's because this implementation was moved to AlgoWrapper. I'll double-check.

PR #605, opened 2021-05-04
Refactor `hunt` command to reuse `ExperimentClient.workon`

bouthilx:
### Why

The workon pipeline was duplicated for the cli and python interface. The
python interface should be reused instead under the hood for the CLI
command.

### How

The consumer has been modified to serve only as a callable function to
execute the users script.
- The handling of interruption is moved to ExperimentClient.workon.
- The fetching of results file is moved from storage to consumer since
it is only relevant for the CLI.
- The pacemaker is removed. The heartbeat is only defined and used
within the ExperimentClient.

Experiment.suggest was incoherent when experiment is done. It would
return None, but the context manager could not work in such case. To
handle this, suggest now raises CompletedExperiment instead of returning
None.


src/orion/storage/legacy.py
```diff
@@ -196,39 +195,25 @@ def register_lie(self, trial):
         """See :func:`orion.storage.base.BaseStorageProtocol.register_lie`"""
         return self._db.write("lying_trials", trial.to_dict())
 
-    def retrieve_result(self, trial, results_file=None, **kwargs):
-        """Parse the results file that was generated by the trial process.
+    def retrieve_result(self, trial, **kwargs):
```
donglinjy, line 198: seems there is almost no need of this `method`?

src/orion/client/experiment.py
```diff
@@ -589,7 +614,15 @@ def observe(self, trial, results):
...
-    def workon(self, fct, max_trials=infinity, **kwargs):
+    def workon(
+        self,
+        fct,
+        max_trials=None,
+        max_broken=None,
+        trial_arg=None,
+        on_error=None,
+        **kwargs,
+    ):
```
donglinjy, line 625: Missing description for new parameters?

src/orion/client/experiment.py
```diff
@@ -609,24 +642,70 @@ def workon(self, fct, max_trials=infinity, **kwargs):
...
         kwargs = flatten(kwargs)
-        while not self.is_done and trials < max_trials:
-            trial = self.suggest()
-            if trial is None:
-                log.warning("Algorithm could not sample new points")
-                return trials
-            kwargs.update(flatten(trial.params))
-            results = fct(**unflatten(kwargs))
-            self.observe(trial, results=results)
+        while not self.is_done and trials - worker_broken_trials < max_trials:
```
donglinjy, line 674: for single worker, `trials - worker_broken_trials < max_trials` seems always make `self.is_done` as `false` only if `worker_broken_trials ` > 0?

donglinjy:
Looks pretty good to me. One thing left should be to merge the `experimentclient workon` code in PR https://github.com/Epistimio/orion/pull/601

src/orion/client/experiment.py
bouthilx: Good catch, thanks!

src/orion/client/experiment.py
bouthilx: I'm not sure I understand. Do you mean that if `trials - worker_broken_trials < max_trials` is true then automatically `self.is_done` cannot be true?

Here `max_trials` and `max_broken` can be different then those for the experiment.

src/orion/storage/legacy.py
bouthilx: Almost. Track may use it to update the results of the trial by fetching its db.

src/orion/client/experiment.py
donglinjy: Forgot this,  I took it wrong, I saw `trials - worker_broken_trials` as `trials + worker_broken_trials`. :)

bouthilx: haha ok :)

bouthilx:
Waiting for #601 before merging.

bouthilx:
@donglinjy I added argument `max_trials_per_worker` in commit d9c200d.

donglinjy:
> @donglinjy I added argument `max_trials_per_worker` in commit [d9c200d](https://github.com/Epistimio/orion/commit/d9c200d8b055506359824b10ea55dfd0e4db8536).

Looks reasonable since the `workon` become the entry for multiple workers.

donglinjy (approved):

PR #557, opened 2021-02-17
Documentation for the visualizations

bouthilx:
[Solve #527 #495]

The documentation for this branch is automatically built and available at https://orion.readthedocs.io/en/doc-viz/

See visualization examples at https://orion.readthedocs.io/en/doc-viz/auto_examples/index.html
And tutorials at https://orion.readthedocs.io/en/doc-viz/auto_tutorials/index.html

## Work in progress
(checked when done and ready for review)

Viz Examples
- [x] Regret curves
- [x] Parallel Coordinates
- [x] Local Parameter Importance
- [x] Partial Dependencies

How-tos
- [x] How to plot
- [x] How to save
- [x] How to share on chart studio

Tutorials
- [x] Python API basics
- [x] Checkpointing trials
- [ ] Benchmarking

## Reviewed and approved 

Viz Examples
- [ ] Regret curves
- [ ] Parallel Coordinates
- [ ] Local Parameter Importance
- [ ] Partial Dependencies

How-tos
- [ ] How to plot
- [ ] How to save
- [ ] How to share on chart studio

Tutorials
- [ ] Python API basics
- [ ] Checkpointing trials
- [ ] Benchmarking

PR #608, opened 2021-05-13
Add Executor class and CLI arguments for parallelism

bouthilx:


src/orion/client/experiment.py
```diff
@@ -734,7 +753,9 @@ def _optimize(self, fct, max_trials, max_broken, trial_arg, on_error, **kwargs):
                         kwargs[trial_arg] = trial
 
                     try:
-                        results = fct(**unflatten(kwargs))
+                        results = self.executor.wait(
+                            [self.executor.submit(fct, **unflatten(kwargs))]
+                        )[0]
```
donglinjy, line 758: Instead of always run `task in task` for trial execution, maybe we can provide an option to support trial executor is different from the hpo parallel executor. Also, the objective function is defined by user, if they already using dask submit in it, the it could make a weird nested tasks structure, `orion worker task -> trial task -> trial launched tasks.`

bouthilx: The reason to use wait(submit(trial)) is to have a break down of the worker's work into a serie of tasks. For instance it makes it possible to visualize the gant chart in Dask dashboard. If the trial execution is parallelizing work as well I think it makes a sensible structure.
```
| ----------------------- Worker 1 task -------------------------------| 
| ------ trial 1 execution ------|  | ------ trial 2 execution ------|
 | ---- task 0 in trial ----|        | ---- task 0 in trial ----|
    | ---- task 1 in trial ----|        | ---- task 1 in trial ----|
      | ---- task 3 in trial ----|          | ---- task 3 in trial ----|

| ----------------------- Worker 1 task -------------------------------| 
| ------ trial 1 execution ------|  | ------ trial 2 execution ------|
 | ---- task 0 in trial ----|        | ---- task 0 in trial ----|
    | ---- task 1 in trial ----|        | ---- task 1 in trial ----|
      | ---- task 3 in trial ----|          | ---- task 3 in trial ----|

<---- time --->
```
If the user uses dask for the parallelisation of Oríon and within its task, using `secede` and `rejoin` makes it possible to have multiple nested layers of execution. It is also possible to do with joblib using different `Parallel()`.

donglinjy: okay, this nested structure seems easy for dask visual. one of my point was, since we are using the same executor for worker task and trial task, then it always has to be this way, even I may just want to run my objective function in the same process as the worker task for efficiency instead of having another dask task to execute it.

bouthilx: Right, the overhead may not be justified in many contexts. We could make it optional. I have difficulty figuring out what good name this option would have. :thinking: Maybe that's a bad sign. :sweat_smile: Something like `nest_execution`? :thinking: :thinking: :thinking:

donglinjy: `nest_execution` souds good. `trial_executor` could also be a choice, it can have a value of dict with same format as worker `executor`, none by default, and can support `dask`, `joblib` or simply `worker` meaning the same executor as worker `executor`.  I guess this is something good to have, maybe can add it later.

bouthilx: This would require additional configuration options to be coherent with other options (also support in global config, not only function arguments). I would be in favor of pushing it for the next release then to make sure this one goes out on time. What do you think?

donglinjy: yeah, I agree with you :)

bouthilx: ok so I will merge right now and complete the dask example inside PR #557

PR #609, opened 2021-05-17
Test different type of categorical dims

bouthilx:
Why:

Some algorithms may have issues with different type of values. For
instance TPE was failing to sample ``None`` values because it was
recognized as `GMM failed to sample` inside TPE.

### DIFF
--- a/src/orion/algo/tpe.py
+++ b/src/orion/algo/tpe.py
@@ -406,8 +406,7 @@ class TPE(BaseAlgorithm):
 
         for j in range(shape_size):
             new_point = sampler(dimension, below_points[j], above_points[j])
-            if new_point is not None:
-                points.append(new_point)
+            points.append(new_point)
 
         return points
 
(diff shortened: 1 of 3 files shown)


src/orion/algo/tpe.py
```diff
@@ -586,13 +575,20 @@ def _build_mixture(self):
...
             pdf = numpy.argmax(self.tpe.rng.multinomial(1, self.weights))
-            new_points = self.pdfs[pdf].rvs(size=5, random_state=self.tpe.rng)
-            for pt in new_points:
+            new_points = list(
+                self.pdfs[pdf].rvs(size=attempts, random_state=self.tpe.rng)
+            )
+            while True:
+                if not new_points:
+                    raise RuntimeError(
+                        f"Failed to sample in interval ({self.low}, {self.high})"
```
donglinjy, line 589: It is still possible if user just re-run the same experiment after get this `RuntimeError`, right?

donglinjy:
Looks good, btw, where does the `None` value comes from at the first place?

donglinjy (approved):

src/orion/algo/tpe.py
bouthilx: Yes, but the user will likely hit this error again. Basically if the interval gets to narrow it means the algorithm has pretty much converged. It may not be true for all dimensions however.

bouthilx:
The None value was one choices of a categorical dimension. So TPE was recognizing it as a _skip_ and the final point was incomplete.

Issue #624, opened 2021-07-21
Support benchmark experiment executor configurable in python api


donglinjy:
For Orion experiment, we support to pass executor configure through `orion.client.experiment.ExperimentClient.tmp_executor()` but benchmark experiment executor currently can only be configured with the worker configuration.

bouthilx:
Done in #634

PR #634, opened 2021-08-17
Support to configure executor for benchmark

donglinjy:
# Description
Instead of global executor configure, support to set executor at benchmark level.
fix issue: https://github.com/Epistimio/orion/issues/624

# Changes

Adding parameter executor of type `orion.executor.base.Executor` to run the benchmark experiments.

```
def get_or_create_benchmark(
    name, algorithms=None, targets=None, storage=None, executor=None, debug=False
):
```

# Checklist

## Tests

## Documentation

## Quality

### DIFF
--- a/src/orion/client/__init__.py
+++ b/src/orion/client/__init__.py
@@ -177,6 +177,8 @@ def build_experiment(
         config_change_type: str, optional
             How to resolve config change automatically. Must be one of 'noeffect', 'unsure' or
             'break'.  Defaults to 'break'.
+    executor: `orion.executor.base.Executor`, optional
+        Executor to run the experiment
 
     Raises
     ------
--- a/src/orion/client/__init__.py
+++ b/src/orion/client/__init__.py
@@ -55,6 +55,7 @@ def build_experiment(
     heartbeat=None,
     working_dir=None,
     debug=False,
+    executor=None,
 ):
     """Build an experiment to be executable
 
@@ -240,7 +241,7 @@ def build_experiment(
 
     producer = Producer(experiment, max_idle_time)
 
-    return ExperimentClient(experiment, producer, heartbeat)
+    return ExperimentClient(experiment, producer, executor, heartbeat)
 
 
 def get_experiment(name, version=None, mode="r", storage=None):
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -693,8 +693,9 @@ class ExperimentClient:
         """
         self._check_if_executable()
 
+        # TODO: should this n_workers always be smaller than executor configured n_workers?
         if n_workers is None:
-            n_workers = orion.core.config.worker.n_workers
+            n_workers = self.executor.n_workers
 
         if max_trials is None:
             max_trials = self.max_trials
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -693,9 +693,15 @@ class ExperimentClient:
         """
         self._check_if_executable()
 
-        # TODO: should this n_workers always be smaller than executor configured n_workers?
         if n_workers is None:
             n_workers = self.executor.n_workers
+        else:
+            if n_workers > self.executor.n_workers:
+                log.warning(
+                    "The required number of workers %s is bigger than exuecutor configured %s",
+                    str(n_workers),
+                    str(self.executor.n_workers),
+                )
 
         if max_trials is None:
             max_trials = self.max_trials
--- a/src/orion/client/experiment.py
+++ b/src/orion/client/experiment.py
@@ -698,7 +698,7 @@ class ExperimentClient:
         else:
             if n_workers > self.executor.n_workers:
                 log.warning(
-                    "The required number of workers %s is bigger than exuecutor configured %s",
+                    "The required number of workers %s is bigger than executor configuration %s",
                     str(n_workers),
                     str(self.executor.n_workers),
                 )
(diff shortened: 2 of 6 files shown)


src/orion/client/__init__.py
```diff
@@ -55,6 +55,7 @@ def build_experiment(
     heartbeat=None,
     working_dir=None,
     debug=False,
+    executor=None,
```
bouthilx, line 58: Must be added to docstring too

bouthilx:
Seems good! Only tests are missing. I think next step after this PR would be to use the executor to parallelize the experiments execution instead of parallelising single experiments sequentially. (All experiments at the same time vs one experiment with parallel workers)

src/orion/client/experiment.py
```diff
@@ -693,8 +693,9 @@ def workon(
         """
         self._check_if_executable()
 
+        # TODO: should this n_workers always be smaller than executor configured n_workers?
```
bouthilx, line 696: Maybe issue a warning if n_workers is greater than the n_workers of executor. That's certainly a scenario we would like to avoid, but maybe it's not critical enough to raise an error.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -306,3 +307,60 @@ def insert_race_condition(*args, **kwargs):
...
+
+    def test_create_with_executor(self, benchmark_config, benchmark_config_py):
+
+        with OrionState():
+            config = copy.deepcopy(benchmark_config_py)
+            bm1 = get_or_create_benchmark(**config)
+
+            assert bm1.configuration == benchmark_config
+            assert bm1.executor.n_workers == orion.core.config.worker.n_workers
+
```
bouthilx, line 319: Maybe ensure that orion.core.config.worker.n_workers != 2 so that we can catch is n_workers is not overridden correctly by Joblib executor.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -306,3 +307,60 @@ def insert_race_condition(*args, **kwargs):
...
+            executor = Joblib(n_workers=2, backend="threading")
+            config["executor"] = executor
+            bm2 = get_or_create_benchmark(**config)
+
+            assert bm2.configuration == benchmark_config
```
bouthilx, line 324: Do you think executor should be part of the config? If this affects parallelism and results it may be better to include it.

tests/unittests/benchmark/test_benchmark_client.py
```diff
@@ -306,3 +307,60 @@ def insert_race_condition(*args, **kwargs):
...
+
+        def foo(x):
+            return [dict(name="result", type="objective", value=x * 2)]
+
+        def optimize(*args, **kwargs):
+            optimize.count += 1
+            return 1
+
+        with create_experiment(
+            exp_config=experiment_config, trial_config={}, statuses=[]
```
bouthilx, line 341: Do we need to pre-fill the DB with a specific experiment config? Could be empty otherwise to keep it simple.

bouthilx (approved):
LGTM! :) Just a few minor comments to address/discuss before ready for merge.

tests/unittests/benchmark/test_benchmark_client.py
donglinjy: You mean experiment_config with only a name and keep all the others empty?

tests/unittests/benchmark/test_benchmark_client.py
donglinjy: I thought this a little bit. I took it out for 2 consideration:
1) this configure is primary useful when user run the one benchmark on multiple workers(manually create benchmark separately and launch it). It should be okay that at different worker use different executor configure.
2) this configure will be persisted, and now we pass-in a executor object directly instead of the executor configure, to do the persist, need to change this or ask executor return a configure for persistence.

src/orion/client/experiment.py
donglinjy: yes, the only issue when n_workers is bigger than executor configured, at the end it will try to run the experiment at free workers and find the experiment already done.

tests/unittests/benchmark/test_benchmark_client.py
bouthilx: Do you think we would need to persist the configuration in the case of the parallelization assessment? I seems to me it would be better to. Maybe that can be another PR, since it would require supporting configuration of the executor through dict arguments instead of simply an instantiated object.

tests/unittests/benchmark/test_benchmark_client.py
bouthilx: I mean using OrionState and passing no configuration at all.

src/orion/client/experiment.py
```diff
@@ -693,9 +693,15 @@ def workon(
         """
         self._check_if_executable()
 
-        # TODO: should this n_workers always be smaller than executor configured n_workers?
         if n_workers is None:
             n_workers = self.executor.n_workers
+        else:
+            if n_workers > self.executor.n_workers:
+                log.warning(
+                    "The required number of workers %s is bigger than exuecutor configured %s",
```
bouthilx, line 701: The required number of workers %s is bigger than executor configuration %s

tests/unittests/benchmark/test_benchmark_client.py
bouthilx: Like here: https://github.com/Epistimio/orion/pull/634/files/cb4edf3667330ad6863cfde698f7591b91a281ed..5032cc1fa5c1323d65f6aa0c405530ac9816f617#diff-850efe1c00bd8d6203e80177ccbf6bcc12058cb1d15c8d56b8902728379bf94cR313

donglinjy: For parallelization assessment, yes, I think we need to pass-in executor name and configure instead of an instance for persistence, this is because it is part of the benchmark and should not change when we launch same benchmark at different nodes.

tests/unittests/benchmark/test_benchmark_client.py
donglinjy: okay, this is how it looks like now :)

Issue #653, opened 2021-09-07
Add match count to user prompt in DB commands


bouthilx:
We running `orion db rm/set`, there is a prompt for the user to confirm the operation, but there is no information about what will be executed (how many trials or exp will be affected?). These stats are printed after the operation but not before (in the prompt). The prompt messages (like here https://github.com/Epistimio/orion/blob/develop/src/orion/core/cli/db/rm.py#L34) should include this information. A function `process_trial_count` should be implemented like `process_trial_rm` to count the matching trials/exp before the operation is executed.

@mgermain

PR #656, opened 2021-09-07
[WIP] Added trial count to delete_trials message

mgermain:
Minimal edit to fix https://github.com/Epistimio/orion/issues/653, I also tried to stick to the recommendation in the issue.

There are much better ways to do this. but that would require a deeper dive into the code.

What do you suggest?

### DIFF
--- a/src/orion/core/cli/db/rm.py
+++ b/src/orion/core/cli/db/rm.py
@@ -32,7 +32,7 @@ To proceed, type again the name of the experiment: """
 
 
 TRIALS_RM_MESSAGE = """
-Matching trials of all experiments above will be deleted.
+Matching trials({}) of all experiments above will be deleted.
 To select a specific version use --version <VERSION>.
 Note that trials of all children of a given version will be deleted.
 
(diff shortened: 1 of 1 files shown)


src/orion/core/cli/db/rm.py
```diff
@@ -32,7 +32,7 @@
 
 
 TRIALS_RM_MESSAGE = """
-Matching trials of all experiments above will be deleted.
+Matching trials({}) of all experiments above will be deleted.
```
bouthilx, line 35:
```suggestion
Matching trials ({}) of all experiments above will be deleted.
```

bouthilx:
Good enough! We can optimize this latter on if needed. Could you please do it as well for `delete_experiments` and add a test for each? Tests would be [here](https://github.com/Epistimio/orion/blob/develop/tests/functional/commands/test_db_rm.py). You can simply verify if the correct number of trials is printed using capsys (doc [here](https://docs.pytest.org/en/6.2.x/capture.html#accessing-captured-output-from-a-test-function)).

PR #659, opened 2021-09-10
Revert to default inf max_trials and pool-size of 1

### LINKED ISSUE #644: Revert to default max_trials inf and pool-size 1
The new default behavior is confusing for users. It is also difficult to determine a good default `max_trials`, so having not enough or to many trials sampled by default at the start of HPO can be annoying for many users. Using inf by default and iterating with pool-size may be the best alternative.

Now that we have a support for n-workers, the argument pool-size we previously deprecated could actually make sense. By default, pool-size should be equal to number of workers. We have `n-workers` set to 1 by default, so by default we would be back to previous behavior; sampling 1 trial at a time, until max_trials. 

@mgermain

bouthilx:
[Fixes #644]

Why:

The new default behavior is confusing for users. It is also difficult to
determine a good default max_trials, so having not enough or to many
trials sampled by default at the start of HPO can be annoying for many
users. Using inf by default and iterating with pool-size may be the best
alternative.

Now that we have a support for n-workers, the argument pool-size we
previously deprecated actually make sense. By default, pool-size
should be equal to number of workers. We have n-workers set to 1 by
default, so by default we are back to previous behavior; sampling 1
trial at a time, until max_trials.

How:

The producer now takes a pool size as argument when producing. The same
applies to ExperimentClient.suggest() and ExperimentClient.workon(). The
pool size is used to sample multiple trials at a time and increase I/O
efficiency.

The producer now keeps track of number of new trials so that
if multiple workers are producing new trials with a non-seed algorithm
(hence they produce different trials and there are no conflicts leading
to backoff) they will stop if they generated together up to `pool_size`
trials.

Note:

Pool-size is moved to to worker configuration instead. Since pool-size
relates to n-workers, which is part of worker configuration, having
pool-size in worker configuration makes more sense.


src/orion/client/experiment.py
```diff
@@ -509,6 +509,16 @@ def suggest(self):
         Otherwise, the algorithm is used to generate a new trial that is registered in storage and
         reserved.
 
+        Parameters
+        ----------
+        pool_size: int, optional
+            Number of trials to sample at a time. If 0, default to global config if defined,
+            else 1.  Increase it to improve the sampling speed if workers spend too much time
+            waiting for algorithms to sample points. An algorithm will try sampling `pool-size`
```
donglinjy, line 517:
```suggestion
            waiting for algorithms to sample points. An algorithm will try sampling `pool_size`
```

src/orion/client/experiment.py
```diff
@@ -652,6 +668,11 @@ def workon(
             objective.
         n_workers: int, optional
             Number of workers to run in parallel. Defaults to value of global config.
+        pool_size: int, optional
+            Number of trials to sample at a time. If 0, defaults to `n_workers` or value of global
+            config if defined.  Increase it to improve the sampling speed if workers spend too much
+            time waiting for algorithms to sample points. An algorithm will try sampling
+            `pool-size` trials but may return less.
```
donglinjy, line 675:
```suggestion
            `pool_size` trials but may return less.
```

docs/src/user/config.rst
```diff
@@ -400,6 +384,20 @@ n_workers
...
+pool_size
+~~~~~~~~~
+
+:Type: int
+:Default: 0
+:Env var:
+:Description:
+    Number of trials to sample at a time. If 0, default to number of workers.
+    Increase it to improve the sampling speed if workers spend too much time
+    waiting for algorithms to sample points. An algorithm will try sampling `pool-size`
```
donglinjy, line 398:
```suggestion
    waiting for algorithms to sample points. An algorithm will try sampling `pool_size`
```

donglinjy:
The new behavior does make more sense to me.

donglinjy (approved):

bouthilx:
Thanks @donglinjy for the review!

PR #662, opened 2021-09-13
Fix grouping of plots in legend

lebrice:


bouthilx:
That's all that was required for this fix! :open_mouth: Thank you!

PR #684, opened 2021-10-26
Rework workon

### LINKED ISSUE #613: Change workon to work on multiple trials locally instead of parallelizing sequential loops.
To limit I/O on DB and support online reports (instead of report only at end), it would be better to centralize the algorithm's loop in a master process. The master process would sample enough trials to feed all workers and suggest new trials anytime a worker is done and waiting for a new trial. We could add an option for backward compatibility to make the master process still load all DB if the user wants to distribute multiple masters and continue using Oríon with multiple independent workers.

Delaunay:
[Fixes #613]

# Description

Refactor how ExperimentClient schedule work.
Each workers now share a single HPO that dispatch work to them

# Checklist

## Tests

## Documentation

## Quality


src/orion/client/experiment.py
```diff
@@ -33,51 +30,39 @@
...
-def reserve_trial(experiment, producer, pool_size, timeout):
+def reserve_trial(experiment, producer, pool_size, timeout=None):
     """Reserve a new trial, or produce and reserve a trial if none are available."""
     log.debug("Trying to reserve a new trial to evaluate.")
 
-    trial = None
-    start = time.time()
-    failure_count = 0
+    if timeout is not None:
+        log.warning("reservation_timeout is deprecated")
```
bouthilx, line 38:
```suggestion
        log.warning("Reservation_timeout is deprecated and will be removed in v0.4.0. Use idle_timeout instead.")
```
We should already plan for the deprecation time. I would be in favor of v0.4.0.

src/orion/client/experiment.py
```diff
@@ -105,16 +90,25 @@ def __init__(self, experiment, producer, executor=None, heartbeat=None):
...
-            **orion.core.config.worker.executor_configuration,
-        )
+        self._executor = executor
+        self._executor_owner = False
         self.plot = PlotAccessor(self)
 
     ###
     # Attributes
     ###
+    @property
```
bouthilx, line 100: I like this! Is it tested? I can't find one.

tests/unittests/client/test_experiment_client.py
```diff
@@ -676,6 +672,9 @@ def amnesia(num=1):
 
             assert len(experiment.fetch_trials()) == 1
 
+            with pytest.raises(WaitingForTrials):
```
bouthilx, line 675: It's a bit confusing now. It could just return [new_value] (without tacking track of a count). First suggest works, next ones will lead to WaitingForTrials.

bouthilx: Actually, to test ReservationRaceCondition we could instead mock producer.produce to just to nothing and return 1.

bouthilx: Ah, it's done at test_suggest_reserve_race_condition. Ok, please forget my second comment. :)

src/orion/client/experiment.py
```diff
@@ -738,6 +729,11 @@ def workon(
             If the callblack returns False, the error will be ignored, otherwise it is counted
             for the threshold `max_broken`. In case of critical errors, you may also directly
             raise an error and force break out of ``workon``.
+        idle_timeout: int, optional
+            Maximum time allowed for idle workers. LazyWorker will be raised if
```
bouthilx, line 733:
```suggestion
            Maximum time (seconds) allowed for idle workers. LazyWorker will be raised if
```

src/orion/client/experiment.py
```diff
@@ -738,6 +729,11 @@ def workon(
...
+            timeout is reached. Such timeout are generally caused when reaching the
+            end of the optimization when no new trials can be sampled for the idle workers.
+            Defaults to ``orion.core.config.worker.idle_timeout``.
```
bouthilx, line 736: Missing `LazyWorkers` in `Raises` section.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+
+    return fct(**unflatten(kwargs))
+
+
+log = logging.getLogger(__name__)
+
+
+@contextmanager
+def _timer(self, name):
+    start = time.time()
```
bouthilx, line 51: `time.process_time`?

bouthilx: Why not make it a method directly?

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+@dataclass
+class _Stat:
+    sample: int = 0
+    scatter: int = 0
+    gather: int = 0
+
+    def time(self, name):
+        """Measure elapsed  time of a given block"""
```
bouthilx, line 66:
```suggestion
        """Measure elapsed time of a given block"""
```

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+        self.client = client
+        self.fct = fct
+        self.pool_size = pool_size
+        self.max_trials_per_worker = max_trials_per_worker
+        self.max_broken = max_broken
+        self.trial_arg = trial_arg
+        self.on_error = on_error
+        self.kwargs = kwargs
+
+        self.gather_timeout = 0.01
```
bouthilx, line 104: Should be an argument.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+
+        if interrupt_signal_code is None:
+            interrupt_signal_code = orion.core.config.worker.interrupt_signal_code
+
+        self.interrupt_signal_code = interrupt_signal_code
+
+    @property
+    def is_done(self):
+        """Returns true if the experiment has finished."""
+        return self.client.is_done
```
bouthilx, line 122: Maybe this should be more specific to the runner. It will also depend on `max_trials_per_worker`, so the runner could be done before the experiment is.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+
+                log.debug(f"Workers have been idle for {idle_time:.2f} s")
+            else:
+                idle_start = time.time()
+                idle_time = 0
+
+            if self.is_idle and idle_time > self.idle_timeout:
+                msg = f"Workers have been idle for {idle_time:.2f} s"
+
+                if self.pending_trials:
```
bouthilx, line 180: How is it possible to have all workers idle and pending_trials?

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+                self.client.release(trial)
+            except AlreadyReleased:
+                pass
+
+    def _suggest_trials(self, count):
+        """Suggest a bunch of trials to be dispatched to the workers"""
+        trials = []
+        for _ in range(count):
+            try:
+                trial = self.client.suggest()
```
bouthilx, line 307: Trials should be created in batch by algorithm using `self.client.suggest(pool_size=count)`. It will only return 1 trial, but if the algorithm needs to generate new trials it will do it in batch. We should pass `count` if `self.pool_size == 0` else pass `self.pool_size`.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+        """Schedule new trials to be computed"""
+        new_futures = []
+        for trial in new_trials:
+            future = self.client.executor.submit(
+                _optimize, trial, self.fct, self.trial_arg, **self.kwargs
+            )
+            self.pending_trials[future] = trial
+            new_futures.append(future)
+
+        self.free_worker -= len(new_futures)
```
bouthilx, line 225: Would it make sense to integrate `free_workers` in the executor instead?

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+                ):
+                    log.error(result.traceback)
+
+                else:
+                    log.error(str(exception))
+                    log.debug(result.traceback)
+
+                # if we receive too many broken trials, it might indicate the user script
+                # is broken, stop the experiment and let the user investigate
+                if self.worker_broken_trials >= self.max_broken:
```
bouthilx, line 285: We should wait and loop over all results before raising this error otherwise we may loose information of completed trials.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def gather(self):
+        """Gather the results from each worker asynchronously"""
+        results = []
+        try:
+            results = self.client.executor.async_get(
+                self.futures, timeout=self.gather_timeout
+            )
+        except (KeyboardInterrupt, InvalidResult):
+            self.release_all()
```
bouthilx, line 237: Where would InvalidResult come from? We don't want to release the trials and set them to interrupted if they are actually broken.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+                if self.worker_broken_trials >= self.max_broken:
+                    raise BrokenExperiment("Worker has reached broken trials threshold")
+
+        if to_be_raised is not None:
+            log.debug("Runner was interrupted")
+            self.release_all()
+            raise to_be_raised
+
+    def release_all(self):
+        """Release all the trials that were reserved by this runner"""
```
bouthilx, line 294: Maybe the docstring should clarify that the executor is expected to be terminated after self.release_all otherwise workers will continue working on trials that are released which could lead to issues.

src/orion/core/__init__.py
```diff
@@ -294,6 +294,18 @@ def define_worker_config(config):
         ),
     )
 
+    worker_config.add_option(
```
bouthilx, line 297: `reservation_timeout` above should be deprecated like the `max_idle_time` above. :sweat_smile: We could set deprecation version to v0.4.0.

src/orion/core/cli/hunt.py
```diff
@@ -205,12 +202,12 @@ def main(args):
     if config.get("worker"):
         worker_config.update(config.get("worker"))
 
-    signal.signal(signal.SIGTERM, _handler)
+    with sigterm_as_interrupt():
```
bouthilx, line 205: the `with` clause can be move directly above the call to `workon`.

src/orion/core/utils/__init__.py
```diff
@@ -180,3 +182,25 @@ def __call__(cls, of_type, *args, **kwargs):
...
+
+
+@contextmanager
+def sigterm_as_interrupt():
+    """Intercept ``SIGTERM`` signals and raise ``KeyboardInterrupt`` instead"""
+    ## Signal only works inside the main process
+    previous = None
+    try:
+        previous = signal.signal(signal.SIGTERM, _handler)
+    except ValueError:
```
bouthilx, line 200: Do we want to silence the error if `signal.signal` is called in a child process? It seems to me we want to know because we should not rely on sigterm_as_interrupt in this situation.

src/orion/core/utils/exceptions.py
```diff
@@ -40,6 +40,15 @@ class RaceCondition(Exception):
     pass
 
 
+class ReservationRaceCondition(Exception):
+    """Raised when a worker tries to reserve a trial that was
+    recently generated but another worker snatched it first.
```
bouthilx, line 45:
```suggestion
    """Raised when a runner tries to reserve a trial that was
    recently generated but another runner snatched it first.
```

src/orion/core/worker/consumer.py
```diff
@@ -30,7 +30,16 @@
 class ExecutionError(Exception):
     """Error raised when Orion is unable to execute the user's script without errors."""
 
-    pass
+    def __init__(self):
```
bouthilx, line 33:
```suggestion
    def __init__(self, return_code=0):
```
Maybe?

src/orion/core/worker/trial.py
```diff
@@ -516,5 +522,5 @@ def __exit__(self, exc_type, exc_value, traceback):
                 self._cm_experiment.release(self._cm_trial, "broken")
             elif self._cm_trial.status == "reserved":
                 self._cm_experiment.release(self._cm_trial)
-        except RuntimeError as e:
+        except AlreadyReleased as e:
```
bouthilx, line 525: `client.release` should be adjusted accordingly: https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L477

src/orion/executor/base.py
```diff
@@ -42,6 +63,25 @@ def wait(self, futures):
...
+        ----------
+        futures: `concurrent.futures.Futures` or equivalent interface
+            The objects returned by ``submit()`` of the executor.
+
+        timeout: int
+            time to wait before checking the other future
+
+        Returns
+        -------
+        returns a list fo results
```
bouthilx, line 80:
```suggestion
        returns a list of results
```

src/orion/executor/dask_backend.py
```diff
@@ -42,10 +81,31 @@ def wait(self, futures):
...
+        results = []
+        tobe_deleted = []
+
+        for future in futures:
+            if timeout:
+                future.wait(timeout)
+
+            if future.ready():
+                try:
+                    results.append(AsyncResult(future, future.get()))
```
bouthilx, line 96: if `future.future.result` raises an error, would that mean that `future.wait()` above would raise the error before it can be caught here?

src/orion/executor/dask_backend.py
```diff
@@ -42,10 +81,31 @@ def wait(self, futures):
         results = self.client.gather(list(futures))
         if self.in_worker:
             rejoin()
+        return [r.get() for r in results]
+
+    def async_get(self, futures, timeout=0.01):
```
bouthilx, line 86: Side effect should be documented here. `futures` will be modified.

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class PoolExecutor(BaseExecutor):
+    """Simple Pool executor.
+
+    Parameters
+    ----------
+
+    n_workers: int
+        Number of worker to spawn
```
bouthilx, line 138:
```suggestion
        Number of workers to spawn
```

src/orion/executor/joblib_backend.py
```diff
@@ -1,39 +1,13 @@
...
+class Joblib(PoolExecutor):
     def __init__(self, n_workers=-1, backend="loky", **config):
-        super(Joblib, self).__init__(n_workers=n_workers)
-        self.backend = backend
-        self.config = config
-
-        self.joblib_parallel = joblib.parallel_backend(
-            self.backend, n_jobs=self.n_workers, **self.config
+        warnings.warn(
+            "Joblib backend is deprecated, use PoolExecutor instead", DeprecationWarning
```
bouthilx, line 11: We should choose a version for the deprecation. I would suggest v0.4.0.

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+    backend: str
+        Pool backend to use; thread or multiprocess, defaults to multiprocess
+
+    """
+
+    BACKENDS = dict(
+        thread=ThreadPool,
+        threading=ThreadPool,
+        multiprocess=Pool,
+        loky=Pool,
```
bouthilx, line 149:
```suggestion
        loky=Pool,   # TODO: For compatibility with joblib backend. Remove in v0.4.0.
```

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+        state = super(PoolExecutor, self).__getstate__()
+        return state
+
+    def __setstate__(self, state):
+        super(PoolExecutor, self).__setstate__(state)
+
+    def submit(self, function, *args, **kwargs) -> AsyncResult:
+        return self._submit_cloudpickle(function, *args, **kwargs)
+
+    def _submit_python(self, function, *args, **kwargs) -> AsyncResult:
```
bouthilx, line 175: Never used, could this be removed?

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(self, future, cloudpickle=False):
+        self.future = future
+        self.cloudpickle = cloudpickle
+
+    def get(self, timeout=None):
+        r = self.future.get(timeout)
+        return pickle.loads(r) if self.cloudpickle else r
+
+    def wait(self, timeout=None):
+        return self.future.wait(timeout)
```
bouthilx, line 50: Is the future object here expected to be a `concurrent.futures.Future`? I cannot find the method wait in the docs: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+            if id(status.future) == id(future):
+                return status.exception
+
+    def async_get(self, futures, timeout=None):
+        results = []
+        tobe_deleted = []
+
+        for future in futures:
+            if timeout:
+                future.wait(timeout)
```
bouthilx, line 196: I cannot find the doc for future.wait so I can't say for sure, but I'm concerned the wait could raise an error if the child process crash during future.wait() instead of during future.get().

src/orion/executor/single_backend.py
```diff
@@ -5,8 +5,43 @@
...
+    def wait(self, timeout=None):
+        try:
+            self.result = self.future()
+        except Exception as e:
+            self.exception = e
+
+    def ready(self):
+        return self.result is None
+
+    def successful(self):
```
bouthilx, line 40: This function should be documented at a base future class. I would not expect it to raise ValueError is the future is not done.

src/orion/executor/dask_backend.py
```diff
@@ -1,13 +1,52 @@
...
+    def wait(self, timeout=None):
+        try:
+            self.future.result(timeout)
+        except TimeoutError:
+            pass
+
+    def ready(self):
+        return self.future.done()
+
+    def successful(self):
```
bouthilx, line 43: I can't find this method in concurrent.futures.Future. We should have a base class specifying the interface of Future If we depart from python's Future.

src/orion/executor/single_backend.py
```diff
@@ -17,13 +52,30 @@ class SingleExecutor(BaseExecutor):
...
     def wait(self, futures):
-        return [future() for future in futures]
+        return [future.get() for future in futures]
+
+    def async_get(self, futures, timeout=0.01):
+        if len(futures) == 0:
+            return []
+
+        results = []
+        try:
```
bouthilx, line 72: No loop because we don't expect to have more than one future at a time?

bouthilx: We could still have multiple futures using async_get(). :thinking:

tests/functional/client/black_box.py
```diff
@@ -44,6 +44,7 @@ def execute():
     if "data" in kwargs:
         kwargs["data"] = [dict(name=kwargs["data"], type="constraint", value=1.0)]
 
+    print("HERE")
```
bouthilx, line 47: Hello :)

tests/functional/configuration/test_all_options.py
```diff
@@ -612,6 +615,7 @@ class TestWorkerConfig(ConfigurationTestSuite):
         "worker-max-trials": 1,
         "worker-max-broken": 8,
         "reservation-timeout": 18,
+        "idle-timeout": 19,
```
bouthilx, line 618: idle-timeout should also be tested here: https://github.com/Epistimio/orion/pull/684/files#diff-fcc8709fe851b3f4c1151b08c4b72c00f1300440b30d4da082c6740a2874fb73R697

tests/unittests/client/test_experiment_client.py
```diff
@@ -1228,48 +1270,38 @@ def test_workon_on_error_raise(self):
...
             with pytest.raises(NotImplementedError) as exc:
-                client.workon(foo, max_trials=5, max_broken=5, on_error=on_error)
+                client.workon(
+                    foo_reraise, max_trials=5, max_broken=5, on_error=on_error
+                )
 
             assert exc.match("Do not ignore this!")
 
     def test_parallel_workers(self, monkeypatch):
         """Test parallel execution with joblib"""
```
bouthilx, line 1286:
```suggestion
        """Test parallel execution with multiprocess"""
```

tests/unittests/client/test_experiment_client.py
```diff
@@ -1228,48 +1270,38 @@ def test_workon_on_error_raise(self):
...
             cfg,
             experiment,
             client,
         ):
 
-            monkeypatch.setattr(client, "_optimize", optimize)
-            optimize.count = 0
             with client.tmp_executor("joblib", n_workers=5, backend="threading"):
-                client.workon(foo, max_trials=5, n_workers=2)
+                trials = client.workon(foo_1, max_trials=5, n_workers=2)
```
bouthilx, line 1295: max_trials should be adjusted in this test so that we can verify that the runner is indeed running with the correct number of workers. The main goal of this test was to verify this in particular, which was done with `optimize.count`.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+        """Show the elapsed time of different blocks"""
+        lines = [
+            f"Sample  {self.sample:7.4f}",
+            f"Scatter {self.scatter:7.4f}",
+            f"Gather  {self.gather:7.4f}",
+        ]
+        return "\n".join(lines)
+
+
+class Runner:
```
bouthilx, line 79: Majority of the tests of ExperimentClient.workon are now tests for Runner. They should be moved to another test module for Runner specifically where additional tests will be needed for the new parts of Runner.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+        """Sample new trials for all free workers"""
+        ntrials = len(self.pending_trials) + self.trials
+        remains = self.max_trials_per_worker - ntrials
+
+        # try to get more work
+        new_trials = []
+        n_trial = min(self.free_worker, remains)
+        should_sample_more = (not self.is_done) and self.free_worker > 0 and remains > 0
+
+        if should_sample_more:
```
bouthilx, line 206: The logic should be tested here. Is `n_trial` passed to `self._suggest_trials()` the value we expect, in a case where we have no pending_trials, no free worker, less remaining than free workers or vice versa, if the experiment is done or broken. There are many moving parts.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+                        f"{msg}; worker has leg room (has_remaining: {self.has_remaining})"
+                        f" and optimization is not done (is_done: {self.is_done})"
+                    )
+
+                raise LazyWorkers(msg)
+
+        return self.trials
+
+    def sample(self):
+        """Sample new trials for all free workers"""
```
bouthilx, line 197: Should note that more samples maybe added to database depending on `self.pool_size`.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+            )
+        except (KeyboardInterrupt, InvalidResult):
+            self.release_all()
+            raise
+
+        to_be_raised = None
+        log.debug(f"Gathered new results {len(results)}")
+
+        # register the results
+        for result in results:
```
bouthilx, line 244: Should be tested that number of free workers is correctly updated through scatter and gather as well as `self.pending_trials`.  I think the rest of this method is well tested with current tests for workon in ExperimentClient.

src/orion/client/runner.py
bouthilx: Looks like we would be missing tests for this part.

src/orion/client/runner.py
bouthilx: Also self.pending_trials would still contain the trials even though they are released. We really expect this class to be deleted after calling this method.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+    def release_all(self):
+        """Release all the trials that were reserved by this runner"""
+        # Sanity check
+        for _, trial in self.pending_trials.items():
+            try:
+                self.client.release(trial)
+            except AlreadyReleased:
+                pass
+
+    def _suggest_trials(self, count):
```
bouthilx, line 302: Should test here behavior between `count` and `self.pool_size`.

src/orion/client/runner.py
```diff
@@ -0,0 +1,323 @@
...
+                idle_end = time.time()
+                idle_time += idle_end - idle_start
+                idle_start = idle_end
+
+                log.debug(f"Workers have been idle for {idle_time:.2f} s")
+            else:
+                idle_start = time.time()
+                idle_time = 0
+
+            if self.is_idle and idle_time > self.idle_timeout:
```
bouthilx, line 177: Should add tests for idle_timeout for each use case supported below (if pending or has remaining).

tests/unittests/executor/test_executor.py
```diff
@@ -0,0 +1,119 @@
...
+
+@pytest.mark.parametrize("backend", backends)
+def test_execute_function(backend):
+    with backend(5) as executor:
+        future = executor.submit(function, 1, 2, c=3)
+        assert executor.wait([future]) == [7]
+
+
+@pytest.mark.parametrize("backend", backends)
+def test_execute_bad_function(backend):
```
bouthilx, line 41: Should test for get_async as well.

src/orion/executor/dask_backend.py
bouthilx: It should also be tested in a generic test for async_get.

tests/unittests/executor/test_executor.py
```diff
@@ -0,0 +1,119 @@
...
+    with backend(5) as executor:
+        future = executor.submit(bad_function, 1, 2, 3)
+        with pytest.raises(BadException):
+            executor.wait([future])
+
+
+@pytest.mark.parametrize("backend", backends)
+def test_execute_async(backend):
```
bouthilx, line 49: There should be a test for the timeout.

tests/unittests/executor/test_executor.py
```diff
@@ -0,0 +1,119 @@
...
+    with backend(5) as executor:
+        futures = [executor.submit(function, 1, 2, i) for i in range(10)]
+
+        total_task = len(futures)
+        results = executor.async_get(futures)
+
+        assert len(results) <= total_task, "Maybe not all tasks were completed"
```
bouthilx, line 56: We should make sure at least some of the functions were executed, otherwise we are not testing the results properly and futures may not be removed.

src/orion/executor/multiprocess_backend.py
```diff
@@ -0,0 +1,209 @@
...
+        return _Future(self.pool.apply_async(function, args=args, kwds=kwargs))
+
+    def _submit_cloudpickle(self, function, *args, **kwargs) -> AsyncResult:
+        payload = cloudpickle.dumps((function, args, kwargs))
+        return _Future(self.pool.apply_async(_couldpickle_exec, args=(payload,)), True)
+
+    def wait(self, futures):
+        return [future.get() for future in futures]
+
+    def _find_future_exception(self, future):
```
bouthilx, line 185: Is this needed somewhere?

src/orion/core/worker/trial.py
Delaunay: isnt it already adjusted ?


```
    def release(self, trial, status="interrupted"):
        """Release a trial.

        Release the reservation and stop the heartbeat.

        Experiment must be in writable ('w') or executable ('x') mode.

        Parameters
        ----------
        trial: `orion.core.worker.trial.Trial`
            Trial to reserve.
        status: str, optional
            Set the trial to given status while releasing the reservation.
            Defaults to 'interrupted'.

        Raises
        ------
        `RuntimeError`
            If reservation of the trial has been lost prior to releasing it.
        `AlreadyReleased`
            If reservation of trial was already released
        `ValueError`
            If the trial does not exist in storage.
        `orion.core.utils.exceptions.UnsupportedOperation`
            If the experiment was not loaded in writable mode.

        """
        ```

File: src/orion/executor/dask_backend.py (hunk 23 above)
Comment (Delaunay) [reply src/orion/executor/dask_backend.py:96]: yeah, good catch, I updated to `wait` captures the exception and it is only thrown when `.get` is called

File: src/orion/executor/dask_backend.py (hunk 24 above)
Comment (Delaunay) [reply src/orion/executor/dask_backend.py:86]: The side effect is documented in the mother. Sphinx should document this method with the mother's docstring in that case

```
"""Retrieve futures that completed, removes them from the list of pending futures
        and return their results

        Parameters
        ----------
        futures: `concurrent.futures.Futures` or equivalent interface
            The objects returned by ``submit()`` of the executor.

        timeout: int
            time to wait before checking the other future

        Returns
        -------
        returns a list fo results

        """
```

File: src/orion/executor/multiprocess_backend.py (hunk 28 above)
Comment (Delaunay) [reply src/orion/executor/multiprocess_backend.py:175]: removed

File: src/orion/executor/multiprocess_backend.py [hunk 48]
```diff
@@ -0,0 +1,209 @@
...
+    def __setstate__(self, state):
+        super(PoolExecutor, self).__setstate__(state)
+
+    def submit(self, function, *args, **kwargs) -> AsyncResult:
+        return self._submit_cloudpickle(function, *args, **kwargs)
+
+    def _submit_python(self, function, *args, **kwargs) -> AsyncResult:
+        return _Future(self.pool.apply_async(function, args=args, kwds=kwargs))
+
+    def _submit_cloudpickle(self, function, *args, **kwargs) -> AsyncResult:
```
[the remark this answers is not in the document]
Delaunay: removed

src/orion/executor/multiprocess_backend.py
Delaunay: * ``_Future`` is https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.AsyncResult.wait

* ``_ThreadFuture`` is https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future

src/orion/executor/multiprocess_backend.py
Delaunay: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.AsyncResult.wait

src/orion/executor/single_backend.py
Delaunay: added

src/orion/executor/dask_backend.py
Delaunay: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.AsyncResult.successful

src/orion/executor/single_backend.py
Delaunay: We cant have multiple futures with async_get() because async_get() is what "schedules" the jobs (for the single_backend)
So no work has started until `fut.get()` is called and we know that none of the other futures are ready because we have not called `.get` yet

tests/functional/client/black_box.py
Delaunay: removed

src/orion/executor/multiprocess_backend.py
Delaunay: removed

tests/unittests/executor/test_executor.py
Delaunay: added assert to make sure we got some results and the finished future got popped

tests/unittests/executor/test_executor.py
Delaunay: added

tests/unittests/executor/test_executor.py
Delaunay: added

tests/unittests/client/test_experiment_client.py
Delaunay: Is it really part of the Orion testing scope ?
if `multiprocessing.Pool`/`Dask`/`ThreadPoolExecutor` does not spawn the correct number of workers it is not a bug for us to solve

tests/functional/configuration/test_all_options.py
Delaunay: my browser does not bring me to the the place your are pointing to

src/orion/client/runner.py
Delaunay: most of those are using workon to access runner, I suggest we create a story to address this tech debt later

src/orion/client/runner.py
Delaunay: I added some tests for it

src/orion/client/runner.py
Delaunay: `self.pool_size` is actually a dead attribute, it is never used

src/orion/client/runner.py
Delaunay: Should we simply remove `self.pool_size` ?

src/orion/client/runner.py
Delaunay: added

src/orion/client/runner.py
Delaunay: I removed the `free workers` accounting now it just depends on the number of pending trial minus the number of worker

src/orion/client/runner.py
Delaunay: We already have `has_remaining()` for that, makes `is_running` easier to understand
Currently each is_running condition is represented by a single property that represent a single condition.
All those conditions are combined in is_running alone.

So each condition are easy to understand and the combination is easy to understand if we mix conditions more it will start to become less straight forward

src/orion/client/runner.py
bouthilx: It should be used, see comment [here](https://github.com/Epistimio/orion/pull/684/files/94238e7496426a1d7b5ecf18b2a955d82985adbc#r762077645)

examples/plotting/plot_4_partial_dependencies.py
```diff
@@ -99,16 +99,16 @@
 import time
 
 experiment = get_experiment("2-dim-exp", storage=storage)
-start = time.clock()
+start = time.process_time()
```
bouthilx, line 102: I think `perf_counter` would be better than `process_time`. https://docs.python.org/3/library/time.html#time.perf_counter

src/orion/client/experiment.py
bouthilx: Looks like tests are still missing for this.

src/orion/client/experiment.py
```diff
@@ -738,12 +732,20 @@ def workon(
...
             Constant argument to pass to `fct` in addition to trial.params. If values in kwargs are
             present in trial.params, the latter takes precedence.
 
         Raises
         ------
+        :class:`orion.core.utils.exceptions.LazyWorkers`
+             If workers stay idle for too long.
+
         :class:`orion.core.utils.exceptions.InvalidResult`
              If results returned by `fct` have invalid format.
```
bouthilx, line 751: Can it still raise ReservationTimeout? I think it is ReservationRaceCondition now, but even that should be caught by the  runner and not be raised up to workon().

src/orion/client/runner.py
bouthilx: We would be missing the tests for runner later on anyway. Moving the existing tests should be relatively painless.

src/orion/client/runner.py
bouthilx: Sounds good!

src/orion/client/runner.py
bouthilx: Looks like this comment still applies.

src/orion/client/runner.py
bouthilx: free_workers is removed, so closing this comment.

src/orion/client/runner.py
bouthilx: This comment still applies.

src/orion/client/runner.py
bouthilx: Is it possible?

src/orion/client/runner.py
bouthilx: Done

src/orion/client/runner.py
bouthilx: Removed, closing the comment.

src/orion/core/cli/hunt.py
bouthilx: Done.

src/orion/core/utils/__init__.py
bouthilx: Done.

src/orion/core/worker/trial.py
bouthilx: Indeed!

src/orion/executor/dask_backend.py
bouthilx: There should be a test for this behavior.

src/orion/executor/multiprocess_backend.py
bouthilx: Documentation of the class should explain what differs from multiprocessing.pool.AsyncResult. I know the code is simple but users should be able to understand this difference from doc directly.

src/orion/executor/multiprocess_backend.py
bouthilx: There is no information on how this is handled in the doc. :( I would expect that the error is only raised during get(), but it is not specified.

tests/functional/configuration/test_all_options.py
bouthilx: Oops, my neither. The changes in code broke the link. Here's in current develop branch: https://github.com/Epistimio/orion/blob/develop/tests/functional/configuration/test_all_options.py#L693

tests/unittests/client/test_experiment_client.py
bouthilx: No indeed, another option is to test that the backend receives the correct number of workers as argument.

src/orion/client/runner.py
```diff
@@ -0,0 +1,344 @@
...
+        self.interrupt_signal_code = interrupt_signal_code
+
+    @property
+    def free_worker(self):
+        """Returns the number of free worker"""
+        return min(
+            max(self.client.executor.n_workers - len(self.pending_trials), 0),
+            # This comes from `workon(n_workers)` it limits the number of worker
+            # that can be used by the executor, even if the executor is created with
+            # 5 workers if pool_size is 2 then only 2 workers will be used
```
bouthilx, line 123: There is a confusion here. The pool-size does not control the number of workers that can be used nor the maximum number of trials that can be pending in total. It controls the number of trials that the algorithm may sample per batch. When we call `suggest(pool_size)`, the algorithm will return `pool_size` samples. We can still call it more times if we require more trials to feed every workers. The pool-size is a way to trade-off between efficiency of algorithm sampling versus quality of samples.  Large pool-size is more efficient in wall clock time, but it may return more trials than necessary, therefore some trials were sampled with less information than would have been available when the additional were really needed. 

Maybe `batch_size` would be more clear than `pool_size`.

src/orion/client/runner.py
```diff
@@ -0,0 +1,344 @@
...
+                pass
+
+        self.pending_trials = dict()
+
+    def _suggest_trials(self, count):
+        """Suggest a bunch of trials to be dispatched to the workers"""
+        trials = []
+        for _ in range(count):
+            try:
+                pool_size = count if self.pool_size == 0 else self.pool_size
```
bouthilx, line 327: That is exactly how `pool_size` is intended to be used.

src/orion/client/runner.py
bouthilx: Ah, so you used it here: https://github.com/Epistimio/orion/pull/684/files#r770818626

tests/unittests/client/test_runner.py
```diff
@@ -0,0 +1,207 @@
...
+    assert (
+        make_runner(5, 2).should_sample() == 2
+    ), "5 processes but only 2 trials allowed"
+
+    assert (
+        make_runner(2, 5).should_sample() == 2
+    ), "2 processes and 5 max trials allowed"
+
+    assert (
+        make_runner(5, 5, 2).should_sample() == 2
```
bouthilx, line 176:
```suggestion
        make_runner(5, 5, 2).should_sample() == 5
```
It should be 5 as explained here: https://github.com/Epistimio/orion/pull/684/files#r770814583

tests/unittests/client/test_runner.py
```diff
@@ -0,0 +1,207 @@
...
+    ), "5 processes and 5 max trials allowed but pool_size is 2"
+
+    runner = make_runner(5, 10)
+    runner.trials = 5
```
bouthilx, line 180:
```suggestion
    runner.trials = 4
```

tests/unittests/client/test_runner.py
```diff
@@ -0,0 +1,207 @@
...
+    assert runner.should_sample() == 5, "5 trials are done. 5 free processes"
```
bouthilx, line 181:
```suggestion
    assert runner.should_sample() == 5, "4 trials are done. 5 free processes"
```
Better to use different values to avoid missing unintentional variable swaps.

src/orion/client/runner.py
bouthilx: I think we can close this discussion and continue it there: https://github.com/Epistimio/orion/pull/684#discussion_r762164264

tests/unittests/executor/test_executor.py
```diff
@@ -0,0 +1,157 @@
...
+def test_execute_async(backend):
+    with backend(5) as executor:
+        futures = [executor.submit(function, 1, 2, i) for i in range(10)]
+
+        total_task = len(futures)
+        results = executor.async_get(futures, timeout=1)
+
+        assert len(results) <= total_task, "Maybe not all tasks were completed"
+        assert len(results) > 0, "We got some results"
+        assert len(futures) == total_task - len(results), "Finished futures got removed"
```
bouthilx, line 66: Looks like these 2 lines are testing the same thing. I think last one is the clearest one.

src/orion/client/experiment.py
Delaunay: fixed

src/orion/core/__init__.py
Delaunay: done

src/orion/executor/multiprocess_backend.py
Delaunay: Why should they understand ? The futures are private within Orion, the differences between AsyncResult and Future are implementation details, that is why we have an adapter that unify the interfaces.

src/orion/client/experiment.py
Delaunay: added

src/orion/client/runner.py
Delaunay: moved

src/orion/client/runner.py
Delaunay: should be used now

src/orion/executor/dask_backend.py
Delaunay: added the test `test_execute_async_exception` that makes sure async_get cannot raise 
async_get is the one that calls Future.wait()

src/orion/executor/multiprocess_backend.py
Delaunay: well now it is tested anyway with `test_execute_async`

tests/functional/configuration/test_all_options.py
Delaunay: done

Delaunay: done

tests/unittests/client/test_experiment_client.py
Delaunay: the first suggestion is not working, if the amnesia.count  is not tracked WaitingForTrials is not raised

tests/unittests/executor/test_executor.py
Delaunay: removed

src/orion/client/experiment.py
bouthilx: Where? I did not find it. :(

tests/unittests/client/test_experiment_client.py
bouthilx: Right, it should only return [0] for WaitingForTrials to be raised. Since that's what we are testing (the fact that it is raising that error) we could only do this and drop the 4 lines below testing the suggest with `new_value`.

src/orion/client/experiment.py
Delaunay: https://github.com/Epistimio/orion/pull/684/files#diff-f06b94773052cdf9a38bfdb3ea74cb852af76b54d9c1acf6841cd98db127d04cR618

src/orion/client/runner.py
bouthilx: This is fixed now.

src/orion/client/experiment.py
Delaunay: https://github.com/Epistimio/orion/blob/4b25e064b7c3725783a1a62f656a7dd443381a63/tests/unittests/client/test_runner.py#L640

src/orion/client/runner.py
```diff
@@ -0,0 +1,334 @@
...
+                ):
+                    log.error(result.traceback)
+
+                else:
+                    log.error(str(exception))
+                    log.debug(result.traceback)
+
+                # if we receive too many broken trials, it might indicate the user script
+                # is broken, stop the experiment and let the user investigate
+                if self.is_broken:
```
bouthilx, line 292: This [comment](https://github.com/Epistimio/orion/pull/684#discussion_r762084215) still applies here.

src/orion/client/runner.py
bouthilx: Closing this comment in favor of https://github.com/Epistimio/orion/pull/684/files#r780406978

src/orion/client/runner.py
```diff
@@ -0,0 +1,334 @@
...
+            if isinstance(result, AsyncException):
+                if (
+                    isinstance(result.exception, ExecutionError)
+                    and result.exception.return_code == self.interrupt_signal_code
+                ):
+                    to_be_raised = KeyboardInterrupt()
+                    self.client.release(trial, status="interrupted")
+                    continue
+
+                if isinstance(result.exception, InvalidResult):
```
bouthilx, line 269: Follow up from this [comment](https://github.com/Epistimio/orion/pull/684#discussion_r762087196) here. InvalidResult is raised when the experiment tries to observe a trial. We should not expect it to occur during trial execution, but it could occur during self.client.observe a few lines above. That's where it should be handled.

src/orion/client/runner.py
bouthilx: Closing this comment in favor of https://github.com/Epistimio/orion/pull/684/files#r780410397

src/orion/client/runner.py
bouthilx: Done, thanks.

src/orion/executor/joblib_backend.py
bouthilx: Done, thanks!

src/orion/executor/multiprocess_backend.py
bouthilx: This part is indeed not at the top layer but it is exposed and available if users also want to use executors for processing other things than Oríon trials.

bouthilx:
For reference, here's a list of comments/discussions that are not closed yet.

- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r762069511
- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r762077645
- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r780406978
- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r762142947
- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r762153535
- [ ] https://github.com/Epistimio/orion/pull/684#discussion_r780410397

src/orion/client/experiment.py
bouthilx: Great, thanks! I think this test would below in test_experiment_client however since it is not testing anything about the runner. It's different for all workon tests, which are involving the runner.

src/orion/client/runner.py
Delaunay: I cannot see the linked comment, can you copy paste it here ?

tests/unittests/client/test_experiment_client.py
Delaunay: then the second suggest will also raise WaitingForTrials

src/orion/client/experiment.py
Delaunay: moved

src/orion/client/runner.py
Delaunay: move the catching of invalid result

src/orion/executor/multiprocess_backend.py
Delaunay: Still not sure what you want me to do about it.
The Future Interface is documented, all the adapters inherit that doc and the adapter mention explicitly which flavor of future they are adapting

src/orion/client/runner.py
Delaunay: No, the line was removed

tests/unittests/client/test_experiment_client.py
Delaunay: I added test_executor_receives_correct_worker_count that does that

src/orion/client/runner.py
bouthilx: Oops, me neither. I managed to find it back:
> We should wait and loop over all results before raising this error otherwise we may loose information of completed trials.

src/orion/client/runner.py
bouthilx: The trial should also be released and set to broken. This would also deserve a specific test.

tests/unittests/client/test_experiment_client.py
bouthilx: Yeah, no need to test for second suggest.

src/orion/executor/multiprocess_backend.py
bouthilx: There are no differences except for cloudpickle right?

src/orion/client/runner.py
Delaunay: updated

src/orion/client/runner.py
Delaunay: added the test & the broken release

src/orion/executor/multiprocess_backend.py
Delaunay: yeah

src/orion/client/runner.py
Delaunay: yep done

src/orion/executor/multiprocess_backend.py
bouthilx: So that's the only thing I see that would need to be documented. It would be great to know why it's needed.

Delaunay: added the comment, it is to have feature parity with Dask, so things that can be pickled with one can be pickled with the other

src/orion/client/runner.py
bouthilx: Alright. There should also be a test to verify that calling gather one 1 broken and 1 completed trial does register the completed trial properly even though the one before in the list was broken with `to_be_raised is not None`.

src/orion/client/runner.py
bouthilx: Thanks!

src/orion/client/runner.py
Delaunay: added test_multi_results_with_failure

bouthilx: Very good, thanks!

tests/unittests/client/test_experiment_client.py
Delaunay: done

tests/unittests/executor/test_futures.py
```diff
@@ -0,0 +1,111 @@
...
+        raise exception
+    return result
+
+
+@pytest.mark.parametrize("backend", backends)
+def test_future(backend):
+    """Make sure the future interface is harmonized"""
+    with backend(5) as executor:
+
+        # Get - OK
```
bouthilx, line 35: I would be best to break these down into separate tests.

src/orion/core/worker/consumer.py
```diff
@@ -30,7 +30,16 @@
 class ExecutionError(Exception):
     """Error raised when Orion is unable to execute the user's script without errors."""
 
-    pass
+    def __init__(self, return_code=0):
+        super(ExecutionError, self).__init__()
+        self.return_code = return_code
+
+
+def _exec_error(return_code):
```
bouthilx, line 38: A simple test involving pickling the error returned by execute_process would help ensure that no future changes incompatible with pickling are introduced.

src/orion/client/runner.py
```diff
@@ -0,0 +1,339 @@
...
+        log.debug("Scheduled new trials")
+
+    def gather(self):
+        """Gather the results from each worker asynchronously"""
+        results = []
+        try:
+            results = self.client.executor.async_get(
+                self.futures, timeout=self.gather_timeout
+            )
+        except KeyboardInterrupt:
```
bouthilx, line 244: Not tested

src/orion/client/runner.py
```diff
@@ -0,0 +1,339 @@
...
+                self.client.release(trial, status="broken")
+
+                if self.on_error is None or self.on_error(
+                    self, trial, exception, self.worker_broken_trials
+                ):
+                    log.error(result.traceback)
+
+                else:
+                    log.error(str(exception))
+                    log.debug(result.traceback)
```
bouthilx, line 289: Codecov reports these line as not tested but it should be covered in [test_runner.py::TestWorkon::test_workon_on_error_ignore](https://github.com/Epistimio/orion/pull/684/files#diff-f06b94773052cdf9a38bfdb3ea74cb852af76b54d9c1acf6841cd98db127d04cR627) and [test_runner.py::TestWorkon::test_workon_on_error_raise](https://github.com/Epistimio/orion/pull/684/files#diff-f06b94773052cdf9a38bfdb3ea74cb852af76b54d9c1acf6841cd98db127d04cR681).

src/orion/client/runner.py
```diff
@@ -0,0 +1,339 @@
...
+        """Release all the trials that were reserved by this runner.
+        This is only called during exception handling to avoid retaining trials
+        that cannot be retrieved anymore
+
+        """
+        # Sanity check
+        for _, trial in self.pending_trials.items():
+            try:
+                self.client.release(trial)
+            except AlreadyReleased:
```
bouthilx, line 315: Not tested

src/orion/client/runner.py
```diff
@@ -0,0 +1,339 @@
...
+                trials.append(trial)
+
+            # non critical errors
+            except WaitingForTrials:
+                break
+
+            except ReservationRaceCondition:
+                break
+
+            except CompletedExperiment:
```
bouthilx, line 336: Not tested

src/orion/executor/single_backend.py
```diff
@@ -5,8 +5,49 @@
...
+from orion.executor.base import (
+    AsyncException,
+    AsyncResult,
+    BaseExecutor,
+    ExecutorClosed,
+    Future,
+)
+
+
+class _Future(Future):
```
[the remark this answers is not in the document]
Delaunay: well yes, single_backends does not have a true future

src/orion/core/worker/consumer.py
Delaunay: what

src/orion/client/runner.py
Delaunay: so ?

bouthilx: So nothing to do, just writing it down here as a reference so we understand why we did not try to address these reports from codecov.

src/orion/core/worker/consumer.py
bouthilx: Adding a test where we call Consumer.execute_process and verify that the output can be pickled will cover this part and ensure we do not introduce unnoticed changes that make the output of Consumer.execute_process unpicklable.

src/orion/executor/single_backend.py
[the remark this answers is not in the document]
bouthilx: It's not really asynchronous indeed, but the different methods should be tested anyhow. There is a common interface across the Futures, like successful raising ValueError if the future is not ready. But in this case, it's never _not ready_, right? :thinking:

src/orion/core/worker/consumer.py
bouthilx: Alright, fixed here: https://github.com/Epistimio/orion/pull/684/commits/cefca8bfdd713ff62584d6bb8cd60b7fcebca7cc

tests/unittests/client/test_runner.py
```diff
@@ -118,6 +124,43 @@ def test_stop_after_max_trial_reached():
...
+    runner = new_runner(0.01, n_workers=16)
+    runner.fct = function
+    client = runner.client
+
+    client.trials.extend([new_trial(i) for i in range(count, -1, -1)])
+
+    def bad_async(*args, **kwargs):
+        raise KeyboardInterrupt()
+
+    client.executor.async_get = bad_async
```
bouthilx, line 139: It would be best to have a test where the KeyboardInterrupt occurs between trials in the gather loop, otherwise we could have buggy code that overwrites completed or broken trials to interrupted and the current test won't detect this. Like first two trials are set to completed and broken and third trial triggers a keyboardinterrupt. Only third and remaining trials should be set to interrupted.

src/orion/client/runner.py
bouthilx: Done here https://github.com/Epistimio/orion/pull/684/commits/0d390c553d734d53440bff566f85eb124b30152e

src/orion/client/runner.py
```diff
@@ -0,0 +1,339 @@
...
+            try:
+                batch_size = count if self.batch_size == 0 else self.batch_size
+                trial = self.client.suggest(pool_size=batch_size)
+                trials.append(trial)
+
+            # non critical errors
+            except WaitingForTrials:
+                break
+
+            except ReservationRaceCondition:
```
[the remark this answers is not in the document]
bouthilx: Done here https://github.com/Epistimio/orion/pull/684/commits/0d390c553d734d53440bff566f85eb124b30152e

src/orion/client/runner.py
bouthilx: Done here https://github.com/Epistimio/orion/pull/684/commits/0d390c553d734d53440bff566f85eb124b30152e

src/orion/client/runner.py
bouthilx: Done here https://github.com/Epistimio/orion/pull/684/commits/0d390c553d734d53440bff566f85eb124b30152e

src/orion/executor/single_backend.py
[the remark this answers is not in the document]
Delaunay: yeah, I mean given that multprocess is builtin in python do we really need the single_backend ?
We could simply default to the threaded implementation.

I understood that single_backend was more for testing that for actual use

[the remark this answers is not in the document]
bouthilx: Single process can be helpful to debug user's functions in case multiprocess/multithread causes issues with the stack trace.

tests/unittests/client/test_runner.py
Delaunay: but that is not what this is testing.
Its only job is to execute the except part of this code.

```
        try:
            results = self.client.executor.async_get(
                self.futures, timeout=self.gather_timeout
            )
        except KeyboardInterrupt:
            self._release_all()
            raise
``

Comment (Delaunay) [reply tests/unittests/client/test_runner.py:139]: if you want the gather loop to ignore keyboard interrupt until everything is processed that is a different feature

Comment (bouthilx) [reply tests/unittests/client/test_runner.py:139]: I was only thinking of what was gathered before the KeyboardInterrupt occured. (during the current loop execution, not during previous gather() calls)

Comment (Delaunay) [reply tests/unittests/client/test_runner.py:139]: if the exception occur during the async_get we dont get partial results
It is akin to having the interrupt just before the report_result() call.
There is no way to know we were close to the report_result call.

File: tests/unittests/client/test_runner.py [hunk 68]
```diff
@@ -95,7 +95,7 @@ def new_runner(idle_timeout, n_workers=2, client=None):
     return runner
 
 
-def function_raise_on_2(lhs, rhs):
+def function_raise_on_2(lhs, sleep):
```
bouthilx, line 98: What about the rhs variable below?

tests/unittests/client/test_runner.py
bouthilx: Oh right, it's before the loop. But then all the rest of the gather method is not _protected_ and a keyboard interrupt at that point would leave them unreleased. 

I think this try-except should cover the whole run() execution instead of just the beginning of gather().

tests/unittests/client/test_runner.py
Delaunay: What about it ?

bouthilx: It is crying, looking for its identity, nowhere defined. :sob:

tests/unittests/client/test_runner.py
Delaunay: but the try-except is not protecting anything if we want to protect the gather execution we need to delay the Keyboard interrupt.

We need something like https://github.com/Delaunay/cqueue/blob/master/msgqueue/backends/queue.py#L420
to delay the interrupt until we can process all the partial results

src/orion/executor/single_backend.py
```diff
@@ -16,14 +16,23 @@
     Future,
 )
 
+class _None:
```
bouthilx, line 20: I think NOT_SET would be less confusing. You can also instantiate a global object instead of defining a new class, like here: https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/config.py#L21.

src/orion/executor/single_backend.py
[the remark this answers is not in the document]
Delaunay: single_backend should work almost identically to the others now.
There is a single exception when wait(0.001) is called, the single_backend will always have a result (i.e timeout is kind of irrelevant since we have no way of stopping execution once we start)

The difference is captured in the tests so that should  be fine

src/orion/executor/single_backend.py
Delaunay: done

tests/unittests/client/test_runner.py
Delaunay: this is done

src/orion/client/runner.py
```diff
@@ -27,6 +28,41 @@
 from orion.core.worker.trial import AlreadyReleased
 from orion.executor.base import AsyncException, AsyncResult
 
+log = logging.getLogger(__name__)
+
+
+class Protected(object):
```
bouthilx, line 34: This context manager is great, but does it risk overridding this signal handler when using the cmdline interface? https://github.com/Epistimio/orion/blob/develop/src/orion/core/cli/hunt.py#L208

Delaunay: Well yes, but it restores it after each scatter-gather loop

Delaunay: so the signal handler will get called, simply later when everything is processed

Delaunay: FYI https://docs.python.org/3/library/signal.html#signal.signal

> The previous signal handler will be returned

Delaunay: check test_interrupted_scatter_gather_custom_signal

tests/unittests/executor/test_futures.py
```diff
@@ -104,7 +104,7 @@ def wait_timeout(self, executor):
 
         future.wait(0.01)
 
-        # SingleExecutor is not truely async so 
+        # SingleExecutor is not truely async so
         # results are always ready after a wait
         if executor != SingleExecutor:
```
bouthilx: `not isinstance(executor, SingleExecutor)`?

src/orion/executor/single_backend.py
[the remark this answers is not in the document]
bouthilx: Sounds good! Thanks a lot!

tests/unittests/executor/test_futures.py
bouthilx: Ah, it was done already. Sorry! I was behind in the commits history.

tests/unittests/client/test_runner.py
```diff
@@ -163,6 +163,37 @@ def slow_gather():
     ), "Trials had time to finish because of the slow gather"
 
 
+class CustomExceptionForTest(Exception):
+    pass
+
+
+def test_interrupted_scatter_gather_custom_signal():
```
bouthilx, line 170: We should also have a test verifying that the previous signal handlers are properly set back on `__exit__`.

tests/unittests/executor/test_futures.py
Delaunay: already fixed

src/orion/client/runner.py
bouthilx: Yeah I know how the signals handlers are overridden and can be restored, thanks. :) I missed that [part]( https://github.com/Epistimio/orion/pull/684/files/4446b76cabe152d59aff4b552b2ee517727552f5..ad715dc0f99d76b152719d43043cb4e1a90a2d54#diff-13cdef139faa8780b0667c70442b698de549631ba0ef27680325d56259270566R63) where you manually call the previous handler. That's good.

Also, the signal handler in orion.core.cli.hunt can be removed now, as the Protected context manager is already handling this properly.

src/orion/executor/single_backend.py
bouthilx: Thanks!

tests/unittests/client/test_runner.py
bouthilx: Great, thanks! :)

tests/unittests/client/test_runner.py
Delaunay: done

bouthilx: Thanks!

src/orion/client/runner.py
bouthilx: For reference, @Delaunay made me realize that the handler in hunt is slightly different because it raises KeyboardInterrupt. So we can keep it as is. I'm closing this discussion.

Thanks!

bouthilx (approved):

PR #705, opened 2021-12-01
PBT implementation

bouthilx:
The algorithm Population Based Training is implemented using trees with _jump_ edges across them to keep track of jumps across trees during _exploitation_ phases. These edges are particularly important to keep track of trial ancestry. If a trial is broken, we look at the true ancestor (from original tree instead of exploited one) to avoid being stuck on a tree that often leads to broken trials and also to increase likelihood of exploiting other trees. This also facilitates the backtracking strategy and visualisations of population evolution.

## Set and create exp working dir inside workon 

The directory was created inside the Consumer which is specific to the
cmdline interface. It should be done inside workon so that the python
and cmdline API uses the same logic. Also the generic WorkingDir class
was not handling directly cases where experiment has or not a defined
working dir. There is no apparent need for a generic class to create
temporary directory so this commit rework the class to handle the
experiment object directly. This also allows to handle in __exit__
resetting experiment.working_dir when it is set temporarily.

## Infer trial working dir based on exp.working_dir 

Why:

The trial working dir should be unique to the trial and depend on the
experiment's working dir. We can use the id of the trial (or variants
ignoring fidelity, experiment id or lies) to define a unique working
dir.

How:

Instead of setting the working dir directly, we set the experiment's
working dir.

## Add parent attribute to Trial 

Why:

Now that we can branch a trial we need to keep a trace to the trial's
ancestry. Also, when branching a trial but keeping the same
hyperparameters, it should not lead to the same ID since it is now a
separate trial that will be executed independently. The hash of a trial
will thus also depend on the value of the parent.

How:

For simplicity the parent attribute only reference to the ID of the
parent trial (full id), just like it is done for the reference to the
experiment. We should use a lazy tree node implementation like for the
EVC experiment tree node to reference trial parent object instead of
trial id. This should be done in the future TrialClient class.

## Move tree.py to utils 

Why:

The tree node will be used for PBT and probably for trial objects as
well in the future so it must be generalized.

## Add Tree.node_depth and Tree.get_nodes_at_depth 

We need these methods for fast retrievals in the tree in PBT algorithm.


src/orion/algo/pbt/pbt.py
```diff
@@ -0,0 +1,877 @@
...
+        total population_size. Default: 50.
+    generations: int, optional
+        Number of generations, from lowest fidelity to highest one. This will determine how
+        many branchings occur during the execution of PBT. Default: 10
+    exploit: dict or None, optional
+        In the mutate part, one can define the customized mutate function with its mutate factors,
+        such as multiply factor (times/divides by a multiply factor) and add factor
+        (add/subtract by a multiply factor). The function must be defined by
+        an importable string. If None, default
+        mutate function is used: ``orion.algo.mutate_functions.default_mutate``.
```
donglinjy, line 113: can you help understand how this `orion.algo.mutate_functions.default_mutate` is used, cause I do not see this in `Exploit`. 

Also, it is not very clear how the value for  this `exploit` parameter should be.

src/orion/algo/pbt/pbt.py
```diff
@@ -0,0 +1,877 @@
...
+    explore: dict or None, optional
+        In the mutate part, one can define the customized mutate function with its mutate factors,
+        such as multiply factor (times/divides by a multiply factor) and add factor
+        (add/subtract by a multiply factor). The function must be defined by
+        an importable string. If None, default
+        mutate function is used: ``orion.algo.mutate_functions.default_mutate``.
```
donglinjy, line 119: Same as `exploit`.

src/orion/algo/pbt/exploit.py
```diff
@@ -0,0 +1,372 @@
...
+import logging
+
+import numpy
+
+from orion.core.utils import GenericFactory
+
+logger = logging.getLogger(__name__)
+
+
+class BaseExploit:
```
donglinjy, line 26: An overall comments about `Exploit`, for Deep Learning training Trial, how do we expect user to re-use the parent model `weights`?

src/orion/algo/pbt/pbt.py
```diff
@@ -0,0 +1,877 @@
...
+
+class Lineage(TreeNode):
+    """
+    Lineage node
+
+    The lineage node is based on :py:class:`orion.core.utils.tree.TreeNode`. It provides
+    additional methods to help represent lineages for PBT, in particular, ``fork``,
+    ``set_jump``, ``get_true_ancestor`` and ``get_best_trial``.
+
+    A lineage node can be connected to a parent and children, like a typical TreeNode, but
```
donglinjy, line 723: This explain confuse me a little bit as in above `Lineages` class, we say `Each lineage is a path from a leaf trial (highest fidelity level) up to the root (lowest fidelity level)`.

src/orion/algo/pbt/pbt.py
```diff
@@ -0,0 +1,877 @@
...
+        Optimisation space with priors for each dimension.
+    seed: None, int or sequence of int
+        Seed for the random number generator used to sample new trials.
+        Default: ``None``
+    population_size: int, optional
+        Size of the population. No trial will be continued until there are `population_size`
+        trials executed until lowest fidelity. If a trial is broken during execution at lowest
+        fidelity, the algorithm will sample a new trial, keeping the population of *non-broken*
+        trials at `population_size`.  For efficiency it is better to have less workers running than
+        total population_size. Default: 50.
```
donglinjy, line 104:
```suggestion
        population_size. Default: 50.
```

src/orion/algo/pbt/pbt.py
```diff
@@ -0,0 +1,877 @@
...
+                    {"of_type": "PerturbExplore", "factor": 1.2, "volatility": 0.0001},
+                ],
+            }
+
+        self.random_search = Random(space)
+        self._queue = []
+
+        fidelity_index = self.fidelity_index
+        if fidelity_index is None:
+            raise RuntimeError(SPACE_ERROR)
```
donglinjy, line 173: where is this `SPACE_ERROR` defined?

src/orion/algo/pbt/pbt.py
bouthilx: Oh right, that is very confusing. Maybe this class should be called a LineageNode?

src/orion/algo/pbt/pbt.py
bouthilx: Oops! I fill fix this, thanks!

src/orion/algo/pbt/pbt.py
bouthilx: Oops, not defined indeed!

src/orion/algo/pbt/exploit.py
bouthilx: The weights should be saved in a file (any name, it does not matter) in `trial.working_dir`. PBT takes care of copying the files over to child trials.

This is explained [here](https://github.com/Epistimio/orion/pull/705/files#diff-a133c725e860ad2cbe9a299bd5a9bd437c737287879d0630c30746e4bafdbfd8R56), but maybe it should be explained in the docstring of `BaseExploit` too?

src/orion/algo/pbt/pbt.py
bouthilx: Fixed in https://github.com/Epistimio/orion/pull/705/commits/753d8d2fc5e494835a9aefc802e4e9d1e717009e

src/orion/algo/pbt/pbt.py
bouthilx: Fixed in https://github.com/Epistimio/orion/pull/705/commits/6471414296e42671582dcdf125393f4e1c6a9f65

src/orion/algo/pbt/pbt.py
bouthilx: Fixed in https://github.com/Epistimio/orion/pull/705/commits/6471414296e42671582dcdf125393f4e1c6a9f65

src/orion/algo/pbt/pbt.py
donglinjy: Sounds better :)

src/orion/algo/pbt/exploit.py
donglinjy: I guess so, user may need to know where they should save and load the weights in the their code, as we only do the copy, right? the actual save and load still depends on user code.

bouthilx: Indeed the save and load depends on user code, otherwise we would need to write specific versions for Pytorch, Tensorflow and other frameworks. Do you think I should improve the documentation of PBT to make this more clear?

donglinjy: yeah, may be mention that with PBT in Orion, where user code should save and load weights, so an under-performing model will be replaced with better performing one.

bouthilx: I added this to documentation: https://github.com/Epistimio/orion/pull/705/commits/6c9b5ffdd301c89c30b796c8a466134e7089388d

Do you think there is something missing?

donglinjy: looks good for me. just one more thing, do we have any place explain in user code, how do they refer to the `trial.working_dir`, as a env value or something?

donglinjy (approved):
lgtm, thanks

src/orion/algo/pbt/exploit.py
bouthilx: It is documented here: https://orion.readthedocs.io/en/stable/user/script.html#command-line-templating. I'll reference it in PBT's doc.

PR #632, opened 2021-08-03
Add tasks from Profet

lebrice:
This is a draft, I will update this description in the future.

# Description
- ~~Adds tasks from OpenML~~ (Removed, will tackle this in another PR)
- Adds tasks from Profet

# TODOs
- [x] De-couple everything out of the warmstart repo
- [x] Add external dependency for simple_parsing
- [x] Add a new extras_require for profet 
  - [ ] Add checks so that the profet tasks are truly optional (things dont crash if the profet extras aren't installed, tests are skipped properly. **I'd actually need some help with this @bouthilx **
  - [x] Test that all the required packages are correctly listed (e.g. by creating a blank new conda env and then running tests)
- [x] Update the profet benchmark example script
- [ ] Tests for Profet tasks (https://github.com/Epistimio/orion/pull/632#discussion_r708512497)
  - [ ] Arguments
    - [x] benchmark
    - [x] max_trials
    - [x] task_id
    - [x] seed
    - [x] input_dir
    - [x] checkpoint_dir
    - [ ] train_config. This one would require to mock the model trained and some functions used to train it to verify that the training config is passed down the call stack properly. No need to test all the way down, only at end of stack is fine (deepest function called where the config is actually used).
    - [x] device
  - [ ] Behaviors
    - [x] Testing the checkpoint based on hash. No clash, unique and reproducible.
    - [ ] self.name. Should it contain something else than task_id? Hash maybe?
    - [x] Test seeding (test_call_is_reproducible)
    - [x] Re-using checkpoint (test_instantiating_task)
    - [x] configuration (test_configuration)
    - [x] With gradients (test_with_gradients)
- [x] Add ~~documentation?~~ docstrings everywhere :)
~~- [ ] Tests for Quadratics task~~
     ~~- [ ] Same basic tests as other tasks~~
     ~~- [ ] Test with_context (https://github.com/Epistimio/orion/pull/632#discussion_r708499306)~~

## Discussions not resolved yet:

- [x] Seeding: https://github.com/Epistimio/orion/pull/632#discussion_r729963874
- [x] Random forest predictor: https://github.com/Epistimio/orion/pull/632#discussion_r698858921
- [x] Citations: 
    - https://github.com/Epistimio/orion/pull/632#discussion_r709454071
    - https://github.com/Epistimio/orion/pull/632#discussion_r707559147
- [x] Getting rid of global configurations: https://github.com/Epistimio/orion/pull/632#discussion_r715779509
[...]

PR #764, opened 2022-01-19
Add Task Wrappers

lebrice:

# Description

# Changes

# Checklist

## Tests

## Documentation

## Quality

# Further comments

### DIFF
--- a/src/orion/benchmark/task/task_wrapper.py
+++ b/src/orion/benchmark/task/task_wrapper.py
@@ -1,0 +1,183 @@
+from abc import ABC
+from dataclasses import asdict, is_dataclass
+from typing import (
+    Any,
+    Dict,
+    Generic,
+    List,
+    TypeVar,
+    Union,
+)
+import copy
+
+import numpy as np
+from orion.algo.space import Dimension, Space
+from orion.benchmark.task.base import BaseTask, BenchmarkTask
+from logging import getLogger as get_logger
+
+from orion.core.utils.points import flatten_dims, regroup_dims
+from orion.core.utils.format_trials import dict_to_trial
+
+from orion.core.io.space_builder import SpaceBuilder
+from orion.core.utils.format_trials import trial_to_tuple, tuple_to_trial
+from orion.core.worker.trial import Trial
+
+logger = get_logger(__name__)
+
+
+TaskType = TypeVar("TaskType", bound=BaseTask)
+InputType = TypeVar("InputType")
+
+
+class TaskWrapper(BaseTask, Generic[TaskType], ABC):
+    """ ABC for a Wrapper around a Task.
+
+    Wrappers could be used to modify the search space, trials, or the results of the task.
+    
+    Parameters
+    ----------
+    task : TaskType
+        A task to wrap.
+    max_trials : int, optional
+        Maximum number of trials. By default None, in which case the max trials is the same as that
+        of the wrapped task.
+    """
+
+    def __init__(self, task: TaskType, max_trials: int = None):
+        # TODO: Should Wrappers pass the `task` as an argument to the super() constructor?
+        # When/How are these objects deserialized?
+        if isinstance(task, dict):
+            if len(task) != 1:
+                raise ValueError(
+                    f"Expected task configuration dict to have only one key (the task name), but "
+                    f"got {task} instead."
+                )
+            task_name, task_config = task.popitem()
+            task = BenchmarkTask(of_type=task_name, **task_config)
+        super().__init__(max_trials=max_trials or task.max_trials)
+        self.task: TaskType = task
+
+    def call(self, x: Any) -> List[Dict]:
+        return self.task.call(x)
+
+    def get_search_space(self) -> Dict[str, str]:
+        return self.task.get_search_space()
+
+    @property
+    def unwrapped(self) -> Union[TaskType, BaseTask]:
+        """ Returns the 'unwrapped' task. """
+        return self.task.unwrapped
+
+    @property
+    def configuration(self) -> Dict[str, Any]:
+        """Return the configuration of the task wrapper (including the wrapped task's configuration).
+        """
+        return {
+            type(self).__qualname__: {
+                "task": self.task.configuration,
+                "max_trials": self.max_trials,
+            }
+        }
+
+
+HParamsType = TypeVar("HParamsType")
+
+
+class FixTaskDimensionsWrapper(TaskWrapper[TaskType]):
+    """ Wrapper around a Task that fixes the values of some of its input dimensions.
+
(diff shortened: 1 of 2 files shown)


src/orion/benchmark/task/task_wrapper.py
```diff
@@ -0,0 +1,183 @@
...
+        else:
+            raise NotImplementedError(
+                f"Don't know how to fix some values for input {trial} of type {type(trial)}"
+            )
+
+        trial_was_modified = False
+        for name, fixed_value in self.fixed_dims.items():
+            params = trial.params
+
+            if name not in params:
```
bouthilx, line 137: I'm copy-pasting a discussion here because we lost trace with the migration.

> @bouthilx bouthilx on 14 Sep 2021 Member
> Could we assume that X should be part of the subspace? The wrapped task defines the new subspace so the algorithms should only sample in that subspace.
> 
>  
> @lebrice lebrice on 14 Sep 2021 Author Contributor
> x is a part of the space of the space of the task wrapper, not necessarily a part of the space of the wrapped task.
> Let me know if that doesn't answer your question.
> 
>  
> @bouthilx bouthilx on 15 Sep 2021 Member
> Yeah so x is part of the subspace (without the fixed dim). That means only the code in the else clause is necessary.
> 
>  
> @lebrice lebrice on 21 Sep 2021 Author Contributor
> I see. But here I also handle the case where points are given that contain that dimension. I'll leave this as-is if that's ok with you.
> 
>  
> @bouthilx bouthilx on 12 Oct 2021 Member
> But is the task supposed to receive points with dimensions not contained in the search space? (They were removed by being fixed). If it's not needed removing it would be simpler for tests and maintenance.
> 
>  
> @lebrice lebrice on 6 Dec 2021 Author Contributor
> Having a dimension being fixed is not the same as not having that dimension. I don't understand this conversation anymore.
> This is a wontfix for me atm.
> 
>  
> @bouthilx bouthilx on 16 Dec 2021 Member
> I agree this should be moved to another PR since it is not used in this one.
> 
> What I was saying above is that for the experiment and algorithm this dimension does not exist, and therefore there should never be trials that transit between algo and experiment with this dimension. If we want to reuse trials from experiments where the dimension was not fixed, these trials should already be adapted by something like the EVC or some kind of knowledge base for warm-starting before they can reach the actual wrapped_task.__call__ method.
> 
>  
> @lebrice lebrice 15 days ago Author Contributor
> Fixed: removed this file.
>

src/orion/benchmark/task/task_wrapper.py
```diff
@@ -0,0 +1,183 @@
...
+                    f"{self._full_space}."
+                )
+
+        # The part of the space that can be modified.
+        self._space: Space = SpaceBuilder().build(self.get_search_space())
+
+    def call(self, x: Any) -> List[Dict]:
+        """ Calls the wrapped task, adding/changing some dimensions in `x` if necessary. """
+        trial: Trial
+        if isinstance(x, np.ndarray):
```
bouthilx, line 121: I'm copy-pasting a discussion here because we lost trace with the migration.

> @bouthilx bouthilx on 14 Sep 2021 Member
> Same comment here about the support for different types of x.
> 
>  
> @lebrice lebrice on 6 Dec 2021 Author Contributor
> Hmm I'm starting to think that this TaskWrapper doesn't really have anything to do with the rest of this PR.
> This is a wontfix for me atm.
> 
>  
> @bouthilx bouthilx on 16 Dec 2021 Member
> Oh, I didn't realize it is not used anywhere in this PR. I would propose you move this class and associated tests to another PR and we can look at it in more details when we have time for it.
> 
>  
> @lebrice lebrice 15 days ago Author Contributor
> Fixed: Removed this file and test from this PR.

This refers to the discussion [here](https://github.com/Epistimio/orion/pull/632#discussion_r707550853)

For reference I'm copy-pasting the current state of the discussion as well in case we might as well loose trace.

 
> Member
> @bouthilx bouthilx on 13 Sep 2021
> I think we had a discussion on the topic but I cannot find it. Why would it support different formats for x? Aren't they supposed to be suggestions from the algorithms which are all following the same convention?
> 
>  
> Contributor
> Author
> @lebrice lebrice on 21 Sep 2021
> Being able to pass ndarrays and tensors is useful for testing
> the benchmark class calls task(**params)
> I want these tasks to support receiving one of their dataclasses as an input
> Also, there are tests for most of these cases, and I don't see supporting too many types of inputs as a deal breaker, but up to you.
> 
>  
> Member
> @bouthilx bouthilx on 24 Sep 2021
> Testing like for unit-testing? I honestly do not see why we would want to test the method with other types than what it would be supposed to handle. Supporting other types does not prevent us from supporting typical point format but it does look like feature creep which isn't great either.
> 
> When you say their dataclasses, you mean something like an instance of FcNetTaskHParams? Soon enough we will drop the point array format and use trial objects instead. At that point the benchmarks should follow this transition and support Trials as the input of the call methods. The call method of the tasks is intended to be following the same interface as the functions users would pass to experiment_client.workon() so we need to be careful not to create inconsistencies.
> 
>  
> Contributor
> Author
> @lebrice lebrice on 6 Dec 2021
> Fixed: Made the input type for each task into a TypedDict, (dictionary with type annotations), rather than dataclasses, and the call of ProfetTask now only supports a single type of input.
> 
>  
> Contributor
> Author
> @lebrice lebrice on 6 Dec 2021
> Also customized the __call__ so that Benchmark calling task(**hparam_dict) won't cause an issue.
> 
>  
> Member
> @bouthilx bouthilx on 16 Dec 2021
> It would be simpler to define def call(**x) no? It is a bit weird/inconsistent otherwise, because Oríon expects the functions that are optimized to take the parameters as arguments, not a dict of the parameters.
> 
>  
> Contributor
> Author
> @lebrice lebrice 15 days ago
> I'd currently argue that it isn't. I can explain why in our next 1-1 meeting.
> 
>  
> Member
> @bouthilx bouthilx 14 days ago
> Ok. We should write back a summary of our discussion here for traceability.
> 
>  
> Member
> @bouthilx bouthilx 14 days ago
> I'll try to summarize here. @lebrice Please correct me if I say anything wrong of forget something. So the issue is that the **kwargs input would causing issue with typing linters because child class would define a different interface than **kwargs. I reread the code and I am a bit confused because I don't see child class overwritting the call method.
> 
> An alternative I would see that could be more typing-friendly is to rename call to something like execute and then in each subclass write call(with, the, proper, arguments) and convert them into a dict passed to execute. It does not add value in terms of features, but it makes the code a bit more explicit.
> 
> On the same topic, tasks are expected to rewrite call, not __call__.
> 
>  
> Contributor
> Author
> @lebrice lebrice 16 minutes ago
> Ok, fixed in c3a3846

Issue #735, opened 2021-12-22
Print version in debug verbosity level


bouthilx:
Should be printed somewhere here if verbosity level is debug: https://github.com/Epistimio/orion/blob/develop/src/orion/core/cli/base.py#L64

bouthilx:
Done #765

PR #765, opened 2022-01-19
Print version on debug verbosity level

nurbal:
# Description
Prints Orion version if verbose level is debug.
```
% orion -vv
```

Implements changes requested in issue #735.

# Changes
Quite straitforward. 

# Checklist

## Tests
- [x] All new and existing tests are passing

## Documentation

## Quality

# Further comments
n/a

### DIFF
--- a/src/orion/core/cli/base.py
+++ b/src/orion/core/cli/base.py
@@ -74,7 +74,7 @@ class OrionArgsParser:
             format="%(asctime)-15s::%(levelname)s::%(name)s::%(message)s",
             level=levels.get(verbose, logging.DEBUG),
         )
-        logger.debug("Orion version : %s",orion.core.__version__)
+        logger.debug("Orion version : %s", orion.core.__version__)
 
         if args["command"] is None:
             self.parser.parse_args(["--help"])
--- a/src/orion/core/cli/base.py
+++ b/src/orion/core/cli/base.py
@@ -9,6 +9,7 @@ import sys
 import textwrap
 
 import orion
+import orion.core
 from orion.core.io.database import DatabaseError
 from orion.core.utils.exceptions import (
     BranchingEvent,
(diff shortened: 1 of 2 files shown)


src/orion/core/cli/base.py
```diff
@@ -71,6 +72,8 @@ def parse(self, argv):
             format="%(asctime)-15s::%(levelname)s::%(name)s::%(message)s",
             level=levels.get(verbose, logging.DEBUG),
         )
+        if verbose >= 2:
+            print("Orion version : " + orion.core.__version__)
```
bouthilx, line 76:
```suggestion
            logger.debug("Orion version : " + orion.core.__version__)
```
We should use the logger instead and log at DEBUG level. Normally we create the logger at the top of the module so it is widely available (like [here](https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L33)). There may be an issue here as we change the configuration with logging.basicConfig. We should make sure that this change is reflected on the configuration of the logger created at the top of this module.

bouthilx, line 76:
```suggestion
        logger.debug("Orion version : " + orion.core.__version__)
```
We should use the logger instead and log at DEBUG level. Normally we create the logger at the top of the module so it is widely available (like [here](https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L33)). There may be an issue here as we change the configuration with logging.basicConfig. We should make sure that this change is reflected on the configuration of the logger created at the top of this module.

src/orion/core/cli/base.py
```diff
@@ -71,6 +74,7 @@ def parse(self, argv):
             format="%(asctime)-15s::%(levelname)s::%(name)s::%(message)s",
             level=levels.get(verbose, logging.DEBUG),
         )
+        logger.debug("Orion version : " + orion.core.__version__)
```
bouthilx, line 77: You have an error here: https://github.com/Epistimio/orion/runs/4939592554?check_suite_focus=true#step:5:15
This is because pylint prefers a format with strings using '%s' instead of concatenating them. So it should rather be

```python
logger.debug("Orion version : %s", orion.core.__version__)
```

Sorry I didn't catch this in the first place. :sweat_smile:

bouthilx:
@nurbal The current error is not due to your PR, it is due to some new release of numpydoc. I made a fix in PR #771. I'll restart the tests when the fix is merged.

bouthilx:
PR #771 is merged. I relaunched the tests.

bouthilx (approved):

PR #771, opened 2022-01-26
Replace exp's stat dict by a dataclass

bouthilx:
The custom section of the docstring was causing issues when compiling
the documentation. Also returning a generic dict was not ideal, creating
a dataclass instead provides better documentation and interface for the
users.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -11,6 +11,7 @@ import copy
 import datetime
 import inspect
 import logging
+from dataclasses import dataclass, field
 
 import pandas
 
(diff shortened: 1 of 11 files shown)


src/orion/core/worker/experiment.py
```diff
@@ -25,6 +26,34 @@
...
+    finish_time : `datetime.datetime`
+       When Experiment reached terminating condition and stopped running.
+    duration : `datetime.timedelta`
+       Elapsed time.
+    """
+
+    trials_completed: int
+    best_trials_id: int
+    best_evaluation: float
+    start_time: field(default_factory=datetime.datetime)
```
Delaunay, line 52: `start: datetime.datetime = field(default_factory=datetime.datetime)`

bouthilx: Thanks! 
Done: ffdd4f9569bfeb36406f9953091956af07dc1e6b

Delaunay (approved):

PR #775, opened 2022-02-01
Fix parsing of hps with "." in name from commandline

### LINKED ISSUE #769: Using parameters with a "." in their name causes a crash
See title.

Python 3.8
Orion v0.2.1

```console
orion hunt -n my_exp python main.py --b_optim.lr~'loguniform(1e-6, 0.5, shape=4)' --debug
  File "/home/fabrice/miniconda3/envs/target_prop/bin/orion", line 8, in <module>
    sys.exit(main())
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/cli/__init__.py", line 37, in main
    return orion_parser.execute(argv)
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/cli/base.py", line 89, in execute
    returncode = function(args)
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/cli/hunt.py", line 198, in main
    experiment = experiment_builder.build_from_args(args)
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/experiment_builder.py", line 752, in build_from_args
    return build(**cmd_config)
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/experiment_builder.py", line 175, in build
    config = consolidate_config(name, version, config)
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/experiment_builder.py", line 264, in consolidate_config
    resolve_config.update_metadata(config["metadata"])
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/resolve_config.py", line 303, in update_metadata
    cmdline_parser.set_state_dict(metadata["parser"])
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/orion_cmdline_parser.py", line 134, in set_state_dict
    self.infer_user_script(self.parser.format(self.parser.arguments))
  File "/home/fabrice/miniconda3/envs/target_prop/lib/python3.8/site-packages/orion/core/io/cmdline_parser.py", line 98, in format
    formatted.append(item.format(**configuration))
KeyError: 'b_optim'
```

lebrice:
# Description
Fixes #769 

# Checklist

## Tests

## Documentation
~~- [ ] I have updated the relevant documentation related to my changes~~

## Quality

### DIFF
--- a/src/orion/core/io/cmdline_parser.py
+++ b/src/orion/core/io/cmdline_parser.py
@@ -91,9 +91,26 @@ class CmdlineParser(object):
         # Arguments with spaces are broken if we use ' '.join(template).format().split(' ').
         # Hence we iterate over the list as-is and format on each item separately.
         formatted = []
+        item: str
         for item in self.template:
             if item.startswith("-"):
                 formatted.append(item)
+            elif (
+                item.startswith("{")
+                and item.endswith("}")
+                and "." in item
+                and item.partition(".")[0] not in ["trial", "experiment"]
+            ):
+                # There is an argument, for example "--a.b~uniform(0,1)".
+                for key, value in configuration.items():
+                    if item == f"{{{key}}}":
+                        formatted.append(str(value))
+                        break
+                else:
+                    raise RuntimeError(
+                        f"Unable to find matching value from configuration {configuration} for "
+                        f"template item '{item}'."
+                    )
             else:
                 formatted.append(item.format(**configuration))
 
--- a/src/orion/core/io/cmdline_parser.py
+++ b/src/orion/core/io/cmdline_parser.py
@@ -91,7 +91,6 @@ class CmdlineParser(object):
         # Arguments with spaces are broken if we use ' '.join(template).format().split(' ').
         # Hence we iterate over the list as-is and format on each item separately.
         formatted = []
-        item: str
         for item in self.template:
             if item.startswith("-"):
                 formatted.append(item)
--- a/src/orion/core/io/cmdline_parser.py
+++ b/src/orion/core/io/cmdline_parser.py
@@ -97,19 +97,15 @@ class CmdlineParser(object):
             elif (
                 item.startswith("{")
                 and item.endswith("}")
-                and "." in item
-                and item.partition(".")[0] not in ["trial", "experiment"]
+                and any(item == f"{{{key}}}" for key in configuration)
             ):
-                # There is an argument, for example "--a.b~uniform(0,1)".
-                for key, value in configuration.items():
-                    if item == f"{{{key}}}":
-                        formatted.append(str(value))
-                        break
-                else:
-                    raise RuntimeError(
-                        f"Unable to find matching value from configuration {configuration} for "
-                        f"template item '{item}'."
-                    )
+                # The argument has an entry with exactly matching name in the configuration.
+                # Extract it from the configuration, rather than try to use `str.format`.
+                # This solves bugs that arise from using strings that are invalid python expressions
+                # (e.g. names with ".", "/", or ":").
+                key = [key for key in configuration if item == f"{{{key}}}"][0]
+                value = configuration[key]
+                formatted.append(str(value))
             else:
                 formatted.append(item.format(**configuration))
 
--- a/tests/unittests/core/io/test_orion_cmdline_parser.py
+++ b/tests/unittests/core/io/test_orion_cmdline_parser.py
@@ -27,23 +27,33 @@ def parser_diff_prefix():
     return OrionCmdlineParser(config_prefix="config2", allow_non_existing_files=True)
 
 
+@pytest.fixture(
+    params=[("some/weird:arg_value", "choices({{'a': 0.5, 'b': 0.5}})", "a")]
+)
+def weird_argument_name_prior_and_value(request):
+    """ Fixture that provides a weird name, along with a prior and value. """
+    weird_param_name, prior, value = request.param
+    yield weird_param_name, prior, value
+
+
 @pytest.fixture
-def commandline():
+def commandline(weird_argument_name_prior_and_value):
     """Return a simple commandline list."""
+    weird_name, prior, _ = weird_argument_name_prior_and_value
     return [
         "--seed=555",
         "--lr~uniform(-3, 1)",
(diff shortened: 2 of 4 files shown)


src/orion/core/io/cmdline_parser.py
```diff
@@ -91,9 +91,26 @@ def format(self, configuration):
         # Arguments with spaces are broken if we use ' '.join(template).format().split(' ').
         # Hence we iterate over the list as-is and format on each item separately.
         formatted = []
+        item: str
```
bouthilx, line 94: lost typing :)

src/orion/core/io/cmdline_parser.py
```diff
@@ -91,9 +91,26 @@ def format(self, configuration):
...
         for item in self.template:
             if item.startswith("-"):
                 formatted.append(item)
+            elif (
```
bouthilx, line 98: Is this part still needed for the flatten() in orion_cmdline_parser solves the issue? The generic parser should not need to dig into the configuration to find the item in other nested dict.

src/orion/core/io/cmdline_parser.py
lebrice: Not really, this is intentional: this makes the code below aware that the `item` variable is of type `str`.

src/orion/core/io/cmdline_parser.py
lebrice: Yes, this is still needed:
1. potential `"{trial.hash_id}"`-like interpolations
2. `"{a.b}".format(**{"a.b": 123})` doesn't work, so we need an additional check here.

> The generic parser should not need to dig into the configuration to find the item in other nested dict

I'm not sure I understand the last part, I don't see any digging happening here, would you mind clarifying this a bit?
Was this comment from before the latest updates, when there was still my hacky code in there?

lebrice:
Renamed, since hps aren't hierarchical, they just have a "." in their name.

tests/unittests/core/io/test_orion_cmdline_parser.py
```diff
@@ -2,8 +2,11 @@
 # -*- coding: utf-8 -*-
 """Example usage and tests for :mod:`orion.core.io.orion_cmdliner_parser`."""
 import os
+import textwrap
+from typing import List
```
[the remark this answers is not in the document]
lebrice: Fixed in 501c3a91ae34f168116e80a65064b2d389754208

src/orion/core/io/cmdline_parser.py
bouthilx: It's more clear thanks! The issue is that `format()` tries to assign a variable `a` with attribute `b`. This problem comes from the fact that the key value is defined by the user, and it could occur with other characters than `.`.

We would also need to adapt the command here for the tests of CmdlineParser to verify such cases: https://github.com/Epistimio/orion/blob/develop/tests/unittests/core/io/test_cmdline_parser.py#L53.

I think the best would be to shield CmdlineParser from weird user argument and use an internal dict mapping from user keys to safe keys, so that the templates we build during `CmdlineParser.parse()` only contain safe keys (no invalid characters for `format()`). That would require building the mapping dict during `parse()`, handle backward compatibility in `state_dict`/`set_state_dict` and replace configuration keys during `format()`. Another solution would be to use curated keys, for which we would remove the problematic characters, but then we risk causing collisions if the user had arguments like --ab and --a.b in the same commandline call.

That's a lot more work so maybe it can be delayed after the sprint?

src/orion/core/io/cmdline_parser.py
bouthilx: I meant that it is a type annotation in a non-typed module.

lebrice: My interpretation of your "type/no-type" directive was that it was about the public-facing API (methods signatures, etc).
There is really no harm in leaving this in, is there?

src/orion/core/io/cmdline_parser.py
bouthilx: Some other characters will lead to the same issue, like `--some:weird arg-name`. Clearly `.` will be the most common source of issue, but the problem we have with it is a symptom of a more general problem.

lebrice: Alright, but this PR is a simple fix for a simple bug, I think it's best if we tackle other issues of this type in another PR, wouldn't you agree?

src/orion/core/io/cmdline_parser.py
lebrice: I see. I have a different opinion on this: I think incremental addition of typing is fine, and I'd say that adding typing inside methods or internal functions is a useful first step towards modernizing the codebase. I'll remove this for now, just so you know, my opinion is that we should embrace these small quality-of-life changes, rather than dis-incentivize them.

lebrice: Fixed in 8541015d207f0293ee5894f53bf8fbe7e52e9539

src/orion/core/io/cmdline_parser.py
bouthilx: It's an adhoc fix for one symptom of a bigger problem. Another PR would aim at fixing the same issue, but at the root of it.

bouthilx: It's not super urgent however, so maybe a temporary plaster solution is better than waiting to solve the main source of the issue. We can keep trace of it in an issue.

bouthilx: Well, looks like you fixed it all! :)

bouthilx (approved):
Thanks a lot for this fix!

src/orion/core/io/cmdline_parser.py
lebrice: Fixed in  5f3d3b840f477524acb8fec70b4b1a5db565416b: Made the check more general, now works with a whole variety of weird argument names, not just those with ".".

Issue #709, opened 2021-12-09
Improve error message when user script fails


bouthilx:
The current error message is misleading: `orion.core.worker.consumer.ExecutionError: Something went wrong. Check logs. Process returned with code -9 !`

First, there are no clear mentions that the issue is that the user's script crashed. This should be clearly mentioned. Second, _Check logs_ sounds like it Oríon was writting some logs somewhere. It is not the case. The error message should only state the the user's script failed, and the user will know that it should look at the output of the script to see if there is an error message. 

On the same topic, is the error message from the script printed in the terminal? For example if the user's script is written in Python, would the stack-trace be printed? It should be.

bouthilx:
Error message was changed in PR #684. We would need to verify that the new error message is useful and adapt it if necessary. Otherwise most important thing is to verify that the call to [subprocess.Popen](https://github.com/Epistimio/orion/blob/develop/src/orion/core/worker/consumer.py#L255) is not catching all stdin/stderr so that the script's stack-trace is visible in the terminal for the user.

PR #777, opened 2022-02-02
added test_user_script_crash test (#709)

nurbal:
# Description
Detects if a by-design crashing black-box script Traceback is visible in the stderr output of orion.

The test succeeds, which indicates that the stacktrace is indeed visible by the user on his terminal output.

Is meant to implement issue #709 

# Changes
Just a test, since the behaviour is already good.
Added a "black box" utility script that can return a code passed by argument, and/or crash if -c argument is specified.

# Checklist

## Tests

## Documentation

## Quality

# Further comments
n/a

### DIFF
--- a/tests/functional/commands/test_hunt_command.py
+++ b/tests/functional/commands/test_hunt_command.py
@@ -35,3 +35,26 @@ def test_no_name(capsys):
     captured = capsys.readouterr().err
 
     assert captured == "Error: No name provided for the experiment.\n"
+
+
+def test_user_script_crash(capfd):
+    """Checks that the Traceback of a crashing user script is output to stderr"""
+
+    orion.core.cli.main(
+        [
+            "--debug",
+            "hunt",
+            "-n",
+            "test",
+            "--exp-max-trials",
+            "1",
+            "./black_box_fail.py",
+            "-c",
+            "-x~uniform(-5,5)",
+        ]
+    )
+
+    out, err = capfd.readouterr()
+
+    assert "Traceback" in err
+    assert "# black_box_fail.py : this line should crash" in err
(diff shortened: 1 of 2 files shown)


tests/functional/commands/test_hunt_command.py
```diff
@@ -35,3 +35,26 @@ def test_no_name(capsys):
     captured = capsys.readouterr().err
 
     assert captured == "Error: No name provided for the experiment.\n"
+
+
+def test_user_script_crash(capfd):
+    """Checks that the Traceback of a crashing user script is output to stderr"""
+
+    orion.core.cli.main(
```
bouthilx, line 43: The test if failing because the script is not found. Pytest is executed from the root so it cannot find this file at the root. You can use monkeypatch to change the working dir temporarily: https://github.com/Epistimio/orion/blob/develop/tests/functional/commands/conftest.py#L61

Note, the first test of this module is not crashing because it has the fixture `one_experiment` which does this change of working directory.

tests/functional/commands/test_hunt_command.py
```diff
@@ -35,3 +35,26 @@ def test_no_name(capsys):
...
+        [
+            "--debug",
+            "hunt",
+            "-n",
+            "test",
+            "--exp-max-trials",
```
bouthilx, line 49:
```suggestion
            "--exp-max-broken-trials",
```

tests/functional/commands/test_hunt_command.py
```diff
@@ -35,3 +35,26 @@ def test_no_name(capsys):
...
+            "1",
+            "./black_box_fail.py",
+            "-c",
+            "-x~uniform(-5,5)",
+        ]
+    )
```
bouthilx, line 55: You could test the exit-code at the same time. The `main` should return it.

PR #778, opened 2022-02-02
Add falcon-cors requirement to make orion server accept CORS requests

notoraptor:
# Description

Hi, @bouthilx ! I suggest this pull request to make orion server support CORS (cross-origin resource sharing) requests.

Currently, Orion will reject any API call made from a server located in a different address than himself. For example, if orion server is at `http://myorionserver.com`, and I make an API call from a local server (`http://localhost`), then API call won't work because servers addresses are different.

# Changes

This PR adds and uses Python package `falcon-cors`. I guess an orion server should be public, so `falcon-cors` currently makes server accepts requests from any other server. But, if necessary, we could later restrict allowed requests to a specific list of addresses.

I need this PR to make `orion-dashbord` API calls work correctly.

# Checklist

## Tests

## Documentation

## Quality

### DIFF
--- a/src/orion/serving/webapi.py
+++ b/src/orion/serving/webapi.py
@@ -9,6 +9,7 @@ Exposes a WSGI REST server application instance by subclassing ``falcon.API``.
 """
 
 import falcon
+from falcon_cors import CORS
 
 from orion.serving.experiments_resource import ExperimentsResource
 from orion.serving.plots_resources import PlotsResource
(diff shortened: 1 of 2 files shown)


src/orion/serving/webapi.py
```diff
@@ -24,7 +25,17 @@ class WebApi(falcon.API):
...
+        # By default, server will reject requests coming from a server
+        # with different origin. E.g., if server is hosted at
+        # http://myorionserver.com, it won't accept an API call
+        # coming from a server not hosted at same address
+        # (e.g. a local installation at http://localhost)
+        # This is a Cross-Origin Resource Sharing (CORS) security:
+        # https://developer.mozilla.org/fr/docs/Web/HTTP/CORS
+        # To make server accept CORS requests, we need to use
+        # falcon-cors package: https://github.com/lwcolton/falcon-cors
+        cors = CORS(allow_all_origins=True)
```
bouthilx, line 37: Shouldn't we use a safe default with just localhost:8000? I am worried that the default may cause security issues. Or maybe CORS is secure enough? I am not familiar with it so I do not know.

bouthilx (approved):
Thanks!

PR #780, opened 2022-02-04
Add missing falcon-cors dependency for conda

bouthilx:


notoraptor:
HI @bouthilx ! It seems `falcon-cors` package is available only on PyPI.

bouthilx:
oh! I'm not super knowledgeable with conda, do you know how we could fix this?

PR #810, opened 2022-02-22
Tests run with no extras and all extras

lebrice:


bouthilx:
Thanks for the PR! I think the way you handled this in `tox.ini` is great! There are some fixes that will be required in the code. Namely the configuration test that rely on dask should be changed to singleexecutor. 

github-action build.yaml should be changed to add a test running with no dependency. I'd say with python 3.9 version. And the current matrix execution for python 3.6 to 3.9 should be executed with all.

bouthilx (approved):
All good! Thanks a lot for this PR!

Issue #802, opened 2022-02-15
Rework how transformed space is passed to algorithms


bouthilx:
The class `SpaceTransformAlgoWrapper` creates a transformed space and pass it to the algorithm. It builds the algorithm with the original space however, and then assigns the original space. This is partly due to the fact that it was previously necessary to build the algorithm before we can access the attributes `requires_{type, shape, dist}`, but this is not necessary anymore with the new Factory introduced in v0.2.0. The other reason is that many algorithms require access to the original space to verify if they are done because this depends on the cardinality of the space, which is lost in transformed space which we convert to Real dimensions. 

The first issue is already addressed and would only require minor refactoring across the algorithms. 

We need to address the second issue. @lebrice suggests that we pull up the `is_done` verification based on cardinality (and thus the original space) into the `SpaceTransformAlgoWrapper`. The other part that would need to be addressed is format_trial and get_trial that depends on the transformed_space to do a transform(reverse(trial)) to get the trial id in the original space. This should not be inside the algorithm itself, the algorithm should be totally impervious to whether the space is the original one or a transformed one, as long as the dimension types are compatible with the requirements of the algorithm. One way to solve this issue would be to extract all concepts of trial registration into a simple in-memory storage that we pass to the algorithm. This storage would then be responsible to taking care of any transformation required. The algorithm would still call the same register, has_suggested, has_observed methods but they would be part of the special algo storage instead of being part of the algorithm.

PR #813, opened 2022-02-24
Test algo can handle broken trials

bouthilx:
When the trial is broken it should be considered as such by the
algorithm, it is not pending anymore.

algo.has_observed(broken_trial) should return True.

### DIFF
--- a/src/orion/algo/base.py
+++ b/src/orion/algo/base.py
@@ -310,9 +310,11 @@ class BaseAlgorithm:
             True if the trial's objective was observed by the algo, False otherwise.
 
         """
-        return (
-            self.get_id(trial) in self._trials_info
-            and self._trials_info[self.get_id(trial)][1] is not None
+        if not self.has_suggested(trial):
+            return False
+        return self._trials_info[self.get_id(trial)][0].status in (
+            "broken",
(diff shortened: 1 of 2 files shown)


src/orion/algo/base.py
```diff
@@ -310,9 +310,11 @@ def has_observed(self, trial):
             True if the trial's objective was observed by the algo, False otherwise.
 
         """
-        return (
-            self.get_id(trial) in self._trials_info
-            and self._trials_info[self.get_id(trial)][1] is not None
+        if not self.has_suggested(trial):
+            return False
+        return self._trials_info[self.get_id(trial)][0].status in (
```
lebrice, line 315: Why does this check only the first index? Is it always a len(1) sequence? If there were to be duplicates, should this do anything different?

bouthilx: It's a bad design. I'd like to address this with a more formal in-memory storage for algorithms are shortly described here: https://github.com/Epistimio/orion/issues/802. 

The format is
 ```python
self._trials_info[some_trial_id] = (trial_object, trial_objective)
```

Previously we only looked at the objective at index 1, now we only look at the trial at index 0.

PR #814, opened 2022-03-01
Typing improvements

lebrice:
General typing improvements.


lebrice:
Will need to reassess each of these commits after #833  is merged. Will reopen a cherry-picked version of this branch at that point.

PR #832, opened 2022-03-16
Add configspace conversion

Delaunay:

# Description

# Changes

# Checklist

## Tests

## Documentation

## Quality

# Further comments


lebrice (commented):
Cool stuff!
Can you try using [singledispatch callables](https://docs.python.org/3/library/functools.html#functools.singledispatch) for the to/from conversion functions?
It would be a lot more intuitive than using the Visitor pattern, since your visitor doesn't have any state anyway: 
```python
from ConfigSpace import ConfigurationSpace
from orion.algo.space import Space
from functools import singledispatch
# ConfigSpace -> Orion
@singledispatch
def to_orion(config_space: ConfigSpace) -> Space:
    """ converts the given config space to an Orion space. """
    raise NotImplementedError(f"Don't have a handler for converting spaces of type {type(config_space)}.")

@to_orion.register
def _convert_uniform_float(config_space: UniformFloatHyperparameter) -> Real:
     ...
@to_orion.register
def _convert_foobar(config_space: FooBarHyperParameter) -> Dimension:
    ...  
 ... # other handlers
 
 # Same thing for Orion -> ConfigSpace
@singledispatch
def to_configspace(orion_space: Space) -> ConfigurationSpace:
    """ Converts the given Orion thingy to a ConfigSpace thingy. """
    raise NotImplementedError(f"Don't know how to convert {orion_space} to an equivalent from ConfigSpace")

@to_configspace.register
def _real_to_uniform(dim: Real) -> UniformFloatHyperparameter:
     ... # convert from Orion to ConfigSpace
```
Another major benefit is that we could then extend the conversions externally (even from within ConfigSpace if you want!) by just registering new handlers for new dimensions when they are added.

Delaunay:
That specific visitor does not have a states, future one could have.
I dont think single dispatch improve the code.

lebrice:
> That specific visitor does not have a states, future one could have.

What kind of state could it have?

> I dont think single dispatch improve the code.

I'm curious: having considered both, you say that you prefer the Visitor to singledispatch, correct? So what advantages do you see in the Visitor pattern vs dispatching?

From my experience, this kind of `if instance(thing, A): do_something() elif isinstance(thing, B): do_something_else()` is exactly what singledispatch is great at simplifying. If you take out the Visitor thing, all you're left with is two functions, `to_orion` and `to_configspace` (your test is also a pretty clear indication of this!), and inside these functions will be a whole bunch of `if isinstance` statements. What `singledispatch` does is to then break up these two functions into smaller, more specific functions. The two generic functions can then be extended later by adding new handlers for new object types as needed.

Delaunay:
The visitor does not use `isinstance` though, only the to_orion conversion use it.

``to_orion``  uses it because Uniform is split between Float & Integer which are converted the same way except for the class name.
If I create a function for each type I will duplicate all that code twice.
So the code with singledispatch will end up being longer and more repetitive.

* Visitor could have a state if the conversion process had to go through a ``SpaceBuilder`` instead.

* Having a public interface allow us to define clearly how users are expected to extend Orion Space (We are planning on adding conditionals, so that is more nodes to visit). Users can look at the visitor is see clearly all the methods he is expected to override to implement a valid feature.

NB: Visitor are more than just for the conversion they can be used to implement any new features outside of ``Dimensions``
Users could implement their own sample/serialization/... using a Visitor.

lebrice:
> If I create a function for each type I will duplicate all that code twice.

I forgot to mention in my initial example that you can use the same handler for more than one type, by cascading the `@to_orion.register(some_type)`. Edited the example above to show this a bit.
Idk if this changes your mind or not, but in any case, just thought you might want to know, since it would indeed really suck if you could only have one type per handler. This way, there is no need to duplicate a bunch of code between handlers.

I like how you made the conversion functions singledispatch callables, that might come in handy in the future!

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,222 @@
...
+
+        if dim.lower:
+            kwargs["low"] = dim.lower
+            kwargs["high"] = dim.upper
+
+    return klass(dim.name, dist, *args, **kwargs)
+
+
+@to_orionspace.register
+def configpsace_to_orionspace(cspace: ConfigurationSpace) -> Space:
```
lebrice, line 208:
```suggestion
def configspace_to_orionspace(cspace: ConfigurationSpace) -> Space:
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
...
+    Notes
+    -----
+    ``ConfigurationSpace`` will set default values for each dimensions of ``Space``
+
+    """
+    space = Space()
+
+    for _, cdim in cspace.get_hyperparameters_dict().items():
+        odim = to_oriondim(cdim)
+        print(cdim, odim)
```
bouthilx, line 257:
```suggestion
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
...
+        assert odim is not None
```
bouthilx, line 258: Should raise a ValueError() with an explicit error message.

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
...
+    def fidelity(self, dim: Fidelity) -> None:
+        """Ignores fidelity dimension as configspace does not have an equivalent"""
+        return None
+
+    def space(self, space: Space) -> ConfigurationSpace:
+        """Convert orion space to configspace"""
+        cspace = ConfigurationSpace()
+        dims = []
+
+        for _, dim in space.items():
```
bouthilx, line 151: I think this part could be a generic convert_dimensions method that we do not need to reimplement for each visitors. On a similar topic, I know this deviates from the Visitor pattern but I think a single method `convert_dimension` like 
```
class Visitor:
    ...
    def convert_dimension(self, dimension: Dimension)
        return getattr(self, _to_snake_case(dimension.__name__))(dimension)
```
Also, maybe a class name along the lines of `DimensionConverter` or `SpaceConverter` would be more explicit. To be fair, this would start to be very close to singledispatch. I agree however that the class structure is helpful for contributors to understand what they need to cover.

Delaunay: But a visitor is not about converting the space, we are using it to do so, but it is a completely generic abstraction that has nothing to do with converting the dimension, that is why the name of the methods are non specific because it is just about visiting the tree nodes.

You could use the visitor to implement any of the method defined inside the Dimension class and subclasses

tests/unittests/algo/test_configspace.py
```diff
@@ -0,0 +1,46 @@
...
+    space.register(Real("u2f", "uniform", -3, 6))
+    space.register(Real("name.u2f", "uniform", -3, 6))
+
+    space.register(Categorical("c1", ("asdfa", 2)))
+    space.register(Categorical("c2", dict(a=0.2, b=0.8)))
+    space.register(Fidelity("f1", 1, 9, 3))
+
+    space.register(Real("n1", "norm", 0.9, 0.1, precision=6))
+    space.register(Real("n2", "norm", 0.9, 0.1, precision=None))
+    space.register(Real("n3", "norm", 0.9, 0.1))
```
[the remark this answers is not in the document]
Delaunay: which non supported priors ?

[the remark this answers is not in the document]
Delaunay: `Fidelity` is there

src/orion/algo/space/configspace.py
Delaunay: Should it ? If the dimension is not supported `NotImplementedError` will be raised.
`to_oriondim()` cannot return None

src/orion/algo/space/configspace.py
bouthilx: I'm worried that we are using an abstraction that will not be used for something else than conversions and that makes the implementation of conversions less intuitive for external contributors. If there is a relatively clear potential for other usages of this abstraction then I think it's fine, but right now I don't see any. What would you have in mind for other uses?

tests/unittests/algo/test_configspace.py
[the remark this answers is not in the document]
Delaunay: The behaviour will be like fidelity, I added the alpha prior for completeness,
& also added a test for the unsupported priors from configspace

src/orion/algo/space/configspace.py
lebrice: Yes I'm not asking that. I'm asking, why is the `odim` asserted to not be None, if you're saying that it can never be None?
Could it be because you use something like `Optional[Dimension]` as the return annotation of `to_oriondim`? If so, then the annotation should be `Dimension`, and the assert isn't necessary.

tests/unittests/algo/test_configspace.py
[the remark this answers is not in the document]
bouthilx: Should we make the conversion just ignore unsupported priors? For the fidelity we know it makes sense to drop it because the algorithms are expected to handle it differently. For the other priors I would be worried that the code crash at some latter weird places, like during wrapper's suggest() trial transformation. That would be hard to understand for the user compared to a crash during space conversion.

src/orion/algo/space/configspace.py
Delaunay: It documents the invariant that odim cannot be none, i.e "stating the fact" as well as checking it is

tests/unittests/algo/test_configspace.py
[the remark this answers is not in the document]
Delaunay: updated

src/orion/algo/space/configspace.py
lebrice: I'm a huge proponent of asserts btw, unlike @bouthilx  that has a bit of Assertophobia :P

There is no need to state this fact here. This doesn't add any information that wasn't already encoded by the type hints. You can remove this assert.

lebrice: Asserts don't check things. They affirm things. IF `odim` could possibly be `None`, that's different. You'd want to do as @bouthilx  says and raise a ValueError. However, your type-signature for `to_oriondim` already indicates it can't be None. The assert here is not needed.

src/orion/algo/space/configspace.py
Delaunay: updated

src/orion/algo/space/configspace.py
Delaunay: fixed

bouthilx: @lebrice he-hum, Assertinmaincodeophobia. :stuck_out_tongue:

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@singledispatch
+def to_oriondim(dim: Hyperparameter) -> Dimension:
+    """Convert a config space hyperparameter to an orion dimension"""
+    raise NotImplementedError(f"Dimension {dim} is not supported by Orion")
+
+
+@to_oriondim.register
+def from_categorical(dim: CategoricalHyperparameter) -> Dimension:
```
lebrice, line 183: The handlers that you register for specific types should have a name that starts with an underscore, to indicate that they shouldn't be used by the end-users, in favour of the the general `to_oriondim` function.

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
...
+
+    if dim.log:
+        dist = "reciprocal"
+
+    return klass(dim.name, dist, *args, **kwargs)
+
+
+@to_oriondim.register(NormalFloatHyperparameter)
+@to_oriondim.register(NormalIntegerHyperparameter)
+def from_normal(dim: Hyperparameter) -> Dimension:
```
lebrice, line 219: Nit: Narrowing the type signature:
```suggestion
def from_normal(dim: Hyperparameter) -> Integer | Real:
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,261 @@
+from functools import singledispatch
```
lebrice, line 1:
```suggestion
from __future__ import annotations
from functools import singledispatch
```

src/orion/algo/space/__init__.py
```diff
@@ -31,6 +31,8 @@
 import copy
```
lebrice, line 31:
```suggestion
from __future__ import annotations
import copy
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +69,64 @@ def __repr__(self):
...
+
+
+class Visitor(Generic[T]):
+    """Visitor interface to iterate over an Orion search space.
+    This can be used to implement new features for ``orion.algo.space.Space``
+    outside of Orion's code base.
+
+    """
+
+    def visit(self, dim: "Dimension") -> T:
```
lebrice:
```suggestion
    def visit(self, dim: Dimension) -> T:
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +69,64 @@ def __repr__(self):
...
+        """Make dimension call its handler"""
+        return dim.visit(self)
+
+    def dimension(self, dim: "Dimension") -> T:
```
lebrice, line 104:
```suggestion
    def dimension(self, dim: Dimension) -> T:
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +69,64 @@ def __repr__(self):
...
+        """Called when the dimension does not have a decicated handler"""
+        pass
+
+    def real(self, dim: "Real") -> T:
```
lebrice, line 108:
```suggestion
    def real(self, dim: Real) -> T:
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +69,64 @@ def __repr__(self):
...
+        """Called by real dimension"""
+        pass
+
+    def integer(self, dim: "Integer") -> T:
+        """Called by integer dimension"""
+        pass
+
+    def categorical(self, dim: "Categorical") -> T:
```
lebrice, line 116:
```suggestion
    def categorical(self, dim: Categorical) -> T:
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +69,64 @@ def __repr__(self):
...
+        """Called by categorical dimension"""
+        pass
+
+    def fidelity(self, dim: "Fidelity") -> T:
+        """Called by fidelity dimension"""
+        pass
+
+    def space(self, space: "Space") -> None:
```
lebrice, line 124:
```suggestion
    def space(self, space: Space) -> None:
```

tests/unittests/algo/test_configspace.py
```diff
@@ -0,0 +1,70 @@
...
+        # Orion space did not have default values
+        # but ConfigSpace always set them
+        if not original.default_value:
+            converted._default_value = None
+
+        assert type(original) == type(converted)
+        assert original == converted
+
+
+@pytest.mark.skipif(IMPORT_ERROR, reason="Running without ConfigSpace")
```
lebrice, line 51: Use `pytest.mark.skip(condition=IMPORT_ERROR, reason="Tests need ConfigSpace", allow_module_level=True)` at the top of the test script instead of having of marking out all the individual tests.

lebrice (changes requested):

src/orion/algo/space/configspace.py
bouthilx: If not using the visitor pattern then I think there is no need to adding the `convert_dimension` to Dimension, this could be part of the base SpaceConverter.

Delaunay: it already does not have it

src/orion/algo/space/configspace.py
Delaunay: I actually like that they can be imported by users

Delaunay: updated anyway

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def categorical(self, dim: Categorical) -> CategoricalHyperparameter:
+        """Convert a categorical dimension into a configspace equivalent"""
+        return CategoricalHyperparameter(
+            name=dim.name,
+            choices=dim.categories,
+            weights=dim._probs,
+        )
+
+    def fidelity(self, dim: Fidelity) -> None:
```
lebrice, line 145: This is incoherent with the type signature of the `SpaceConverter[HyperParameter]` class.

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +71,64 @@ def __repr__(self):
...
+    if frag:
+        frags.append("".join(frag).lower())
+
+    return "_".join(frags)
+
+
+T = TypeVar("T")
+
+
+class SpaceConverter(Generic[T]):
```
lebrice, line 95: I don't really see the point of having a `SpaceConverter`. Is it to force an external library to provide conversion functions functions for the given types of spaces?

Also, this generic signature isn't really respected by the `ToConfigSpace` class, right? Since it doesn't support `Fidelity`-type spaces, it returns `None` in that case, and so it becomes a `SpaceConverter[Optional[HyperParameter]]`!

src/orion/algo/space/configspace.py
bouthilx: Oh right, I missred the diff. Sorry about that!

lebrice (changes requested):

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+@singledispatch
+def to_oriondim(dim: Hyperparameter) -> Dimension:
+    """Convert a config space hyperparameter to an orion dimension"""
+    raise NotImplementedError(f"Dimension {dim} is not supported by Orion")
+
+
+@to_oriondim.register
+def _from_categorical(dim: CategoricalHyperparameter) -> Dimension:
```
lebrice, line 184: Nit:
```suggestion
def _from_categorical(dim: CategoricalHyperparameter) -> Categorical
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+    """Builds a categorical dimension from a categorical hyperparameter"""
+    choices = {k: w for k, w in zip(dim.choices, dim.probabilities)}
+    return Categorical(dim.name, choices)
+
+
+@to_oriondim.register(UniformIntegerHyperparameter)
+@to_oriondim.register(UniformFloatHyperparameter)
+def _from_uniform(dim: Hyperparameter) -> Dimension:
```
lebrice, line 192: Nit:
```suggestion
def _from_uniform(dim: Hyperparameter) -> Integer | Real:
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+    """Convert from orion space to configspace
+
+    Notes
+    -----
+    ``ConfigurationSpace`` will set default values for each dimensions of ``Space``
+
+    """
+    space = Space()
+
+    for _, cdim in cspace.get_hyperparameters_dict().items():
```
lebrice, line 258: Nit:
```suggestion
    for cdim in cspace.get_hyperparameters_dict().values():
```

tests/unittests/algo/test_configspace.py
```diff
@@ -0,0 +1,70 @@
+import pytest
+
+from orion.algo.space import Categorical, Fidelity, Integer, Real, Space, to_orionspace
+from orion.algo.space.configspace import IMPORT_ERROR, UnsupportedPrior, to_configspace
+
+pytest.mark.skip(
+    IMPORT_ERROR, reason="Running without ConfigSpace", allow_module_level=True
+)
```
bouthilx, line 8:
```suggestion
if IMPORT_ERROR:
    pytest.skip(
        "Running without ConfigSpace", allow_module_level=True
    )
```

src/orion/algo/space/__init__.py
```diff
@@ -67,6 +71,64 @@ def __repr__(self):
...
+    """SpaceConverter iterates over an Orion search space.
+    This can be used to implement new features for ``orion.algo.space.Space``
+    outside of Orion's code base.
+
+    """
+
+    def convert_dimension(self, dimension: Dimension) -> T:
+        """Call the dimension conversion handler"""
+        return getattr(self, _to_snake_case(type(dimension).__name__))(dimension)
```
lebrice, line 104: This is essentially a singledispatch callable in disguise: You have handlers for each type below, and you dispatch based on the "type" of the argument, indirectly.
However, having an implicit dependency on the name of the dimension classes is very brittle, and hard to keep track of. Could you make this more explicit somehow?
I also don't think this should be inside a class, since there isn't (and most likely won't be) any state.

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+            name=dim.name,
+            choices=dim.categories,
+            weights=dim._probs,
+        )
+
+    def fidelity(self, dim: Fidelity) -> None:
+        """Ignores fidelity dimension as configspace does not have an equivalent"""
+        return None
+
+    def space(self, space: Space) -> ConfigurationSpace:
```
lebrice, line 149: Move this logic to a `Space` handler of `to_configspace`:
```python
@to_configspace.register(Space)
<this code>
```

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+def to_configspace(space: Space) -> ConfigurationSpace:
+    """Convert orion space to configspace
+
+    Notes
+    -----
+    ``ConfigurationSpace`` will set its own default values
+    if not set inside ``Space``
+
+    """
+    conversion = ToConfigSpace()
```
lebrice, line 173: This should be the "root" of the singledispatch callable, and just return a `NotImplementedError` for the given argument.

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+        for _, dim in space.items():
+            cdim = self.convert_dimension(dim)
+
+            if cdim:
+                dims.append(cdim)
+
+        cspace.add_hyperparameters(dims)
+        return cspace
+
+
```
lebrice, line 163: Make this a singedispatch callable, and copy-paste the methods of the `ToConfigSpace` class into handler functions for each supported dimension type below. 
```suggestion
@singledispatch
```

tests/unittests/algo/test_configspace.py
```diff
@@ -0,0 +1,69 @@
...
+    space.register(Categorical("c1", ("asdfa", 2)))
+    space.register(Categorical("c2", dict(a=0.2, b=0.8)))
+    space.register(Fidelity("f1", 1, 9, 3))
+
+    space.register(Real("n1", "norm", 0.9, 0.1, precision=6))
+    space.register(Real("n2", "norm", 0.9, 0.1, precision=None))
+    space.register(Real("n3", "norm", 0.9, 0.1))
+    space.register(Integer("n4", "norm", 1, 2))
+
+    newspace = to_configspace(space)
```
lebrice, line 32: This to me is a very clear sign that you only need a function.

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+def _qantization(dim: Dimension) -> float:
+    """Convert precision to the quantization factor"""
+    if dim.precision:
+        return 10 ** (-dim.precision)
+    return None
+
+
+class ToConfigSpace(SpaceConverter[Hyperparameter]):
```
lebrice, line 62: Remove, see below.

src/orion/algo/space/__init__.py
lebrice: Also, what happens if a new type of dimension is added, and there is no method with that name? This will raise an `AttributeError`, but you'd ideally like a `NotImplementedError` with an informative message.

lebrice (changes requested):

src/orion/algo/space/__init__.py
lebrice: bump

src/orion/algo/space/configspace.py
```diff
@@ -0,0 +1,262 @@
...
+    def fidelity(self, dim: Fidelity) -> None:
+        """Ignores fidelity dimension as configspace does not have an equivalent"""
+        return None
+
+    def space(self, space: Space) -> ConfigurationSpace:
+        """Convert orion space to configspace"""
+        cspace = ConfigurationSpace()
+        dims = []
+
+        for _, dim in space.items():
```
lebrice, line 154:
```suggestion
        for dim in space.values():
```

tests/unittests/algo/test_configspace.py
```diff
@@ -0,0 +1,69 @@
+import pytest
```
lebrice, line 1: Also, it could probably be useful to also test the transformed dimensions `TransformedDimension`, `ReshapedDimension`, etc. that are passed to the constructors of the algorithms.

If you choose to support these kind of dimensions, you can use this kind of pattern to get the delegate function based on the type of the wrapped dimension, instead of the type of the wrapper:

```python
@to_configspace.register(TransformedDimension)
def _transformed_to_configspace(dim: TransformedDimension) -> HyperParameter:
    unwrapped_dim = dim.original_dimension
    # Get the handler for the unwrapped dim:
    handler_fn = to_configspace.dispatch(type(unwrapped_dim))    
    converted_unwrapped_dim = handler_fn(unwrapped_dim)
    # do something with it? Like re-add the transformation somehow?
    return converted_unwrapped_dim  # just drop the transformation for now.
```

lebrice:
> The visitor does not use `isinstance` though, only the to_orion conversion use it.

It actually does, indirectly, since it dispatches based on the name of the type. This is equivalent to doing `isinstance`.

src/orion/algo/space/__init__.py
lebrice: > Users can look at the visitor and see clearly all the methods he is expected to override to implement a valid feature.

I guess this answers my first question. However, it's clear from the `fidelity` example that you actually don't want to enforce that every type of dimension has to be supported by subclasses. Therefore, it's not actually an interface, if all its members are optional. Therefore it's only use, as far as I can see, is to be a list of names of methods that "can" be implemented by subclasses.

> NB: Visitor are more than just for the conversion they can be used to implement any new features outside of Dimensions 
Users could implement their own sample/serialization/... using a Visitor.

At this point, why not just let the user code create a function that does what they want to do, e.g. serialize / print / do something for the types of `Dimension` of Orion they want to support?

bouthilx: To be fair the fidelity is the only exception. I don't see how the conversions could be usable if they do not support all other type of dimensions. I think the abstract class is the clearest way to express the interface expected for contributors. The only thing needed would be to clarify the exception for the fidelity.

src/orion/algo/space/__init__.py
bouthilx: If we make it an abstract class then non-implemented methods would already raise an informative message during code development. It's true there is no state so making it a class isn't super useful for functionality's sake but I think it is nevertheless useful to make the interface more clear. The other solution would be to have a module for these functions only (without the methods for the conversion the other way around).

In any case, we should make it such that if the dev forgets to write some methods it fails with a clear message at execution or some tests automatically detects this.

tests/unittests/algo/test_configspace.py
bouthilx: Looks like most of the dispatch would be done this way since algorithms receives transformed space. The use of strings to map the conversion methods may be more convenient than dispatch because of this.

lebrice: That's incorrect: https://github.com/Epistimio/orion/pull/832/files#diff-eec91bf300bfcaf8b606f6046f67669737948168c0e7a6b00133c91965f5ddd1R104

The name of the class is used for dispatching. The issue is the same.

src/orion/algo/space/__init__.py
bouthilx: Actually, this should use the property `type` to work with transformed dimensions. Looks like we should also have tests for converting transformed spaces. 
```suggestion
    def convert_dimension(self, dimension: Dimension) -> T:
        """Call the dimension conversion handler"""
        return getattr(self, _to_snake_case(dimension.type))(dimension)
```

tests/unittests/algo/test_configspace.py
bouthilx: Indeed: https://github.com/Epistimio/orion/pull/832#discussion_r849864754

src/orion/algo/space/__init__.py
lebrice: As long as:
- we don't dispatch based on the names of classes, and 
- the solution is consistent with itself and the base class,
then I don't have a problem with using an ABC and marking all the methods with @abstractmethod.

lebrice: > It's true there is no state so making it a class isn't super useful for functionality's sake but I think it is nevertheless useful to make the interface more clear. The other solution would be to have a module for these functions only (without the methods for the conversion the other way around).

Since the abstract base class would only have static methods, this is actually a good use-case for [Protocols](https://docs.python.org/3/library/typing.html#typing.Protocol):
```python
from typing_extensions import Protocol

T = TypeVar("T")

class SpaceConverter(Protocol[T]):
     integer: Callable[[Integer], T]
     real: Callable[[Real], T]
     fidelity: Callable[[Fidelity], T | None]
```

This way, if you have a module like so:
```python
# nevergrad/conversion.py (or something like that)

def integer(dim: Integer) -> NeverGradStuff:
     ...
    
def real(dim: Real) -> NeverGradReal:
    ...
```

The procol can be used just like the abstract base class, and has the benefit of allowing type checkers to also check against a module:
```python
from nevergrad import conversion

space_converter: SpaceConverter[NeverGradStuff] = conversion    # Passes, since all the required methods are defined.
 ```

lebrice (commented):
I think some issues will pop up with transformed spaces, it would be a good idea to add some tests for those.

Feel free to disregard my other comments about the design. If something comes up, I'll make a PR.

bouthilx:
This version of the conversion would be sufficient for HPOBench that @donglinjy is working on. I will clean up this PR and merge, and then tackle the conversion of transformed space in another PR.

bouthilx:
There is a great opportunity here to improve the transformation process and uniformize it with the conversion of spaces. This will take a bit more work, so I will merge this PR in it's current form to allow @donglinjy progress meanwhile on the integration of HPOBench. The interface of the conversion class will likely change but that will be simple to adjust to.

bouthilx (approved):

PR #833, opened 2022-03-16
Add Registry, rework how Space is passed to algorithms

### LINKED ISSUE #802: Rework how transformed space is passed to algorithms

lebrice:
# Description

Fixes #802 

# Changes

Algorithms now receive the space they will actual work with in their constructor.

# Checklist

## Tests

## Documentation

## Quality


src/orion/algo/asha.py
```diff
@@ -214,27 +237,37 @@ class ASHABracket(HyperbandBracket):
...
+        budgets: list[BudgetTuple],
+        repetition_id: int,
+    ):
+        super().__init__(hyperband=asha, budgets=budgets, repetition_id=repetition_id)
+        self.asha: ASHA = asha
+
+    def sample(self, num: int) -> list[Trial]:
         """Sample a new trial with lowest fidelity"""
         should_have_n_trials = self.rungs[0]["n_trials"]
+        # BUG: Shouldn't this be `sample_from_bracket`?
```
bouthilx, line 252: Wait, how could this not fail? Does it mean this is never called?

src/orion/algo/base.py
```diff
@@ -252,6 +223,7 @@ def observe(self, trials):
     def register(self, trial):
         """Save the trial as one suggested or observed by the algorithm.
 
+        TODO: Update this docstring:
```
bouthilx, line 226: TODO :)

src/orion/algo/base.py
```diff
@@ -263,20 +235,19 @@ def register(self, trial):
...
-        return len(self._trials_info)
+        return len(self.registry)
 
     @property
     def n_observed(self):
-        """Number of completed trials observed by the algorithm"""
-        return sum(bool(point[1] is not None) for point in self._trials_info.values())
+        """Number of completed trials observed by the algorithm."""
+        return sum(
+            trial.status in ("completed", "broken") for trial in self.registry.values()
```
bouthilx, line 249: Could use self.registry.has_observed(trial) to avoid duplicating the logic.

src/orion/algo/evolution_es.py
```diff
@@ -201,43 +211,47 @@ class BracketEVES(HyperbandBracket):
...
+    ):
+        super().__init__(evolution_es, budgets, repetition_id)
+        self.eves = evolution_es
         self.search_space_without_fidelity = []
-        self._candidates = {}
+        self._candidates: dict[int, list[Trial]] = {}
 
+        self.mutate_attr: dict = {}
         if evolution_es.mutate:
+            # TODO: This is inconsistent with the docstring of the constructor, which says it can be
```
bouthilx, line 224: TODO :)

src/orion/algo/hyperband.py
```diff
@@ -145,18 +170,33 @@ class Hyperband(BaseAlgorithm):
...
 
-        fidelity_dim = space[fidelity_index]
+        fidelity_dim: Fidelity = space[fidelity_index]
+
+        # NOTE: This isn't a Fidelity, it's a TransformedDimension<Fidelity>
+        from orion.core.worker.transformer import TransformedDimension
+
+        # TODO: Double-check with @bouthilx: Currently bypassing (possibly more than one)
+        # `TransformedDimension` wrappers to get the 'low', 'high' and 'base' attributes.
+        while isinstance(fidelity_dim, TransformedDimension):
```
bouthilx, line 197: Perhaps we should try to avoid wrapping the Fidelity dimension in the transformed space. This fix is hacky but necessary until this issue is resolved. In any case, I think it should be done in another PR when we have time. I'd be in favor of leaving this here as-is for this PR.

src/orion/algo/asha.py
lebrice: Never called, indeed.

src/orion/algo/evolution_es.py
lebrice: Should it be a string? or do we want a dict with {"function": "<function_path>"} ?

src/orion/algo/asha.py
bouthilx: Oh wow, thanks for noticing this dead limb! We should section it. :)

src/orion/algo/evolution_es.py
bouthilx: As you point out in the comment, it could be a string or a dict. We should first test whether it is a string and if so set `mutate_attr = {'function': evolution_es.mutate}`.

src/orion/algo/hyperband.py
```diff
@@ -145,18 +170,33 @@ class Hyperband(BaseAlgorithm):
...
+    def __init__(
+        self,
+        space: Space,
+        seed: int | Sequence[int] | None = None,
+        repetitions: int | float = numpy.inf,
+    ):
+        # NOTE: Need to set the attribute before calling super().__init__ because it calls seed_rng.
+        self.brackets: Optional[list[HyperbandBracket]] = []
+        super().__init__(space, repetitions=repetitions, seed=seed)
+        self.seed = seed
```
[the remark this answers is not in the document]
bouthilx: You mean it can't tell the type because it is set to self in `super`? Can't it tell the type from super() execution?

src/orion/algo/base.py
bouthilx: Fixed in baf15cbeaf8ecaafa304090046f47b0b3b150550. Thanks!

src/orion/algo/evolution_es.py
```diff
@@ -310,7 +331,7 @@ def _mutate_population(self, red_team, blue_team, rung, population_range, fideli
                     point[j] = self.eves.population[j][i]
 
                 trial = format_trials.tuple_to_trial(point, self.space)
-                trial = self.eves.format_trial(trial)
+                # trial = self.eves.format_trial(trial)
```
bouthilx, line 334:
```suggestion
```

src/orion/algo/hyperband.py
```diff
@@ -279,12 +330,14 @@ def register_samples(self, bracket, samples):
                 self.get_id(sample, ignore_fidelity=True, ignore_parent=True)
                 not in self.trial_to_brackets
             ):
+                assert self.brackets is not None
```
bouthilx, line 333:
```suggestion
```
Do we need this? If so it raise a RuntimeError with an informative error message.

src/orion/algo/hyperband.py
```diff
@@ -279,12 +330,14 @@ def register_samples(self, bracket, samples):
...
                 self.trial_to_brackets[
                     self.get_id(sample, ignore_fidelity=True, ignore_parent=True)
                 ] = self.brackets.index(bracket)
 
-    def promote(self, num):
-        samples = []
+    def promote(self, num: int) -> list[Trial]:
+        samples: list[Trial] = []
+        assert self.brackets is not None
```
bouthilx, line 340: Same here

src/orion/algo/hyperband.py
```diff
@@ -229,42 +276,46 @@ def seed_rng(self, seed):
...
 
         :param state_dict: Dictionary representing state of an algorithm
         """
+        super().set_state(state_dict)
         self.seed_rng(state_dict["seed"])
         self.rng.set_state(state_dict["rng_state"])
-        self._trials_info = state_dict["_trials_info"]
         self.trial_to_brackets = state_dict["trial_to_brackets"]
-
+        assert self.brackets is not None
```
bouthilx, line 309: Oh, same here

src/orion/algo/hyperband.py
```diff
@@ -293,19 +346,20 @@ def promote(self, num):
 
         return samples
 
-    def sample(self, num):
-        samples = []
+    def sample(self, num: int) -> list[Trial]:
+        samples: list[Trial] = []
+        assert self.brackets is not None
```
bouthilx, line 351: Another

src/orion/algo/hyperband.py
```diff
@@ -319,18 +373,18 @@ def suggest(self, num):
...
+            which case it will return None.
 
         """
         self._refresh_brackets()
 
         samples = self.promote(num)
 
         samples.extend(self.sample(max(num - len(samples), 0)))
-
+        assert self.brackets is not None
```
bouthilx, line 387: Another

src/orion/algo/hyperband.py
```diff
@@ -349,22 +403,24 @@ def suggest(self, num):
                 f"{self.__class__.__name__} cannot suggest new samples and must wait "
                 "for trials to complete."
             )
-
+        # TODO: Shouldn't this return None in this case?
         return []
```
bouthilx, line 407: Either is fine. I think we should maybe eventually only return lists, which will be empty when the algo is unable to sample new trials. This would simplify a bit the handling of algo suggestions.

src/orion/algo/hyperband.py
```diff
@@ -349,22 +403,24 @@ def suggest(self, num):
...
+        assert self.budgets is not None
         all_brackets_done = all(
             bracket.is_done for bracket in self.brackets[-len(self.budgets) :]
         )
         return executed_times - int(not all_brackets_done)
 
-    def _refresh_brackets(self):
+    def _refresh_brackets(self) -> None:
         """Refresh bracket if one hyperband execution is done"""
+        assert self.brackets is not None
```
bouthilx, line 423: Another

src/orion/algo/hyperband.py
```diff
@@ -376,23 +432,26 @@ def _refresh_brackets(self):
             if self.executed_times < self.repetitions:
                 self.append_brackets()
 
-    def append_brackets(self):
+    def append_brackets(self) -> None:
+        assert self.brackets is not None
```
bouthilx, line 436: Another

src/orion/algo/hyperband.py
```diff
@@ -376,23 +432,26 @@ def _refresh_brackets(self):
...
         self.brackets = self.brackets + self.create_brackets()
         # Reset brackets seeds
         self.seed_brackets(self.seed)
 
-    def create_brackets(self):
+    def create_brackets(self) -> list[HyperbandBracket]:
+        assert self.budgets is not None
```
bouthilx, line 442: Oh a slightly different one. :)

src/orion/algo/hyperband.py
```diff
@@ -376,23 +432,26 @@ def _refresh_brackets(self):
...
         return [
             self.create_bracket(i, bracket_budgets, self.executed_times + 1)
             for i, bracket_budgets in enumerate(self.budgets)
         ]
 
-    def _get_bracket(self, trial):
+    def _get_bracket(self, trial: Trial) -> HyperbandBracket:
         """Get the bracket of a trial"""
         _id_wo_fidelity = self.get_id(trial, ignore_fidelity=True, ignore_parent=True)
+        assert self.brackets is not None
```
bouthilx, line 451: Another

src/orion/algo/hyperband.py
```diff
@@ -426,10 +485,22 @@ def observe(self, trials):
...
 
     @property
-    def is_done(self):
-        """Return True, if all required execution been done."""
+    def is_done(self) -> bool:
+        """Return True, if all required execution has been done."""
         if self.executed_times >= self.repetitions:
             return True
+        fidelity_index = self.fidelity_index
+        assert fidelity_index is not None
```
bouthilx, line 493: Another

src/orion/algo/hyperband.py
```diff
@@ -426,10 +485,22 @@ def observe(self, trials):
                 continue
 
     @property
-    def is_done(self):
-        """Return True, if all required execution been done."""
+    def is_done(self) -> bool:
+        """Return True, if all required execution has been done."""
         if self.executed_times >= self.repetitions:
             return True
+        fidelity_index = self.fidelity_index
```
bouthilx, line 492: Perhaps this test could be in BaseAlgorithm. If there is a fidelity dimension, we take it into consideration, otherwise only count the trials. This should apply to any fidelity-based algorithm.

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
```diff
@@ -216,11 +262,12 @@ def sample_from_bracket(self, bracket, num):
 
                 trials.append(trial)
                 self.register(trial)
+                assert self.brackets is not None
```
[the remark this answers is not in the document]
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
```diff
@@ -349,22 +403,24 @@ def suggest(self, num):
...
         return []
 
     @property
-    def executed_times(self):
+    def executed_times(self) -> int:
         """Counter for how many times Hyperband been executed"""
         if not self.brackets:
             return 0
         executed_times = self.brackets[-1].repetition_id
+        assert self.budgets is not None
```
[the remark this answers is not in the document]
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
lebrice: Fixed in 35ce6867769d77028c87f7bb015adfd349315f85

src/orion/algo/hyperband.py
```diff
@@ -497,56 +570,65 @@ def seed(self, seed):
...
         )
-
-    def _get_results(self, trial):
+        # Note: replaces this:
+        # self._get_results(trial)[
+        #     self.hyperband.get_id(trial, ignore_fidelity=True, ignore_parent=True)
+        # ] = (
+        #     trial.objective.value if trial.objective else None,
+        #     copy.deepcopy(trial),
+        # )
```
bouthilx, line 593:
```suggestion
```

src/orion/algo/hyperband.py
lebrice: Disagree on this one.
The exception would already have been raised here when instantiating the object:
https://github.com/lebrice/orion/blob/35ce6867769d77028c87f7bb015adfd349315f85/src/orion/algo/hyperband.py#L188-190
Hence using an assert is perfectly adequate here.

src/orion/algo/hyperband.py
lebrice: Does every fidelity-based algorithm want this logic? Could a new fidelity-based algorithm want something different?
(Related question: Do we have any fidelity-based algorithms that aren't based on Hyperband? If so, do they need this?)

src/orion/algo/hyperband.py
```diff
@@ -497,56 +570,65 @@ def seed(self, seed):
...
-    def sample(self, num):
+    def sample(self, num: int) -> list[Trial]:
         """Sample a new trial with lowest fidelity"""
         should_have_n_trials = self.rungs[0]["n_trials"]
         n_trials = len(self.rungs[0]["results"])
         request = max(min(should_have_n_trials - n_trials, num), 0)
         if request == 0:
             return []
-
+        # BUG: Shouldn't this be `sample_from_bracket`?
```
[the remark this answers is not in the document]
lebrice: It might be that `request` is always equal to 0, since we never call `suggest` with a big enough number. 
In any case, I'd argue that it is out of the scope of this PR to do this kind of debugging + refactoring.

src/orion/algo/hyperband.py
```diff
@@ -557,16 +639,23 @@ def get_candidates(self, rung_id):
         if self.has_rung_filled(rung_id + 1):
             return []
 
-        rung = self.rungs[rung_id]["results"]
+        rung_results = self.rungs[rung_id]["results"]
         next_rung = self.rungs[rung_id + 1]["results"]
+        # BUG: What if some of the objectives are None? Then comparison between None and floats here
```
bouthilx, line 629: I think we should filter them to only select objectives that are not None in `rung_results`. We can then sort safely.

src/orion/algo/pbt/pbt.py
```diff
@@ -384,14 +402,15 @@ def _fork_lineages(self, num):
 
             logger.debug("Forking trial %s to %s", trial_to_branch, new_trial)
 
+            assert new_trial is not None
```
bouthilx, line 405: Another assert

src/orion/algo/pbt/pbt.py
```diff
@@ -436,6 +456,7 @@ def _generate_offspring(self, trial):
             ]
 
             new_trial = trial_to_branch.branch(params=new_params)
+            # TODO: Keep this? or not?
             new_trial = self.space.transform(self.space.reverse(new_trial))
```
bouthilx, line 460: I don't understand why I did not use self.format_trial. 

I think there is a problem here with the separate registries and the fact that the algorithm only see the trials in the transformed space. PBT copies over the working_dir of the trials based on their IDs, but the ID that the user will get is based on the params in the original space. So the user and PBT will work with different working_dirs.

I don't see how we could easily solve this if we get rit of transform(reverse()) here. :thinking: Maybe fixing this should be part of another PR.

src/orion/algo/pbt/pbt.py
```diff
@@ -802,6 +823,7 @@ def fork(self, new_trial):
             identical working_dir would imply that different trials have identical parameters.
 
         """
+        # BUG: Here getting the working_dir property is causing an issue.
```
bouthilx, line 826: Is it related to my comment here? https://github.com/Epistimio/orion/pull/833/files#r834557496

src/orion/algo/base.py
lebrice: Fixed in 83c4670183e6ca925b2ec00f5024f1a19f601938
(might not be pushed already)

src/orion/algo/evolution_es.py
lebrice: Fixed in e15258e96c880198314ad8b3399080efb2993122

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __iter__(self) -> Iterator[Trial]:
+        return iter(self._trials.values())
+
+    def __len__(self) -> int:
+        return len(self._trials)
+
+    @property
+    def state_dict(self) -> dict:
+        return {"_trials": self._trials}
```
bouthilx, line 47: Maybe we should deepcopy here and in set_state to be safe.

src/orion/algo/hyperband.py
lebrice: Can I mark this as resolved?

lebrice: Oh nevermind I'll remove the TODO first.

lebrice: Fixed in 1b74ab37bfcb95f3741d52a875b91d495c0db89c
(might not be pushed yet)

src/orion/algo/hyperband.py
lebrice: The alternative would be to overwrite the `fidelity_index` property in this class, and have its annotation be a `str` instead of `str | None`, and raise an error inside the property if there is not fidelity property. Would you like that better?

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+            "_mapping": self._mapping,
+        }
+
+    def set_state(self, statedict: dict):
+        self.original_registry.set_state(statedict["original_registry"])
+        self.transformed_registry.set_state(statedict["transformed_registry"])
+        self._mapping = statedict["_mapping"]
+
+    def __iter__(self) -> Iterator[Trial]:
+        return iter(self.original_registry)
```
bouthilx, line 106: Not sure why we would iterate on the mapping, but if we would do I would assume we want the mapped (original_trial, transformed_trials).

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __len__(self) -> int:
+        return len(self.original_registry)
+
+    def __contains__(self, trial: Trial):
+        return trial in self.original_registry
```
bouthilx, line 112: Same for contains.

src/orion/algo/pbt/pbt.py
```diff
@@ -375,7 +392,8 @@ def _fork_lineages(self, num):
                 logger.debug("Skipping trial %s", trial)
                 skipped_trials.append(trial)
                 continue
-
+            # NOTE: This assumes that new_trial is not None.
+            assert new_trial is not None
```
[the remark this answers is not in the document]
lebrice: I'm adding an assert here because that is exactly what the code was already doing, but written more explicitly.
In this particular case, the next few lines would have crashed at runtime if `self._generate_offspring` had returned `None` for `new_trial`, and if `None` was passed as an argument to `fork(new_trial)`.
 
See what I mean? This PR isn't exactly responsible for fixing existing bugs in PBT or other algorithms. This assert here just makes the existing assumptions more explicit, and easier to identify / debug / solve later.

src/orion/algo/pbt/pbt.py
lebrice: See other reply, but indeed this one is not needed (since the one above is sufficient)

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+    def get_trials(self, original_trial: Trial) -> list[Trial]:
+        """Return the registered transformed trials that map to the given trial in the original
+        space.
+        """
+        return self.get(original_trial, [])
+
+    def register(self, original_trial: Trial, transformed_trial: Trial) -> str:
+        """Register an equivalence between the given original trial and the transformed trial."""
+        # NOTE: Choosing not to register the trials here, and instead do it more manually.
+        # original_id = self.original_registry.register(original_trial)
```
bouthilx, line 133: On one hand it may be confusing to have the side effect of populating the registries when calling mapping.register. On the other hand we risk having incoherent original and transformed registries with the mapping if we rely on updating them externally. Handling them here would not block external updates on the registries anyhow, so maybe we are at risk for incoherence anyway and we better make it explicit like you do here. :thinking:

src/orion/algo/registry.py
lebrice: Good catch!

src/orion/algo/registry.py
lebrice: Sorry to ask, but do I need to change something here?

src/orion/algo/tpe.py
```diff
@@ -238,19 +256,30 @@ def __init__(
...
 
-    @space.setter
-    def space(self, space):
-        """Set the space of TPE and initialize it"""
-        self._space = space
-        self._initialize()
+    # @space.setter
+    # def space(self, space: Space | TransformedSpace) -> None:
+    #     """Set the space of TPE and initialize it"""
+    #     self._space = space
```
bouthilx, line 280:
```suggestion
```

src/orion/algo/tpe.py
```diff
@@ -238,19 +256,30 @@ def __init__(
...
 
-    def _initialize(self):
+    def _initialize(self) -> None:
         """Initialize TPE once the space is transformed"""
```
bouthilx, line 283:
```suggestion
    def _verify_space(self) -> None:
        """Verify space is supported"""
```

src/orion/algo/tpe.py
```diff
@@ -271,35 +300,35 @@ def _initialize(self):
             if shape and len(shape) != 1:
                 raise ValueError("TPE now only supports 1D shape.")
 
-    def seed_rng(self, seed):
+    def seed_rng(self, seed: int | Sequence[int] | None) -> None:
         """Seed the state of the random number generator.
 
         :param seed: Integer seed for the random number generator.
         """
+        super().seed_rng(seed)
```
bouthilx, line 308: Why add super call?

src/orion/algo/tpe.py
```diff
@@ -308,47 +337,51 @@ def suggest(self, num=None):
         num: int, optional
             Number of trials to sample. If None, TPE will sample all random trials at once, or a
             single trial if it is at the Bayesian Optimization stage.
-        :param num: how many sets to be suggested.
 
         .. note:: New parameters must be compliant with the problem's domain
            `orion.algo.space.Space`.
         """
+        # TODO: Code below assumes that `num` is an int, but docstring says it can be None.
```
bouthilx, line 344: Doc is out-of-date. You can correct docstring and remove the assert.

src/orion/algo/tpe.py
```diff
@@ -308,47 +337,51 @@ def suggest(self, num=None):
...
+                v = max(num_samples - len(samples), 0)
+                candidates = self._suggest_bo(v)
+                logger.debug("BO candidates: %s", candidates)
 
             if not candidates:
+                # Perhaps the BO algo wasn't able to suggest any points? Break in that case.
                 break
-
+            for candidate in candidates:
+                if candidate:
```
bouthilx, line 360: This can be None?

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+            trial_id = _get_id(trial_or_id._cm_trial)
+        elif isinstance(trial_or_id, Trial):
+            trial_id = _get_id(trial_or_id)
+        elif isinstance(trial_or_id, str):
+            trial_id = trial_or_id
+        else:
+            raise NotImplementedError(trial_or_id)
+        return trial_id in self._trials
+
+    def __getitem__(self, item: int | str) -> Trial:
```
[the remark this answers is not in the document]
lebrice: Hmm interesting, good point. The only operations we perform over the registry:
- append (register)
- check for membership (`__contains__`)
- iterate over the registry.

This also fits with the [`Collection` base class](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). I'll remove the int support of `__getitem__` and see if the tests still pass.

src/orion/algo/tpe.py
```diff
@@ -308,47 +337,51 @@ def suggest(self, num=None):
...
-        ids = set(self._trials_info.keys())
         retries = 0
         while len(trials) < num and retries < self.max_retry:
             for candidate in function(num - len(trials)):
-                candidate_id = self.get_id(candidate)
-                if candidate_id not in ids:
-                    ids.add(candidate_id)
+                if candidate not in self.registry:
+                    self.register(candidate)
+                    # self.registry.register(candidate)
```
bouthilx, line 375:
```suggestion
```

src/orion/algo/tpe.py
```diff
@@ -414,9 +447,15 @@ def _suggest_one_bo(self):
             params[dimension.name] = dim_samples
 
         trial = format_trials.dict_to_trial(params, self.space)
-        return self.format_trial(trial)
+        # return self.format_trial(trial)
```
bouthilx, line 446:
```suggestion
```

src/orion/algo/tpe.py
```diff
@@ -478,9 +537,16 @@ def _sample_real_point(self, dimension, below_points, above_points, is_log=False
...
-    def _sample_int_point(self, dimension, below_points, above_points):
+    def _sample_int_point(
+        self,
+        dimension: Integer,
+        below_points: numpy.ndarray | Sequence[numpy.ndarray],
+        above_points: numpy.ndarray | Sequence[numpy.ndarray],
+    ):
         """Sample one value for integer dimension based on the observed good and bad points"""
         low, high = dimension.interval()
+        assert isinstance(low, int)
```
bouthilx, line 544: Another assert

src/orion/algo/registry.py
lebrice: Fixed in eaa2254453102689c230a4b5fb12952b9be77b20 (not pushed yet)

src/orion/algo/registry.py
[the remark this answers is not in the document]
lebrice: Fixed in eaa2254453102689c230a4b5fb12952b9be77b20 (not pushed yet)

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+        self.original_registry = original_registry
+        self.transformed_registry = transformed_registry
+        self._mapping: dict[str, set[str]] = defaultdict(set)
+
+    @property
+    def state_dict(self) -> dict:
+        return {
+            "original_registry": self.original_registry.state_dict,
+            "transformed_registry": self.transformed_registry.state_dict,
+            "_mapping": self._mapping,
```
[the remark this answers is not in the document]
lebrice: Fixed in eaa2254453102689c230a4b5fb12952b9be77b20 (not pushed yet)

src/orion/algo/registry.py
lebrice: Fixed in aa72f1f9b7c21c97ad466444fa987bea1ff060d2

src/orion/algo/tpe.py
```diff
@@ -518,18 +586,19 @@ def _sample_categorical_point(self, dimension, below_points, above_points):
...
+        for trial in self.registry:
             if trial.status != "completed":
                 trial = self.strategy.infer(trial)
 
             if trial is not None:
                 trials.append(trial)
-
-        sorted_trials = sorted(trials, key=lambda trial: trial.objective.value)
+        # NOTE: This assumes that all trials have an objective. Making assumption explicit.
+        assert all(trial.objective is not None for trial in trials)
```
bouthilx, line 596: Like for Hyperband, I would filter in the loop above instead and log a warning if that happens.

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+    def set_state(self, statedict: dict):
+        self.original_registry.set_state(statedict["original_registry"])
+        self.transformed_registry.set_state(statedict["transformed_registry"])
+        self._mapping = statedict["_mapping"]
+
+    def __iter__(self) -> Iterator[Trial]:
+        return iter(self.original_registry)
+
+    def __len__(self) -> int:
+        return len(self.original_registry)
```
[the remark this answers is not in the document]
lebrice: How could a trial be in the original registry, but not in the mapping?
This shouldn't ever happen, right?

src/orion/algo/registry.py
lebrice: Same, as-in, we should only check if it is in the mapping, rather than checking if it is in the original registry?

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __contains__(self, trial: Trial):
+        return trial in self.original_registry
+
+    def __getitem__(self, item: Trial) -> list[Trial]:
+        trial_id = _get_id(item)
+        if trial_id not in self._mapping:
```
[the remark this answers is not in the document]
lebrice: Fixed in aa72f1f9b7c21c97ad466444fa987bea1ff060d2

src/orion/algo/tpe.py
```diff
@@ -630,20 +700,23 @@ def sample(self, num=1, attempts=None):
                     f"Failed to sample in interval ({self.low}, {self.high})"
                 )
             elif index is None:
+                # TODO: Why is this recursive? What is the proper return type of this function?
```
bouthilx, line 699: That's to handle situations where we sample outside of the bounds. When TPE is converging to a very narrow space it becomes difficult to sample within that space.

src/orion/algo/tpe.py
```diff
@@ -594,25 +663,26 @@ def __init__(
...
             self.pdfs.append(norm(mu, sigma))
 
-    def sample(self, num=1, attempts=None):
+    def sample(self, num: int = 1, attempts: int | None = None) -> list[numpy.ndarray]:
         """Sample required number of points"""
         if attempts is None:
             attempts = self.base_attempts
-
-        point = []
+        # TODO: See below, unsure about the return type for this function.
```
bouthilx, line 674: Seems like numpy.ndarray or scalar? https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.rvs.html

src/orion/algo/tpe.py
lebrice: If the BaseAlgorithm ever adds a default implementation for seed_rng, then this will save us from having to debug flaky tests etc.
Until then, it's a no-op. Doesn't hurt to have it IMO.

src/orion/core/io/experiment_builder.py
```diff
@@ -505,10 +505,16 @@ def _instantiate_algo(space, max_trials, config=None, ignore_unavailable=False):
...
-        algo = SpaceTransformAlgoWrapper(
-            algo_constructor, space=space, **backported_config
-        )
-        algo.algorithm.max_trials = max_trials
+
+        requirements = backward.get_algo_requirements(algo_constructor)
+        from orion.core.worker.transformer import build_required_space
+
+        original_space = space
+        transformed_space = build_required_space(space, **requirements)
```
bouthilx, line 513: There should be a helper function to this. Building algos outside experiment creation will be cumbersome otherwise since algos can hardly be used on non transformed spaces based on their requirements.

src/orion/core/utils/tree.py
```diff
@@ -327,7 +342,13 @@ def has_depth(node, children):
 
         return sum([node.item for node in nodes], [])
 
-    def map(self, function, node):
+    # TODO: Trying to type-annotate this function with overloads is a complete nightmare.
```
bouthilx, line 345: Abort this file for this PR maybe?

src/orion/core/worker/primary_algo.py
```diff
@@ -21,39 +33,67 @@ class SpaceTransformAlgoWrapper:
...
         self._space = space
-        requirements = backward.get_algo_requirements(algo_constructor)
-        self.transformed_space = build_required_space(space, **requirements)
-        self.algorithm = algo_constructor(space, **algorithm_config)
-        self.algorithm.space = self.transformed_space
+        self.algorithm: AlgoType = algorithm
+        self.registry = Registry()
+        self.registry_mapping = RegistryMapping(
+            original_registry=self.registry,
+            transformed_registry=self.algorithm.registry,
```
bouthilx, line 51: Right, the transformed_registry is the one from the algorithm. Because of this it is better indeed that mapping does not register for the registers otherwise the algorithm would need to account for the fact that someone externally may mess with its registry...

src/orion/core/worker/primary_algo.py
```diff
@@ -21,39 +33,67 @@ class SpaceTransformAlgoWrapper:
...
         """Return a state dict that can be used to reset the state of the algorithm."""
-        return self.algorithm.state_dict
+        # TODO: There's currently some duplicates between:
+        # - self.registry_mapping.original_registry and self.registry
+        # - self.registry_mapping.transformed_registry and self.algorithm.registry
+        return copy.deepcopy(
+            {
+                "algorithm": self.algorithm.state_dict,
+                "registry": self.registry.state_dict,
+                "registry_mapping": self.registry_mapping.state_dict,
```
bouthilx, line 82: I wounder if we should not recreate the registry_mapping based on the `_mapping`. The original registry and the transformed registry are handling their state_dict separately, so this should not be part of the state_dict of the registry_mapping. So set_state here could be simply `self.registry_mapping = RegistryMapping(self.registry, self.algorithm.registry, mapping=state_dict['registry_mapping'])` or something like that. What do you think?

src/orion/algo/tpe.py
```diff
@@ -238,19 +256,30 @@ def __init__(
...
+        self.equal_weight = equal_weight
+        self.prior_weight = prior_weight
+        self.full_weight_num = full_weight_num
+        self.max_retry = max_retry
+        self.parallel_strategy = parallel_strategy
+        self._initialize()
 
     @property
-    def space(self):
+    def space(self) -> Space | TransformedSpace:
```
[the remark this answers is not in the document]
lebrice: Fixed in a6b4f190d65f8cf9f0f01d3aed49ec780a54551f (not pushed yet)

src/orion/algo/tpe.py
lebrice: Fixed in 5bce0b8a430543624c811628b28b9f096cec9c64

src/orion/algo/tpe.py
lebrice: Fixed in ac71d33c27db4e7cf7d620c392e41d0762fda69e (not pushed yet)

src/orion/algo/tpe.py
lebrice: This assert is fine. Again, this is making the `TypeError` that you'd get anyway with the `range` below more explicit.

src/orion/algo/tpe.py
lebrice: Like for HyperBand, here I'm only pointing out an implicit assumption of the existing code (sorting would raise an exception if any objective is `None`)
Improvements to the algorithms belong in another PR IMO.

src/orion/core/utils/tree.py
lebrice: No! :stuck_out_tongue: 
Fixed in a5dac7573a3b7ff2bc879502cef1d28d849d2b66

src/orion/core/worker/primary_algo.py
lebrice: Hmm, interesting. I think it doesn't really hurt to call `set_state` a few extra times than necessary.
The alternative would be to make it so `RegistryMapping.set_state` only saves/restores `self._mapping`, and just trust that the registries themselves are always saved/restored at the same time.

src/orion/core/io/experiment_builder.py
lebrice: Where should this helper function be placed?

src/orion/algo/asha.py
lebrice: Fixed in eaa2254453102689c230a4b5fb12952b9be77b20

src/orion/algo/hyperband.py
lebrice: Fixed in 3193cad11790c20076ca4de4b048aa00f6d76168 (not pushed yet)

src/orion/algo/hyperband.py
lebrice: Again, this is out-of-scope IMO. 
This assertion here is making the implicit error more explicit. It'll be the job of another PR to debug/improve these algos as needed.

lebrice: Also, more generally, in this kind of case, I think leaving a "BUG" comment is enough, but if you want me to make an issue to keep track of this, I can also make one.

src/orion/algo/pbt/pbt.py
lebrice: Fixed in 168d99384cc8b8d2cb8c35c7097599d8b03c0783

src/orion/algo/pbt/pbt.py
lebrice: Resolved?

src/orion/algo/pbt/pbt.py
lebrice: Not quite. This was before I copied over the experiment in `_copy_status_and_results` of `primary_algo.py`.

Fixed in 0a2a2494da470b591a6bf06584f1fe9f750c881a

src/orion/algo/tpe.py
lebrice: Fixed in 9711d1737afae8adba70986762e8219f4a268a66 (not pushed at the time of writing this)

src/orion/algo/tpe.py
lebrice: Fixed in 040db502d7c79823ef8130affb10a76f911198a1

src/orion/core/utils/tree.py
lebrice: More seriously, this is used in PBT. I added the missing type annotations to PBT in 91a7c2221fbadde3397aa843e7901a08358c1dea, and pbt.py benefits from having fully-typed `TreeNode` (and `LineageNode` by extension)

src/orion/core/io/experiment_builder.py
lebrice: Added a helper function in 0ae1340407e9e37909ecb66d2516dc325ec9c6f7, let me know where you'd like it to be moved.

src/orion/algo/registry.py
```diff
@@ -0,0 +1,147 @@
+""" Classes that serve as an in-memory storage of trials for the algorithms. """
```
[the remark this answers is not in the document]
lebrice: Fixed in fbfc9f2e6910c012b0425709e07937a445ac9683

src/orion/core/worker/primary_algo.py
```diff
@@ -79,64 +119,133 @@ def suggest(self, num):
...
-            trial = self.transformed_space.reverse(transformed_trial)
-            self._verify_trial(trial, space=self.space)
-            trials.append(trial)
-
+            original = self.transformed_space.reverse(transformed_trial)
+            if original in self.registry:
+                # We already have a trial that is equivalent to this one.
+                # Copy over the status and results and observe it.
+                transformed_trial = _copy_status_and_results(
+                    original_trial=original, transformed_trial=transformed_trial
```
bouthilx, line 135: Shouldn't we fetch the original from registry to do this? `original` is a transformed trial that was suggested from the algo, so it should not have results. `self.register[_get_id(original.id)]` could have results and a different status however.

src/orion/core/worker/primary_algo.py
```diff
@@ -79,64 +119,133 @@ def suggest(self, num):
...
+                )
+                self.algorithm.observe([transformed_trial])
+            else:
+                # We haven't seen this trial before. Register it.
+                trials.append(original)
+
+            # Register the trial in our registry.
+            self.registry.register(original)
```
bouthilx, line 143: If the original was already in register, we may overwrite the status of the previous one, losing results and completed status for instance. Unless I am not reading the code correctly and there is no issue, these cases should be tested explicitly to avoing reintroducing these issues.

src/orion/core/worker/primary_algo.py
```diff
@@ -79,64 +119,133 @@ def suggest(self, num):
...
+            # Here we register the transformed trial just in case the algorithm hasn't already
+            # registered the trial in its `suggest`.
+            self.algorithm.register(transformed_trial)
```
bouthilx, line 146: This may result in weird behaviors if the algorithm is not expecting to have trials registered externally. :thinking: I don't have a strong opinion.

src/orion/core/worker/primary_algo.py
```diff
@@ -79,64 +119,133 @@ def suggest(self, num):
...
+                    f"Observing trial {trial} (transformed as {transformed_trial}), even "
+                    f"though it wasn't suggested by the algorithm."
+                )
+                # NOTE: @lebrice Here we don't want to store the transformed trial in the
+                # algo's registry (by either calling `self.algorithm.register(transformed_trial)` or
+                # `self.algorithm.registry.register(transformed_trial))`, because some algos can't
+                # observe trials that they haven't suggested. We'd also need to perform all the
+                # logic that the algo did in `suggest` (e.g. store it in a bracket for HyperBand).
+                # Therefore we only register it in the wrapper, and store the equivalence between
+                # these two trials in the registry mapping.
```
bouthilx, line 192: I think this is the correct way to handle this situation. And to be coherent, this suggests that we should not register the trials during `suggest` in the algorithm registry. It should be the algorithm's responsibility.

src/orion/core/worker/primary_algo.py
```diff
@@ -205,3 +316,15 @@ def _verify_trial(self, trial, space=None):
                 f"Trial {trial.id} not contained in space:"
                 f"\nParams: {trial.params}\nSpace: {space}"
             )
+
+
+def _copy_status_and_results(original_trial: Trial, transformed_trial: Trial) -> Trial:
+    """Copies the results, status, and other data from `transformed_trial` to `original_trial`."""
+    transformed_trial = copy.deepcopy(transformed_trial)
```
bouthilx, line 323: What would you think of doing the opposite?
```
copied_transformed_trial = copy.deepcopy(original_trial)
copied_transformed_trial._params = transformed_trials._params
return copied_transformed_trials
```

src/orion/testing/algo.py
```diff
@@ -106,6 +112,23 @@ def customized_mutate_example(search_space, rng, old_value, **kwargs):
     return new_value
 
 
+AlgoType = TypeVar("AlgoType", bound=BaseAlgorithm)
+
+
+def create_algo(
```
bouthilx, line 118: I think this wrapper could be moved to primary_algo.py.

tests/conftest.py
```diff
@@ -24,10 +24,10 @@
 from orion.core.worker.trial import Trial
 from orion.storage.base import get_storage, setup_storage, storage_factory
 from orion.storage.legacy import Legacy
-from orion.testing import OrionState, mocked_datetime
 
 # So that assert messages show up in tests defined outside testing suite.
 pytest.register_assert_rewrite("orion.testing")
+from orion.testing import OrionState, mocked_datetime
```
bouthilx, line 30: Does the pytest.register_assert_rewrite if we import the module afterwards?

tests/unittests/algo/test_tpe.py
```diff
@@ -620,90 +632,137 @@ def test_sample_real_dimension(self):
...
+                dist_requirement=TPE.requires_dist,
+            )
+            algo = TPE(transformed_space, **tpe_kwargs)
+            wrapper = SpaceTransformAlgoWrapper(algo, space=space)
+            return wrapper
+
+        tpe = _create_algo(space=space, seed=1, n_initial_points=10)
+        # tpe = TPE(space=space, seed=1)
+
+        # tpe.n_initial_points = 10
```
bouthilx, line 680:
```suggestion

```

bouthilx:
When PR is ready and approved, we need to look at TPE xfails.

tests/unittests/algo/test_tpe.py
```diff
@@ -821,45 +886,60 @@ def test_is_done_cardinality(self):
...
+            v = self._suggest(num, sample)
+            print(f"num={num}, Suggesting {v} ({len(values)} values left.)")
+            return v
 
         monkeypatch.setattr("orion.algo.tpe.TPE._suggest_random", _suggest_random)
         monkeypatch.setattr("orion.algo.tpe.TPE._suggest_bo", _suggest_random)
+
         self.force_observe(RANGE, algo)
-        assert algo.n_observed == RANGE
+        # # BUG: How can a trial be in the registry, if that wasn't observed?
```
bouthilx, line 920: You mean the test is failing because there are more points in algo.registry than in algo.algorithm.registry?

tests/unittests/client/test_experiment_client.py
```diff
@@ -637,7 +637,9 @@ def test_suggest_new(self):
 
             trial = client.suggest()
             assert trial.status == "reserved"
-            assert trial.params["x"] == 57.57
+            # assert trial.params["x"] == 57.57  # BUG: This check is failing atm.
+            # FIXME: Changing the hard-coded value here so the test now passes.
+            assert trial.params["x"] == 83.40
```
bouthilx, line 642: That's because of the change in algo/random.py right? If so that's fine, you can change the value and remove the comment.

tests/unittests/core/test_primary_algo.py
```diff
@@ -65,50 +79,67 @@ def test_init_and_configuration(self, dumbalgo, palgo, fixed_suggestion_value):
...
-        ] * 4
-        fixed_suggestion_value = fixed_suggestion
+        trials = palgo.suggest(1)
+        assert trials is not None
+        assert trials[0].params == fixed_suggestion.params
+        ptrials = palgo.suggest(4)
+        # NOTE: Now, if an algorithm has already suggested the same trial, we don't return a
+        # duplicate.
+        assert not ptrials
+        # assert ptrials is not None
```
bouthilx, line 102: Was this for debugging? Can it be remove or uncommented?

src/orion/algo/hyperband.py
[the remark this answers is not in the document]
bouthilx: Ah ok, thanks for the explanation!

src/orion/algo/hyperband.py
bouthilx: Create issue #856

src/orion/algo/hyperband.py
bouthilx: Do you mean doing this for Hyperband (and subclasses) specifically or BaseAlgorithm? That would be reasonable for Hyperband. For BaseAlgorithm we would need a different error message.

src/orion/algo/hyperband.py
bouthilx: Yes the logic would be valid for any fidelity-based algorithms. The idea is that as long as a trial is not a full fidelity than it is not considered completed and the algorithm has not exhausted the space.

PBT does not inherit from Hyperband. It would not apply for PBT but it overwrites BaseAlgorithm.is_done anyway.

src/orion/algo/hyperband.py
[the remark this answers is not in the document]
bouthilx: Agreed! Created #857

src/orion/core/worker/primary_algo.py
lebrice: Good point. I'll only register the trial in the case where it wasn't already in the registry, just to be safe.

src/orion/algo/hyperband.py
bouthilx: It's already difficult to keep track of all the issues. Leaving BUG comments is to likely to be missed for a long time in my opinion. I can create the issues this time as we discuss. Next time feel free to create the issues as you find bugs like these.

Regarding the assert, it's better to use explicit error messages. The users will be confused if they get asserts like this and won't understand what is happening. We can at least add a message to the assert, but the type of error message itself always convey information.

I created #858

src/orion/algo/pbt/pbt.py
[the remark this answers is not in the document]
bouthilx: `_generate_offspring` returns `(Trial, Trial)` or `(None, None)` so we should never get to this point with `new_trial = None`.  At least if we keep the assert there should be a message saying `new_trial should not be None if trial_to_branch is not, please report this issue at https://github.com/Epistimio/orion/issues`.

src/orion/core/worker/primary_algo.py
lebrice: Ouh, interesting idea! Will try that!

tests/conftest.py
lebrice: I was getting a warning because it can't register assert rewrites if the module has already been imported. I think this is the correct ordering, but I might be wrong.

lebrice:
> When PR is ready and approved, we need to look at TPE xfails.

In another PR though.

src/orion/algo/registry.py
bouthilx: Commit is for hyperband.

src/orion/algo/registry.py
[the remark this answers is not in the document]
bouthilx: Indeed, it should not. But since the trial is registered externally there can be a bug outside of the registry that cause this.  :thinking:

src/orion/algo/registry.py
bouthilx: No I'm sorry :smile: I was just sharing my thoughts in case we should discuss this, but I latter agreed with your solution: https://github.com/Epistimio/orion/pull/833#discussion_r834688021

src/orion/algo/tpe.py
bouthilx: Alright, sounds good.

src/orion/algo/tpe.py
bouthilx: I agree it's outside the scope of this PR, but I have the same opinion as in the other discussion regarding the error message.

src/orion/core/io/experiment_builder.py
bouthilx: I think inside the wrapper module would make sense.

src/orion/core/worker/primary_algo.py
bouthilx: Yep :slightly_smiling_face:

tests/unittests/algo/test_tpe.py
lebrice: No just an old comment while I was debugging this.

src/orion/core/worker/primary_algo.py
lebrice: Fixed in f728f658f31e39ed7c4ae66f90b83ae4b8b77018

src/orion/core/worker/primary_algo.py
```diff
@@ -172,31 +281,33 @@ def configuration(self):
...
         return self._space
 
-    def get_id(self, point, ignore_fidelity=False):
-        """Compute a unique hash for a point based on params"""
-        return self.algorithm.get_id(
-            self.transformed_space.transform(point), ignore_fidelity=ignore_fidelity
-        )
+    def get_id(self, trial: Trial, ignore_fidelity: bool = False) -> str:
+        """Compute a unique hash for a trial based on params"""
+        # TODO: Double-check this with @bouthilx: Should we check the registry mapping? Or create a
```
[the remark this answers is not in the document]
lebrice: Fixed in ff00c6595a7a0f2a8f1751373c7b7736666280b9

src/orion/core/worker/primary_algo.py
lebrice: Fixed in eebe53ca24d35db5b9cb917c3a319e4755d41266

src/orion/core/worker/primary_algo.py
lebrice: Fixed in eebe53ca24d35db5b9cb917c3a319e4755d41266

tests/unittests/algo/test_tpe.py
lebrice: Fixed in 44b9a6bf777716b7aca388ee1bfbf7b87d8ec476

tests/unittests/client/test_experiment_client.py
lebrice: Fixed in f1af73ef03984b341ead602cad188e559b6694fb

tests/unittests/core/test_primary_algo.py
lebrice: Fixed in 44b9a6bf777716b7aca388ee1bfbf7b87d8ec476

src/orion/testing/algo.py
lebrice: Fixed in 52ba2515be09d9ebd5e500dfb10b28a34851f978

src/orion/core/worker/primary_algo.py
lebrice: I guess I'll only call `register` if `has_suggested` is false, would that be better, you think?

src/orion/algo/hyperband.py
lebrice: There's a little catch: Here we ignore any `max_trials`, but we don't in the BaseAlgorithm. How could we then use super().is_done here? See what I mean?

src/orion/algo/hyperband.py
lebrice: Fixed in 4680ca57fe0c708393889463c7b326f20b4ee386

src/orion/algo/hyperband.py
lebrice: Fixed in 2c1f8c1dc08715e867ffa436e97245fc083630e7

src/orion/algo/pbt/pbt.py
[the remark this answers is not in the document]
lebrice: Interesting! The annotation that I had for the return value of `_generate_offspring` was `tuple[Trial | None, Trial | None]`, which isn't as tight as it can be!
Instead, I've typed it with `tuple[Trial, Trial] | tuple[None, None]`.

The reason for this assertion is, the type checker isn't yet smart enough to understand that if we went through the if above and that therefore `trial_to_branch` is of type `Trial`, then the second tuple item is also non-None. If it were smart enough to figure that out, then I'd gladly remove the assert. But for now, it isn't.

I think the note above the assert is enough, since the users will never see this. This is just for convenience and the type checker, hence no strings, and no friendly messages. 

"Fixed" in 10835fdfbd9aab74c6f427592a89e7bb721d40e9

src/orion/algo/registry.py
lebrice: Oops, wrong commit, but it's fixed: https://github.com/lebrice/orion/blob/eebe53ca24d35db5b9cb917c3a319e4755d41266/src/orion/algo/registry.py#L45

src/orion/algo/registry.py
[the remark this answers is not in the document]
lebrice: Wrong commit id, but it's fixed: https://github.com/lebrice/orion/blob/eebe53ca24d35db5b9cb917c3a319e4755d41266/src/orion/algo/registry.py#L109

src/orion/algo/registry.py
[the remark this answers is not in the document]
lebrice: Fixed in 70dfd2d6468c8300d7f1fde79d4b3d20b03dca87 
(or just in the file here: Fixed in 70dfd2d6468c8300d7f1fde79d4b3d20b03dca87)

src/orion/algo/registry.py
lebrice: Fixed: https://github.com/lebrice/orion/blob/70dfd2d6468c8300d7f1fde79d4b3d20b03dca87/src/orion/algo/registry.py#L126

src/orion/algo/tpe.py
lebrice: This isn't for the user. It's for the devs.Since we already check for the Integer dim, there's no point in checking again, right?
This assert is the result of loose typing on the Dimension objects, really. If the annotation for `interval` on `Integer` gave `tuple[int, int]`, then I wouldn't mind taking out this assert.
Can we mark this as resolved?

src/orion/algo/tpe.py
lebrice: I disagree. The code is explicit enough already, re-writing it in english next to it isn't good.

src/orion/core/io/experiment_builder.py
lebrice: Fixed in 52ba2515be09d9ebd5e500dfb10b28a34851f978

src/orion/core/worker/primary_algo.py
lebrice: Not sure I understand. Do you want the RegistryMapping to save-restore the original and transformed registries? or only `_mapping` ?

src/orion/algo/tpe.py
lebrice: Can we mark this as resolved please? Not going to budge on this. The TypeError is much less explicit than the assertion code itself.

src/orion/algo/hyperband.py
bouthilx: Oh I see. Maybe we should break is_done down internally into 2 methods like _has_exhausted_cardinality and _has_reached_max_trials. BaseAlgo would use both in is_done, but child classes could reuse one or the other if necessary.

src/orion/algo/hyperband.py
bouthilx: Alright, and for assert we leave it like that and defer the discussion here: https://github.com/Epistimio/orion/issues/859

src/orion/algo/registry.py
bouthilx: And set_state?https://github.com/lebrice/orion/blob/eebe53ca24d35db5b9cb917c3a319e4755d41266/src/orion/algo/registry.py#L49

src/orion/algo/tpe.py
bouthilx: Yep, deferring discussion here: https://github.com/Epistimio/orion/issues/859

src/orion/algo/tpe.py
bouthilx: Deferring discussion here: https://github.com/Epistimio/orion/issues/859

src/orion/core/worker/primary_algo.py
bouthilx: Only `_mapping`. Other option is to not support `set_state` and force the creation of a new `RegistryMapping(state=state_dict['registry_mapping'])`. Where `state_dict['registry_mapping'] = self.registry_mapping.state_dict`.

src/orion/core/worker/primary_algo.py
bouthilx: We should make sure to add a test for this since the issue was not detected.

src/orion/core/worker/primary_algo.py
bouthilx: If we do this then we never get the trial result registered in the register and when we suggest duplicates of completed trials we will never observe these results because we ignored them when we observed the original trial (we did not register the trial with results since it was already in the registry). We should have tests for this situation.

src/orion/core/worker/primary_algo.py
bouthilx: I'm not sure. I feel like this should be handled by the algorithm itself. If it cannot support observing trials it has not suggested, it should not be fooled by registering the trials externally so that has_suggested() is True.

tests/conftest.py
bouthilx: Can you please mess with a test in orion.testing and see if the assert output is the verbose one from pytest or the basic one from python?

src/orion/algo/hyperband.py
lebrice: Ok makes sense, I'll try something like that.

src/orion/algo/registry.py
lebrice: Fixed in 15ea577e2f3f8f8407a4a311d2e4d39f8e11056e (not pushed atm)

src/orion/core/worker/primary_algo.py
lebrice: Fixed in 1be4bfa21b3565f345ebcab94140d31ecc7278f5 (not pushed atm)

src/orion/core/worker/primary_algo.py
lebrice: Agreed. Do you already have an idea of what this test would do, by any chance?

Here's what I'm thinking so far:
- Setup the algo with some completed trials.
- Get a suggested trial from the algo, and somehow make it match an existing trial when translated back into the original space (e.g. through a pretty aggressive `Precision` transform?)
- Make sure that the algo observes the trial it suggested, with the results and status of the existing trial.

src/orion/core/worker/primary_algo.py
lebrice: IMO calling `self.algorithm.register` isn't too intrusive (but `self.algorithm.registry.register` would be), since algos can overwrite `register` to only register trials they already have suggested, if they want to.

Anyhoo, what should I be doing here?

tests/conftest.py
lebrice: With this ordering, we now get the verbose output:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_asha.TestGenericASHA object at 0x7f1d6de97640>, config = {'num_brackets': 2, 'num_rungs': 5, 'repetitions': 3, 'seed': 123456}
space = None, kwargs = {}, base_algo_type = <class 'orion.algo.asha.ASHA'>
original_space = Space([Fidelity(name=f, low=1, high=10, base=2),
       Real(name=x, prior={uniform: (0, 1), {}}, shape=(), default value=None),
       Real(name=y, prior={uniform: (0, 1), {}}, shape=(), default value=None)])
algo = <orion.core.worker.primary_algo.SpaceTransformAlgoWrapper object at 0x7f1d6d9b0040>
@py_assert1 = <orion.algo.asha.ASHA object at 0x7f1d6dad96a0>, @py_assert3 = 200, @py_assert6 = 'BOB', @py_assert5 = False

    def create_algo(self, config=None, space=None, **kwargs):
        """Create the algorithm based on config.
    
        Parameters
        ----------
        config: dict, optional
            The configuration for the algorithm. ``self.config`` will be used
            if ``config`` is ``None``.
        space: ``orion.algo.space.Space``, optional
            Space object to pass to algo. The output of ``self.create_space()``
            will be used if ``space`` is ``None``.
        kwargs: dict
            Values to override algorithm configuration.
        """
        config = copy.deepcopy(config or self.config)
        config.update(kwargs)
        base_algo_type = orion.algo.base.algo_factory.get_class(self.algo_name)
        original_space = space or self.create_space()
        algo = create_algo(space=original_space, algo_type=base_algo_type, **config)
        algo.algorithm.max_trials = self.max_trials
>       assert algo.algorithm.max_trials == "BOB"
E       AssertionError: assert 200 == 'BOB'
E        +  where 200 = <orion.algo.asha.ASHA object at 0x7f1d6dad96a0>.max_trials
E        +    where <orion.algo.asha.ASHA object at 0x7f1d6dad96a0> = <orion.core.worker.primary_algo.SpaceTransformAlgoWrapper object at 0x7f1d6d9b0040>.algorithm

src/orion/testing/algo.py:186: AssertionError
```

With the other ordering, we still get the rich output! However, there is an additional warning logged:
```
(...)
>       assert algo.algorithm.max_trials == "BOB"
E       AssertionError: assert 200 == 'BOB'
E        +  where 200 = <orion.algo.asha.ASHA object at 0x7f4dcfa9cb80>.max_trials
E        +    where <orion.algo.asha.ASHA object at 0x7f4dcfa9cb80> = <orion.core.worker.primary_algo.SpaceTransformAlgoWrapper object at 0x7f4dcf9a7370>.algorithm

src/orion/testing/algo.py:186: AssertionError
=============================================================== warnings summary ================================================================
tests/conftest.py:31
  /home/fabrice/Source/orion/tests/conftest.py:31: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: orion.testing
    pytest.register_assert_rewrite("orion.testing")
```

src/orion/algo/hyperband.py
lebrice: Fixed in 1aedf7b34ff5cca1e282c778a2892e07634f6daa

src/orion/core/worker/primary_algo.py
bouthilx: Sounds like a plan! :smile:

src/orion/core/worker/primary_algo.py
bouthilx: Sorry I got confused with `observe()`. Yes, what you propose is good here in `suggest()`, only register if the original was not already know.

src/orion/core/worker/primary_algo.py
bouthilx: But they typically register the trial when they suggest it, that's how they know latter on if the suggested the trial. Since we call self.algorithm.observe(trial) I think we do not need to call self.algorithm.register(trial) beforehand. The algo should be responsible of it inside observe.

tests/conftest.py
bouthilx: Oh, the line above is doing the import already. Ok, both orders are fine. Thanks for verifying!

src/orion/core/worker/primary_algo.py
lebrice: Can we mark this as resolved?

src/orion/core/worker/primary_algo.py
lebrice: Can we mark this as resolved?

bouthilx (approved):
Thanks a lot for this great PR!

src/orion/core/worker/primary_algo.py
bouthilx: Fixed in https://github.com/Epistimio/orion/pull/833/commits/3c7718784dd1036a8126f87b470acca0220043a1

PR #849, opened 2022-03-22
Add DEHB

### LINKED ISSUE #840: Migrate orion.algo.dehb

bouthilx:
Fixes #840


src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+                self.observe_one(trial)
+                logger.debug(
+                    "Observe trial %s (Remaining %d)", trial, len(self.job_infos)
+                )
+
+    def observe_one(self, trial: Trial) -> None:
+        """Observe a single trial"""
+
+        # Get all the job sampled by DEHB, it might be more than one
+        job_infos = self.job_infos.get(self.get_id(trial), [])
```
lebrice, line 460: Note to self, this kind of stuff (using `self.get_id`) might get impacted by #833 . Need to double-check.

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+
+        if len(job_infos) == 0:
```
lebrice, line 462:
```suggestion
        if not job_infos:
```

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+        # Store the result for later, if we sample
+        # a trial that is too alike for us to evaluate
+        results = cost, fitness
+        self.job_results[self.get_id(trial)] = results
+
+        for job_info in job_infos:
+            cost = job_info["budget"]
+            self._dehb_observe(job_info, cost, fitness)
+
+    def _dehb_observe(self, job_info, cost, fitness):
```
lebrice, line 481: Nit: add type annotations

src/orion/algo/dehb/logger.py
```diff
@@ -0,0 +1,27 @@
...
+    """Fake logger to prevent file creation by loguru"""
+
+    def __getattr__(self, _item):
+        """always return something that does nothing"""
+        return passmethod
+
+
+def remove_loguru():
+    """Override loguru logger class to ignore all its call"""
+    # TODO: wtf is that?
```
lebrice, line 20: Nit: remove

lebrice (changes requested):

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+        space: Space = None,
+        seed: Optional[int] = None,
+        mutation_factor: float = 0.5,
+        crossover_prob: float = 0.5,
+        mutation_strategy: str = "rand1",
+        crossover_strategy: str = "bin",
+        boundary_fix_type: str = "random",
+        min_clip: Optional[int] = None,
+        max_clip: Optional[int] = None,
+        max_age: Optional[int] = np.inf,
```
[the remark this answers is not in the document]
lebrice: Also, nitpick, and totally optional, but I personally prefer using the newer, more intuitive syntax for types, like using `int | float | None`, and such.
To enable that, use `from __future__ import annotations`. Then you can also use `list[int]` instead of `List[int]`, `dict[str, Trial]` instead of `Dict[str, Trial]`, etc.

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+    def _initialize(self) -> None:
+        # Convert to configpace
+
+        self.cs = convert_space(self.space)
+
+        dimensions = len(self.cs.get_hyperparameters())
+
+        # Initialize
+        self.seed_rng(self.seed)
+        DEHBImpl.__init__(
```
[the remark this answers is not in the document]
bouthilx: Fixed in 412e0c8fbfcb18883d361cafa39b5292bcef6cc5

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+            raise UnsupportedConfiguration(
+                f"Boundary fix type {boundary_fix_type} not supported"
+            )
+
+        # We need the transformed algo to initialize DEHB
+        # so store the arguments for after the constructor
+
+        self._original = space
+        BaseAlgorithm.__init__(
+            self,
```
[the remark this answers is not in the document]
bouthilx: Fixed in 412e0c8fbfcb18883d361cafa39b5292bcef6cc5

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+        self.job_infos = defaultdict(list)
+        self.job_results = dict()
+        self.duplicates = defaultdict(int)
+
+    def f_objective(self, *args, **kwargs):
+        """Not needed for Orion, the objective is called by the worker"""
+        pass
+
+    @property
+    def space(self) -> Space:
```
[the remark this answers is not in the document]
bouthilx: Fixed in 412e0c8fbfcb18883d361cafa39b5292bcef6cc5

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+        """Seed the state of the random number generator.
+
+        Parameters
+        ----------
+        seed: int
+            Integer seed for the random number generator.
+
+        """
+        np.random.seed(seed)
+        if hasattr(self, "cs"):
```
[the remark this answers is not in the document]
bouthilx: Yeah we could now that we are sure the test-suite is verifying this. We did so before to make sure we would not forget to seed algorithms.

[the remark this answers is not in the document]
bouthilx: Created issue here: https://github.com/Epistimio/orion/issues/880
But addressed in 412e0c8fbfcb18883d361cafa39b5292bcef6cc5 anyhow.

src/orion/algo/dehb/dehb.py
bouthilx: Fixed in 412e0c8fbfcb18883d361cafa39b5292bcef6cc5. I needed to keep it because the algorithm we wrap may actually suggest identical trials within the transformed space so we still need to handle them that way. I adjusted the note about duplicates however since the source of duplicates is just for discrete dimensions and not because of transformations.

src/orion/algo/dehb/brackets.py
```diff
@@ -0,0 +1,42 @@
...
+
+    IMPORT_ERROR = None
+except ImportError as exc:
+    IMPORT_ERROR = exc
+
+
+logger = logging.getLogger(__name__)
+
+
+class SHBracketManager(SHBracketManagerImpl):
```
lebrice, line 21: This will cause an error if the try-import doesn't succeed, since `SHBracketManagerImpl` won't be defined. Same with the other try-import in the other file.

One solution is to just declare a dummy class for SHBracketManagerImpl inside the `except` block.

tests/unittests/algo/test_dehb.py
```diff
@@ -0,0 +1,126 @@
...
+import itertools
+
+import numpy
+import pytest
+
+from orion.algo.dehb.dehb import IMPORT_ERROR, UnsupportedConfiguration
+from orion.core.utils import backward, format_trials
+from orion.testing.algo import BaseAlgoTests
+
+if IMPORT_ERROR:
```
lebrice, line 12: I don't really like this IMPORT_ERROR solution. Could you try using `pytest.importorskip` instead?
If it doesn't work, or makes the code ugly, then I guess leave this as it is.

tests/unittests/algo/test_dehb.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+# These are the total number of suggestions that the algorithm will make
+# for each "phase" (including previous ones).
+# The maximum number is 32 and then it will be done and stop suggesting mode.
+# COUNTS = [8 + 4 * 3, 4 + 2 + 4, 2 + 1, 1]
+COUNTS = [8 + 4 * 3, 4 + 2 + 4]
+COUNTS = numpy.cumsum(COUNTS)
+
+TestDEHB.set_phases(
```
lebrice, line 123: Add the phases using the new syntax instead:
```python
phases: ClassVar[list[TestPhase]] = [
    TestPhase("random", 0, "space.sample"),
    ...
]
```

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+                setattr(self, k, v)
+            else:
+                logger.error("DEHB does not have attribute %s", k)
+
+        np.random.set_state(state_dict["numpy_GlobalState"])
+        self.cs.random.set_state(state_dict["numpy_RandomState"])
+
+
+# pylint: disable=too-many-public-methods
+class DEHB(DEHBImpl, BaseAlgorithm):
```
lebrice, line 191: Question: Why does this still need to inherit from `DEHBImpl`, even after factoring out the `DEHBImpl` portion into the `dehb` attribute?
Ideally this would only subclass `BaseAlgorithm`, and call `super().__init__(space, ...)`.

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+
+        """
+        self.dehb.seed(seed)
+
+    @property
+    def is_done(self) -> bool:
+        """Return True, if an algorithm holds that there can be no further improvement."""
+        return self.dehb._is_run_budget_exhausted(None, self.rung, None)
+
+    def sample_to_trial(self, sample: np.array, fidelity: int) -> Trial:
```
lebrice, line 370: Just a heads-up, you might get some docs errors by using `np` in annotations instead of `numpy`. Try it like this, and if it works, then great! Otherwise, just change everything from `np` to `numpy`.

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+        """Convert a ConfigSpace sample into a trial"""
+        config = self.vector_to_configspace(sample)
+        hps = OrderedDict()
```
lebrice, line 373: No need since python 3.7. Just use `dict` instead:
```suggestion
        hps = {}
```

tests/unittests/algo/test_dehb.py
```diff
@@ -0,0 +1,126 @@
...
+    def test_suggest_n(self, mocker, num, attr):
+        algo = self.create_algo()
+        spy = self.spy_phase(mocker, num, algo, attr)
+        trials = algo.suggest(3)
+        assert len(trials) == 3
+
+    @pytest.mark.xfail
+    def test_is_done_cardinality(self):
+        """Fails because of https://github.com/Epistimio/orion/issues/598"""
+        space = self.update_space(
```
lebrice, line 61: Question: In this test, why is it doing `space = self.update_space` if it's going to overwrite all the dimensions of the space anyway?

tests/unittests/algo/test_dehb.py
```diff
@@ -0,0 +1,126 @@
...
+        assert i + 1 == space.cardinality
+
+        assert algo.is_done
+
+    @pytest.mark.skip(reason="See https://github.com/Epistimio/orion/issues/599")
+    def test_optimize_branin(self):
+        pass
+
+    # pylint: disable=arguments-differ
+    @pytest.mark.parametrize("num", [100000, 1])
```
lebrice, line 92: Why such a large number? Does it make any difference?
If so, maybe just add a comment to explain why such a large number is alright in this case.

src/orion/algo/dehb/dehb.py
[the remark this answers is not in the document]
lebrice: Not quite: See the other comment: https://github.com/Epistimio/orion/pull/849/files#r840940112

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+        Max clip when boundary fix method is clip
+        Default: ``None``
+
+    max_age: Optional[int]
+        Default: ``np.inf``
+
+    """
+
+    requires_type: ClassVar[Optional[str]] = None
+    requires_dist = None
```
lebrice, line 246: Can you also add the type variables for the other two class attributes as well?
```suggestion
    requires_dist: ClassVar[Optional[str]] = None
```

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+    requires_shape = "flattened"
```
lebrice, line 247:
```suggestion
    requires_shape: ClassVar[Optional[str]] = "flattened"
```

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+
+        if boundary_fix_type not in FIX_MODES:
+            raise UnsupportedConfiguration(
+                f"Boundary fix type {boundary_fix_type} not supported"
+            )
+
+        # We need the transformed algo to initialize DEHB
+        # so store the arguments for after the constructor
+
+        self._original = space
```
lebrice, line 282: Not true anymore! This might be misleading, I suggest we adapt this so it doesn't have this kind of two-stage init anymore.

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,509 @@
...
+from collections import OrderedDict, defaultdict
+from copy import deepcopy
+from typing import List, Optional
+
+import numpy as np
+
+from orion.algo.dehb.brackets import SHBracketManager
+from orion.algo.dehb.logger import remove_loguru
+
+remove_loguru()
```
lebrice, line 19:
```suggestion
```
(since it's a no-op anyway)

src/orion/algo/dehb/dehb.py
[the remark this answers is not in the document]
bouthilx: Looks better indeed! I addressed these comments in ebfff61c98c062202355f2ce3c308661e9fc80a1. I removed `max_age` from the arguments because I was unable to find out what it is used for, it is not documented in source code nor mentioned in the paper.

src/orion/algo/dehb/dehb.py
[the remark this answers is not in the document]
bouthilx: Oopsie! Fixed in 14b4f947

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+            n_workers=1,
+            client=None,
+            # No need for the user function
+            f=None,
+        )
+
+        self.rung = None
+        self.cs = None
+        self.seed = seed
+        self.job_infos = defaultdict(list)
```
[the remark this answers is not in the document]
bouthilx: Fixed in ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+            eta=fidelity_dim.base,
+            # Disable their Dask Integration
+            n_workers=1,
+            client=None,
+            # No need for the user function
+            f=None,
+        )
+
+        self.rung = None
+        self.cs = None
```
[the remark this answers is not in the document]
bouthilx: No more present in code.

src/orion/algo/dehb/dehb.py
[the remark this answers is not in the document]
bouthilx: And https://github.com/Epistimio/orion/commit/14b4f947407cc2a2c60f77b8a091a74bc0a792c1

src/orion/algo/dehb/dehb.py
```diff
@@ -0,0 +1,513 @@
...
+            max_budget=fidelity_dim.high,
+            eta=fidelity_dim.base,
+            # Disable their Dask Integration
+            n_workers=1,
+            client=None,
+            # No need for the user function
+            f=None,
+        )
+
+        self.rung = None
```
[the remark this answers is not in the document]
bouthilx: Fixed in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/dehb.py
bouthilx: Factored out in https://github.com/Epistimio/orion/commit/412e0c8fbfcb18883d361cafa39b5292bcef6cc5 and then typed in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/logger.py
bouthilx: Removed in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/brackets.py
bouthilx: Added in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

tests/unittests/algo/test_dehb.py
bouthilx: Not sure what is better. Otherwise we end up with something like this
```python
dehb_module = pytest.importorskip('orion.algo.dehb.dehb', reason='skipping DEHB tests')
UnsupportedConfiguration = dehb_module.UnsupportedConfiguration
```
I don't think that's an improvement over current version.

tests/unittests/algo/test_dehb.py
bouthilx: Indeed! Fixed in 7ec37866

src/orion/algo/dehb/dehb.py
bouthilx: It was a mistake indeed, I just forgot to remove it. Fixed in https://github.com/Epistimio/orion/commit/14b4f947407cc2a2c60f77b8a091a74bc0a792c1

src/orion/algo/dehb/dehb.py
bouthilx: I add to change this indeed, thanks!

tests/unittests/algo/test_dehb.py
bouthilx: Oh, first `space` is a configuration, a dictionary. I improved the var names in aeb1d570.

tests/unittests/algo/test_dehb.py
bouthilx: Yeah, that's quite large. :) It's good to have a buffer if the algo samples many duplicates but it probably does not need to be that large.

src/orion/algo/dehb/dehb.py
bouthilx: Added in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/dehb.py
bouthilx: Added in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

src/orion/algo/dehb/dehb.py
bouthilx: Fixed in b5775969

src/orion/algo/dehb/dehb.py
bouthilx: Removed in https://github.com/Epistimio/orion/commit/ebfff61c98c062202355f2ce3c308661e9fc80a1

tests/unittests/algo/test_dehb.py
lebrice: No I understand, what I'm asking is, why is the `space_dict` created using `self.update_space`, instead of just a dict? (Since it's overwriting all the entries in `type(self).space` anyway)?

tests/unittests/algo/test_dehb.py
lebrice: I agree though, it's still uglier. Marking this as resolved.

tests/unittests/algo/test_dehb.py
bouthilx: Fixed in 8b1bace

PR #853, opened 2022-03-23
Drop support for python 3.6

lebrice:
Title.

This is going to enable us to use nice type hints everywhere thanks to https://peps.python.org/pep-0563/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       max-parallel: 4
       matrix:
         platform: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


bouthilx:
I suggest we bump all tests to py3.9 by default. They are currently at py3.7.

lebrice:
I'd rather do that in another PR, in the interest of time. Is that ok with you @bouthilx ?

lebrice:
Created another pr for that: #855

bouthilx (approved):
All good, thanks!

PR #855, opened 2022-03-24
Upgrade default python to 3.9 in build.yaml

lebrice:
Is this sufficient @bouthilx ? Or do I need to change something else?


bouthilx (approved):
I think that's sufficient, thanks! :)

PR #864, opened 2022-03-28
Add MOFA

wagnersj:
# Description
Adds the MOFA algorithm to Orion.

## Tests
- [X] All new and existing tests are passing (`$ tox -e py38`; replace `38` by your Python version if necessary)

## Documentation
- [X] I have updated the relevant documentation related to my changes

## Quality


src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,390 @@
...
+        while (
+            len(trials) < num
+            and len(self.current_trials_params) > 0
+            and not self.is_done
+        ):
+            trial_params = self.current_trials_params.pop()
+            trial_params.update(self.frozen_param_values)
+            trial = dict_to_trial(trial_params, self.space)
+            if self.has_observed(trial):
+                similar_trial, _ = self._trials_info[self.get_id(trial)]
```
lebrice, line 209: Note to self: This will be impacted by #833

src/orion/algo/mofa/sampler.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Samples trial parameters from an orthogonal Latin hypercube.
+    """
+
+    def __init__(self):
+        return
+
+    @staticmethod
+    def _is_prime(number):
+        if number < 2 or (number % 2 == 0 and number > 2):
```
lebrice, line 26: 2 is prime, and as far as I can tell, this will not work, but instead raise an error with the `range(3, 2, 2)` below.
How about something a bit more explicit:
```suggestion
        if number == 2:
            return True
        if number < 2 or number % 2 == 0:
            return False
```

src/orion/algo/mofa/sampler.py
```diff
@@ -0,0 +1,119 @@
...
+
+import logging
+
+import numpy as np
+from scipy.stats.qmc import LatinHypercube
+
+logger = logging.getLogger(__name__)
+
+
+class Sampler:
```
lebrice, line 16: A class with only static methods is a bit of a code smell. Is the class even needed at this point?
You could probably remove the class, and use the module instead as the base, e.g.:

From this:
```python
from orion.algo.mofa.sampler import Sampler
foo = Sampler.generate_trials(...)
```
to this:
```python
from orion.algo.mofa import sampler
foo = sampler.generate_trials(...)
```

src/orion/algo/mofa/sampler.py
lebrice: Nevermind, range(3,2,2) doesn't raise an error, it's just an empty iterable, and so this doesn't go through the for loop, it just returns `True`, as it should.

tests/unittests/algo/mofa/test_mofa.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_config_threshold_le0_isnot_valid(self):
+        with pytest.raises(ValueError):
+            self.create_algo(config={"threshold": 0})
+
+    def test_config_threshold_ge1_isnot_valid(self):
+        with pytest.raises(ValueError):
+            self.create_algo(config={"threshold": 1})
+
+    @pytest.mark.skip(
+        reason="MOFA converges too fast and does not observe the whole space"
```
lebrice, line 61: Haha I guess that's a good thing then! :)

lebrice (commented):
:+1:

src/orion/algo/mofa/sampler.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/8e12929ff7d85949bc4b8743f4e6baa105f73482

src/orion/algo/mofa/mofa.py
lebrice: In this particular case, you'd want to call `similar_trial = self.registry.get_existing(trial)`, or `self.registry[self.get_id(trial)]`.
You might want to take a quick look at the `Registry` class, which is in `orion.algo.registry`. It's essentially a container of Trials, which can be indexed using the trial IDs.
Anywhere where you might have been doing something like `for id, trial in self._trials_info.items()`, you should now be doing `for trial in self.registry`, basically.

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,389 @@
...
+        self.roi_space = None
+        self.completed_trials = []
+        self.duplicates = defaultdict(list)
+        self.converged = False
+        self.n_trials = None
+
+        # Initialize a dictionary for the frozen values of each hyperparameter
+        self.frozen_param_values = {}
+
+    def _initialize(self):
```
lebrice, line 106: This block can now be done directly in `__init__`:

```suggestion
```

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,389 @@
...
+            out if it cannot make a good suggestion at the moment (it may be waiting for other
+            trials to complete), in which case it will return None.
+
+
+        Notes
+        -----
+        New parameters must be compliant with the problem's domain `orion.algo.space.Space`.
+
+        """
+        if not hasattr(self, "current_trials_params"):
```
lebrice, line 192: You can remove this now, the space you get in `__init__` is the transformed space, so no need for this two-step initialization.

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+        factorial_importance_analysis.append((key, best_level, importance))
+
+    return pd.DataFrame(
+        factorial_importance_analysis, columns=["param", "best_level", "importance"]
+    )
+
+
+def select_new_region_of_interest(
+    factorial_importance_analysis, space, threshold, n_levels
```
lebrice, line 358: Nitpick:
```suggestion
    factorial_importance_analysis: dict[str, float], space: Space, threshold: float, n_levels: int
```
Btw add `from __future__ import annotations` as the first line if you want to add this.

bouthilx:
About `test_is_done_max_trials()`, one solution would be to avoid testing for `max_trials` within MOFA since it is already done within the wrapper, and the wrapper holds the ground truth in terms of how many trials have been actually executed and completed. That would mean removing the part `or super().is_done` in `MOFA.is_done` property.

src/orion/algo/mofa/_version.py
```diff
@@ -0,0 +1,677 @@
+# This file helps to compute a version number in source trees obtained from
```
bouthilx, line 1: This file should be removed. That was only relevant for the external plugin, not as part of the core repo.

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,386 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MOFA(BaseAlgorithm):
+    """
+    MOdular FActorial Design (MOFA).
+
+    For more information on the algorithm, see original paper: MOFA: Modular
+    Factorial Design for Hyperparameter Optimization https://arxiv.org/abs/2011.09545
+
```
bouthilx, line 31:
```suggestion
    Factorial Design for Hyperparameter Optimization https://arxiv.org/abs/2011.09545
    
    Xiong, Bo, Yimin Huang, Hanrong Ye, Steffen Staab, and Zhenguo Li.
    "MOFA: Modular Factorial Design for Hyperparameter Optimization."
    arXiv preprint arXiv:2011.09545 (2020).

```

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,386 @@
...
+        if hasattr(self, "current_trials_params"):
+            state_dict["current_trials_params"] = copy.deepcopy(
+                self.current_trials_params
+            )
+        return state_dict
+
+    def set_state(self, state_dict):
+        """Reset the state of the algorithm based on the given state_dict
+
+        :param state_dict: Dictionary representing state of an algorithm
```
bouthilx, line 150:
```suggestion
        Parameters
        ----------
        state_dict: dict
            Dictionary representing state of an algorithm
```

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,386 @@
...
+        num: int, optional
+            Number of trials to suggest. The algorithm may return less than the number of trials
+            requested.
+
+        Returns
+        -------
+        list of trials or None
+            A list of trials representing values suggested by the algorithm. The algorithm may opt
+            out if it cannot make a good suggestion at the moment (it may be waiting for other
+            trials to complete), in which case it will return None.
```
bouthilx, line 183:
```suggestion
        list of trials
            A list of trials representing values suggested by the algorithm. The algorithm may opt
            out if it cannot make a good suggestion at the moment (it may be waiting for other
            trials to complete), in which case it will return an empty list.
```

src/orion/algo/mofa/sampler.py
```diff
@@ -0,0 +1,110 @@
+"""
+MOFA Sampler module
+"""
+
+# from skopt.sampler import Lhs
+# from skopt.space import Space as SkSpace
+
```
bouthilx, line 7:
```suggestion
```

tests/unittests/algo/mofa/test_mofa.py
```diff
@@ -0,0 +1,106 @@
...
+        assert algo.is_done
+        assert len(trials) == space.cardinality
+
+
+n_trials = (
+    TestMOFA.config["index"]
+    * TestMOFA.config["n_levels"] ** TestMOFA.config["strength"]
+)
+
+TestMOFA.set_phases(
```
bouthilx, line 100: This should be moved as part of the class attributes following merge of #874. You can look at one of the algorithms test-suite that @lebrice adapted to see how you should do for MOFA.

docs/src/code/algo/mofa.rst
```diff
@@ -0,0 +1,5 @@
+MOFA Algorithm
+===================
+
+.. automodule:: orion.algo.mofa.mofa
+   :members:
```
bouthilx, line 5:
```suggestion
   :members:

```
To fix doc8 issue.

src/orion/algo/mofa/_version.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/268378e0e47b56f5d9fe35f35fa77a1cb2a99aae

wagnersj:
About `test_is_done_max_trials()`, MOFA does not do checks on `max_trials`. I did try @bouthilx 's suggestion to remove the `super().is_done` check in the `is_done()` function, but that does not solve the problem (actually causes another test to fail).

The `test_is_done_max_trials()` fails with the following message: `RuntimeError: Algorithm cannot sample more than 102 trials. Is it normal?`. This happens because the `BaseAlgoTests.force_observe()` function call within `test_is_done_max_trials()` gets empty trials from the `suggest()`function and eventually sees 5 empty trial returns as a sign of a failure. The empty trials are not being returned by `MOFA.suggest()` but by `SpaceTransformAlgoWrapper.suggest()`.

I've investigated deeper. The problem is that `MOFA` and `SpaceTransformAlgoWrapper` are both checking for duplicates. I think the best way to handle this is to remove the checks in `MOFA` and let `SpaceTransformAlgoWrapper` do it.

src/orion/algo/mofa/mofa.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/b3f58efb4dc69ffc97b85c11b959110c69933d4e

src/orion/algo/mofa/mofa.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/b3f58efb4dc69ffc97b85c11b959110c69933d4e

src/orion/algo/mofa/sampler.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/b3f58efb4dc69ffc97b85c11b959110c69933d4e

src/orion/algo/mofa/mofa.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/8a3e1741227b516bb43852da278a8d8725deae6d

tests/unittests/algo/mofa/test_mofa.py
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/dde480d96ad10bb7251a06eab7f3600cb3a8f530

lebrice:
The test that wasn't working should now work: The force_observe will always observe the required number of trials.

docs/src/code/algo/mofa.rst
wagnersj: Addressed in https://github.com/Epistimio/orion/pull/864/commits/710af4da5865c02381c797fda7cd3b5e24b1c76d

src/orion/algo/mofa/mofa.py
```diff
@@ -0,0 +1,410 @@
...
+        self.frozen_param_values = copy.deepcopy(state_dict["frozen_param_values"])
+        if "current_trials_params" in state_dict:
+            self.current_trials_params = copy.deepcopy(
+                state_dict["current_trials_params"]
+            )
+        elif hasattr(self, "current_trials_params"):
+            self.current_trials_params = []
+
+    def suggest(self, num: int) -> list[Trial]:
+        """Suggest a `num`ber of new sets of parameters.
```
bouthilx, line 185:
```suggestion
        """Suggest a number of new sets of parameters.
```
This also caused issues in other PR of plugin integration. It's hopefully the only issue with doc tests.

wagnersj:
All tests on this PR are passing except for the Python 3.7 tests. This is due to MOFA needing scipy v1.8.0 which is only supported in Python >= 3.8. Are we still planning on supporting Python 3.7? If so, is there a way to for the CI to skip tests on MOFA if Python 3.7 tests are being run?

bouthilx:
It looks like its failing when the extra dependencies of the other libraries are not installed. Could it be depending on a dependency that is not in the core dependencies but is in the extra of another algorithm?

wagnersj:
It looks like the problem in build #1716 is due to changes merged in PR #913 in the `force_observe()` method. There is a failure in a call to the `get_num()` function of the `BaseAlgoTest` class.

tests/unittests/algo/mofa/test_mofa.py
```diff
@@ -0,0 +1,126 @@
...
+
+    n_trials = config["index"] * config["n_levels"] ** config["strength"]
+
+    phases: ClassVar[list[TestPhase]] = [
+        TestPhase("1st-run", 0, "space.sample"),
+        TestPhase("2nd-run", n_trials, "space.sample"),
+        TestPhase("3rd-run", n_trials * 2, "space.sample"),
+    ]
+
+    def get_num(self, num):
```
lebrice, line 61:
```suggestion
    @classmethod
    def get_num(cls, num):
```
This should solve it @wagnersj

lebrice:
Hey @wagnersj : There's a simple conflict to solve in setup.py, and you can just apply the fix I mention above.

I test this locally, checked out develop, merged your branch, fixed the simple conflict in setup.py, and ran the tests for mofa in python 3.9: Everything works!

bouthilx (approved):

PR #867, opened 2022-03-29
Detect _param_names from __init__ signature

### LINKED ISSUE #860: Proposition: Rework how configuration is saved through `BaseAlgorithm.__init__`
The current mechanism for storing the configuration of a custom algorithm is to pass all the kwargs to `super().__init__(**kwargs)`, like so:
```python
def __init__(
        self,
        space: Space,
        n_values: int | dict[str, int] = 100,
        seed: int | Sequence[int] | None = None,
    ):
        super().__init__(space, n_values=n_values, seed=seed)
        self.n = 0
        ...
        something = foo(self.n_values)   # type checker complains that `self.n_values` can't be found

```
The `__init__` of `BaseAlgorithm` then saves the names of the arguments passed in an attribute, and then sets the values on `self`:
```python
def __init__(self, space, **kwargs):
        log.debug(
            "Creating Algorithm object of %s type with parameters:\n%s",
            type(self).__name__,
            kwargs,
        )
        self._space = space
        self._param_names = list(kwargs.keys())
        # Instantiate tunable parameters of an algorithm
        for varname, param in kwargs.items():
            setattr(self, varname, param)

        # TODO: move this inside an initialization function.
        if hasattr(self, "seed"):
            self.seed_rng(self.seed)

        self.registry = Registry()
```

I see one main disadvantage of doing things this way:
- When using attributes set in this way, type checkers complain that the attribute is not defined (since it doesn't have a `self.n_values = n_values`, for example. 

I'd also argue that it seems a bit too much responsibility to give to `BaseAlgorithm.__init__` to save the names, set the values, and initialize the 'base' attributes like `self.registry`, etc.

## Suggestion:

I suggest that we still allow this mechanism, but that we additionally allow saving the attributes directly on `self` instead and discover `_param_names` programmatically, like so:

```python
self._arg_names = list(kwargs or inspect.signature(type(self)).parameters)
```
`_arg_names` will contain the names of all the constructor arguments.

lebrice:
Fixes #860 (just the `__init__` part)

### DIFF
--- a/tests/unittests/algo/test_base.py
+++ b/tests/unittests/algo/test_base.py
@@ -120,11 +120,7 @@ def test_arg_names(pass_to_super: bool):
 
     Also checks that the auto-generated configuration dict acts the same way.
     """
-    # NOTE: If there is a **kwargs argument in __init__ signature, and it is not passed to the
-    # super().__init__(self, **kwargs), then what should the _arg_names be? Should we just raise a
-    # warning in that case? Or just assume that the algo will create its own configuration dict?
-    # (could also raise a warning only inside the `configuration` property, in the case where we
-    # can't fetch the value of the kwargs argument)
+
     class SomeAlgo(BaseAlgorithm):
         def __init__(self, space, foo: int = 123, bar: str = "heyo"):
             if pass_to_super:
--- a/tests/unittests/algo/test_base.py
+++ b/tests/unittests/algo/test_base.py
@@ -2,6 +2,9 @@
 # -*- coding: utf-8 -*-
 """Example usage and tests for :mod:`orion.algo.base`."""
 
+import pytest
+
+from orion.algo.base import BaseAlgorithm
 from orion.algo.space import Integer, Real, Space
 from orion.core.utils import backward, format_trials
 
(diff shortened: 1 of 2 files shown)


tests/unittests/algo/test_base.py
```diff
@@ -108,3 +111,39 @@ def test_is_done_max_trials(monkeypatch, dumbalgo):
...
+    """Test that the `_arg_names` can be determined programmatically when the args aren't passed to
+    `super().__init__(space, **kwargs)`.
+
+    Also checks that the auto-generated configuration dict acts the same way.
+    """
+    # NOTE: If there is a **kwargs argument in __init__ signature, and it is not passed to the
+    # super().__init__(self, **kwargs), then what should the _arg_names be? Should we just raise a
+    # warning in that case? Or just assume that the algo will create its own configuration dict?
+    # (could also raise a warning only inside the `configuration` property, in the case where we
+    # can't fetch the value of the kwargs argument)
```
bouthilx, line 127: Algo configuration is expected to be compliant with algorithm's constructor signature, so that we can build an algorithm with `AlgoClass(space, **configuration)`. I think we could safely ignore any of kwargs and rely on the signature. 

Also, setting the attributes by passing them to super() is always very confusing to new contributiors. I would be in favor of deprecating this and force algos to set their attributes themselves using `self.attr = attr`.

lebrice: Ok, so you're ok with these changes then? (I'll remove the comment)

bouthilx (approved):

PR #883, opened 2022-04-06
Add number of suggest() attempts in Algo wrapper

### LINKED ISSUE #882: Algo wrapper may return no trials even if algo is not done
See problem discussed here: https://github.com/Epistimio/orion/pull/864#issuecomment-1086325806

lebrice:
Fixes #882 (at least I think it does)

### DIFF
--- a/src/orion/core/worker/primary_algo.py
+++ b/src/orion/core/worker/primary_algo.py
@@ -180,8 +180,8 @@ class SpaceTransformAlgoWrapper(BaseAlgorithm, Generic[AlgoType]):
                     self.registry.register(original)
                     trials.append(original)
 
-                # NOTE: Here we DON'T register the transformed trial, we let the algorithm do it itself
-                # in its `suggest`.
+                # NOTE: Here we DON'T register the transformed trial, we let the algorithm do it
+                # itself in its `suggest`.
                 # Register the equivalence between these trials.
                 self.registry_mapping.register(original, transformed_trial)
 
--- a/src/orion/core/worker/primary_algo.py
+++ b/src/orion/core/worker/primary_algo.py
@@ -137,48 +137,72 @@ class SpaceTransformAlgoWrapper(BaseAlgorithm, Generic[AlgoType]):
         New parameters must be compliant with the problem's domain `orion.algo.space.Space`.
 
         """
-        transformed_trials = self.algorithm.suggest(num)
-
-        if transformed_trials is None:
-            return None
 
         trials: list[Trial] = []
-        for transformed_trial in transformed_trials:
-            if transformed_trial not in self.transformed_space:
-                raise ValueError(
-                    f"Trial {transformed_trial.id} not contained in space:\n"
-                    f"Params: {transformed_trial.params}\n"
-                    f"Space: {self.transformed_space}"
-                )
-            original = self.transformed_space.reverse(transformed_trial)
-            if original in self.registry:
-                logger.debug(
-                    "Already have a trial that matches %s in the registry.", original
-                )
-                # We already have a trial that is equivalent to this one.
-                # Fetch the actual trial (with the status and possibly results)
-                original = self.registry.get_existing(original)
-                logger.debug("Matching trial (with results/status): %s", original)
+        max_suggest_attempts = 100
 
-                # Copy over the status and results from the original to the transformed trial
-                # and observe it.
-                transformed_trial = _copy_status_and_results(
-                    original_trial=original, transformed_trial=transformed_trial
-                )
-                logger.debug(
-                    "Transformed trial (with results/status): %s", transformed_trial
-                )
-                self.algorithm.observe([transformed_trial])
-            else:
-                # We haven't seen this trial before. Register it.
-                self.registry.register(original)
-                trials.append(original)
+        # NOTE: Should we also do this kind of repeated checking until the algo produces `num`
+        # trials?
 
-            # NOTE: Here we DON'T register the transformed trial, we let the algorithm do it itself
-            # in its `suggest`.
-            # Register the equivalence between these trials.
-            self.registry_mapping.register(original, transformed_trial)
-        return trials
+        for suggest_attempt in range(1, max_suggest_attempts + 1):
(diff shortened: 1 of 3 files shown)


lebrice:
Hey @bouthilx  there's a single test failing: `tests/functional/algos/test_algos.py::test_with_fidelity[tpe]`
I'm unable to reproduce the failure on my end (with python 3.8, the test consistently passes)..

Would you mind trying it on your end, whenever you have 5 mins, and letting me know if it also works on your end?
Thanks!

bouthilx:
This test seems to be randomly failed since a few weeks. I suspect it is failing for the same unknown reason in this PR which isn't related to your modifications. It would be great if we could fix it soon thought.

src/orion/core/worker/primary_algo.py
```diff
@@ -137,48 +137,72 @@ def suggest(self, num: int) -> list[Trial] | None:
...
-        for transformed_trial in transformed_trials:
-            if transformed_trial not in self.transformed_space:
-                raise ValueError(
-                    f"Trial {transformed_trial.id} not contained in space:\n"
-                    f"Params: {transformed_trial.params}\n"
-                    f"Space: {self.transformed_space}"
-                )
-            original = self.transformed_space.reverse(transformed_trial)
-            if original in self.registry:
+        max_suggest_attempts = 100
```
bouthilx, line 142: This should be configurable.

tests/unittests/core/test_primary_algo.py
```diff
@@ -161,8 +161,16 @@ def __init__(self, space: Space, fixed_suggestion: Trial):
...
     def suggest(self, num):
-        self.register(self.fixed_suggestion)
-        return [self.fixed_suggestion]
+        # NOTE: can't register the trial if it's already here. The fixed suggestion is always "new",
+        # but the algorithm actually observes it at some point. Therefore, we don't overwrite what's
+        # already in the registry.
+        if not self.has_suggested(self.fixed_suggestion):
+            self.register(self.fixed_suggestion)
+            return [self.fixed_suggestion]
+        # TODO: Does it make sense for the algorithm to just not suggest anything, if it has already
```
bouthilx, line 170: We would expect the algorithm not to return duplicates, but the Producer is supposed to be able to handle duplicates if it ever happens. In this case since we are testing the algo only the correct behavior would be to return an empty list after first call to suggest. Does it mess up with some tests? It seems not right?

lebrice: Yeah it doesn't mess up the tests because the `fixed_suggestion` attribute is changed from the outside after each observed trial.

src/orion/core/worker/primary_algo.py
lebrice: As a constructor argument to the wrapper? Or parameter to `suggest` ?

lebrice:
> This test seems to be randomly failed since a few weeks. I suspect it is failing for the same unknown reason in this PR which isn't related to your modifications. It would be great if we could fix it soon thought.

Added a skip here: 8e153ee0295575d1016eb3743bed341bcadd34f1
Also created an issue about this: #887

src/orion/core/worker/primary_algo.py
lebrice: Made it an attribute in 872f094281e8e855f0bf5be05b11b7988131b228, is this sufficient, you think?

src/orion/core/worker/primary_algo.py
```diff
@@ -137,48 +137,72 @@ def suggest(self, num: int) -> list[Trial] | None:
...
+
+        # NOTE: Should we also do this kind of repeated checking until the algo produces `num`
```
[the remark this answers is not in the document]
lebrice: Removed the NOTE in 619d6d659e621b9c8b560419f3f4d6c1f0ab03f0

tests/unittests/core/test_primary_algo.py
lebrice: Removed TODO in f492ffd5529907b1299db77c9ee403cccf39e612

src/orion/core/worker/primary_algo.py
bouthilx: Sorry for the delay. It's a bit tricky here since at the constructor level it would add an argument specific for the wrapper, should I don't think should be exposed to the users since the define the algo configuration directly, they are not even aware that there is a wrapper. Same for suggest, it should follow the same interface as the algorithms it is wrapping. Maybe your solution is the best one for now and we can reassess later on if needed.

bouthilx:
> > This test seems to be randomly failed since a few weeks. I suspect it is failing for the same unknown reason in this PR which isn't related to your modifications. It would be great if we could fix it soon thought.

> Added a skip here: https://github.com/Epistimio/orion/commit/8e153ee0295575d1016eb3743bed341bcadd34f1
> Also created an issue about this: https://github.com/Epistimio/orion/issues/887

Thanks!

src/orion/core/worker/primary_algo.py
```diff
@@ -137,48 +138,68 @@ def suggest(self, num: int) -> list[Trial] | None:
...
-                # Fetch the actual trial (with the status and possibly results)
-                original = self.registry.get_existing(original)
-                logger.debug("Matching trial (with results/status): %s", original)
 
-                # Copy over the status and results from the original to the transformed trial
-                # and observe it.
-                transformed_trial = _copy_status_and_results(
-                    original_trial=original, transformed_trial=transformed_trial
-                )
+        for suggest_attempt in range(1, self.max_suggest_attempts + 1):
```
bouthilx, line 144: I think we should add a test for this new behavior. Not in the test-suite but in the wrapper's set of tests.

lebrice: Fixed in bb73b8272eef6e04aa98230f8680d94233a6ea3f

lebrice:
@bouthilx the Pickleddb tests seem to be failing for some reason? I don't really understand why though, would you mind taking a look?

bouthilx (approved):
All good, thanks!

PR #885, opened 2022-04-13
Keep same extension when created tmp config file

bouthilx:
The users script may not work properly if the config file does not have the expected extension. We should keep the same one instead of turning it into .conf.

@Hnekoeiq

### DIFF
--- a/src/orion/core/worker/consumer.py
+++ b/src/orion/core/worker/consumer.py
@@ -118,9 +118,6 @@ class Consumer(object):
             True if the trial was successfully executed. False if the trial is broken.
 
         """
-        log.debug("Consumer context: %s", trial.working_dir)
-        os.makedirs(trial.working_dir, exist_ok=True)
-
         results_file = self._consume(trial, trial.working_dir)
 
         log.debug("Parsing results from file and fill corresponding Trial object.")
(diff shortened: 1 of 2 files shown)


tests/unittests/core/worker/test_consumer.py
```diff
@@ -79,6 +79,40 @@ def test_trial_working_dir_is_created(config):
...
+
+    assert os.path.exists(trial.working_dir)
+    assert os.path.exists(config_file.name)
+
+    _, original_ext = os.path.splitext(config_path)
+    _, tmp_ext = os.path.splitext(config_file.name)
+
+    assert original_ext == tmp_ext
+
+    shutil.rmtree(trial.working_dir)
```
lebrice, line 113: Why is it necessary to do this? Shouldn't the working directory in the config always be based on the `temp_path` fixture?

src/orion/core/worker/consumer.py
```diff
@@ -197,9 +194,16 @@ def get_execution_environment(self, trial, results_file="results.log"):
 
         return env
 
-    def _consume(self, trial, workdirname):
+    def _prepare_config(self, trial, workdirname):
```
lebrice, line 197: Nit: I know your stance is to add type hints to new modules, but I think new methods might also be a good opportunity to add type hints.

lebrice (approved):

tests/unittests/core/worker/test_consumer.py
bouthilx: It's because of this line: https://github.com/Epistimio/orion/pull/885/files#diff-5fe3b7fe336f5a8a566e677336db8fc5f708b6da4e5662fef48c7bfc95fc1c42R34. We could perhaps convert it to `temp_path`.

bouthilx:
Must adapt doc after merging #957

Issue #884, opened 2022-04-11
Refactor / remove `spy_phase` of BaseAlgoTests


lebrice:
This `spy_phase` doesn't seem very intuitive to me. It makes the tests much more opaque than they need to be.

There are other ways of creating fixtures on classes dynamically, for instance by setting a class attribute in `__init_subclass__`. This would be preferable, in my opinion. 

Here is what I suggest to change:
- Standardize the use of the  `phases` class attribute in all algo tests
- Remove the following:
    - spy_phase
    - spy_attr
    - recursive_getattr
    - recursive_setattr
    - methods_with_phase
    - (etc etc)
- Add a new `phase` fixture that gets parametrized with the value from `cls.phases` and yields the current phase.
- (a bit unsure about this one): Add a new `algo` fixture that subclasses can overwrite that uses this `phase` fixture, OR
- Make `create_algo` observe all the initial trials required for the current phase before returning the algo instance.
    - It might be a good idea to only create one such instance per phase, and then just provide deepcopies to each individual tests, just to make the tests a bit quicker to run?

Any thoughts @bouthilx ?

bouthilx:
I think these are very good ideas. It's not clear to me what you mean with the incremental phases compared to the current phases but that seems to be just a minor detail. Reducing the running time of the tests by sharing the initial algo state would be quite helpful indeed. That will only work if the algo is copied with all internal state properly (like some algos may use global RNG in which case they need to be reseeded). That's pretty bad but cannot be avoided when wrapping some libraries...

lebrice:
Indeed you're right about the incremental phase thing being confusing. I'll edit the post above.

lebrice:
> Reducing the running time of the tests by sharing the initial algo state would be quite helpful indeed

Yeah I was just thinking about it more, Ideally, the fixture would parametrize the test such that the algo state after, say, `10` trials is kept,  and then reused to create the algo state after 20 trials, rather than re-creating a new algo from scratch, and making it observe 20 trials. However, this would require each test to be run through all phases in-sequence before moving to the next test, which isn't how pytest currently does it: It goes through one phase for all tests, then another phase for all tests, etc.

PR #886, opened 2022-04-14
Simplify tests suite, remove spy_phase and related functions

lebrice:
# Description

This PR addresses #884 , by removing the pretty "spy_phase" stuff from the test suite. This makes tests a lot more transparent.

# Changes

- All tests of a BaseAlgorithmTests sublass are now parametrized to run through each of the different test phases.
  - A new `first_phase_only` decorator can be used to disable this behaviour.
  - Running tests in all phases has actually been useful: I've identified a few subtle bugs related to using set_state and get_state after the initial number of trials have been observed.
- a new `phase` fixture can be requested to get the current `TestPhase`.
- `BaseAlgoTests.create_algo` now creates the algo, and makes it observe the right number of trials for the current phase before returning it.

This PR also fixes the bugs that were highlighted by this change in the test suite, for all the algorithms.

## Tests

## Documentation

## Quality


src/orion/algo/hyperband.py
```diff
@@ -296,6 +297,8 @@ def set_state(self, state_dict: dict) -> None:
         self.seed_rng(state_dict["seed"])
         self.rng.set_state(state_dict["rng_state"])
         self.trial_to_brackets = state_dict["trial_to_brackets"]
+        self.budgets = state_dict["budgets"]
+        self.brackets.clear()
         while len(self.brackets) < len(state_dict["brackets"]):
             self.append_brackets()
         assert len(self.brackets) == len(state_dict["brackets"]), "corrupted state"
```
bouthilx, line 304: Is this one of the subtil bugs you referred to in the PR description?

src/orion/algo/pbt/pbt.py
```diff
@@ -377,7 +370,14 @@ def _sample(self, num: int) -> list[Trial]:
 
     def _get_depth_of(self, fidelity: Any) -> int:
         """Get the depth of a fidelity in the lineages"""
-        return self.fidelities.index(fidelity)
+        # BUG: Issue with rounding, asking for fidelity of `10` but list has `10.00000004`.
```
bouthilx, line 373: Is this bug fixed by this if-else?

tests/unittests/algo/test_evolution_es.py
```diff
@@ -520,3 +520,29 @@ def assert_callbacks(
...
+        if _are_equal(new_state, state):
+            # If the state is the same, the trials should be the same.
+            assert a == b
+        else:
+            # If the state is different, the trials should be different.
+            assert a != b
+
+        new_algo.set_state(state)
+        c = new_algo.suggest(1)[0]
+        # TODO: For EvolutionES, the params are identical, but the ids are different.
```
bouthilx, line 546: That's concerning. Did you found out why? Please let me know if you would need help to investigate this.

tests/unittests/algo/test_evolution_es.py
```diff
@@ -504,10 +504,10 @@ def infer_repetition_and_rung(self, num: int):
 
         return 1, budgets.index(num)
 
-    def assert_callbacks(
-        self, spy, num: int, algo: SpaceTransformAlgoWrapper[EvolutionES]
-    ):
-
+    def assert_callbacks(self):
+        # TODO: Move this to wherever it belongs, probably in the cleanup of a fixture.
```
bouthilx, line 508: It's not really a cleanup, it's more like a generic test that is added at the end of the tests. But it is not really necessary to run at end of all tests and could actually be a separate test on its own, using the phase parametrization too.

src/orion/testing/algo.py
```diff
@@ -143,19 +95,69 @@ class BaseAlgoTests:
...
+            cls.algo_type = orion.algo.base.algo_factory.get_class(cls.algo_name)
+        if not cls.algo_name:
+            cls.algo_name = cls.algo_type.__name__.lower()
 
+    @pytest.fixture(name="first_phase")
     @classmethod
-    def set_phases(cls, phases):
+    def first_phase(cls, _phase: TestPhase):
+        if _phase != cls.phases[0]:
+            pytest.skip(reason="Test runs only on first phase.")
```
bouthilx, line 157: That will pollute the test list with a lot of skips. :thinking: It would nice if the phases could be used when a method has the argument `phase` or something like that, like the usual way of using pytest fixtures.

src/orion/testing/algo.py
```diff
@@ -172,54 +174,80 @@ def set_phases(cls, phases):
...
+
+        # Seed the randomness before we observe anything.
+        if seed is not None:
+            algo.seed_rng(seed)
+
+        if cls._current_phase is not None and cls._current_phase.n_trials > 0:
+            n_previous_trials = cls._current_phase.n_trials
+            if n_previous_trials >= cls.max_trials:
+                raise ValueError(
+                    f"Test isn't configured properly: max_trials ({cls.max_trials}) is larger than "
```
bouthilx, line 221:
```suggestion
                    f"Test isn't configured properly: max_trials ({cls.max_trials}) is smaller than "
```

src/orion/testing/algo.py
```diff
@@ -172,54 +174,80 @@ def set_phases(cls, phases):
...
+                    f"the total number of trials seen so far when in phase ({cls._current_phase}). "
+                    f"Increasing max_trials might be a good idea. "
+                )
+
+            # Force the algo to observe the given number of trials.
+            cls.force_observe(num=n_previous_trials, algo=algo)
+            # TODO: Should we check that the algo has indeed observed the right number of trials that we
+            # want, and that the max_trials hasn't been busted, if present?
+            assert algo.n_observed == n_previous_trials
+
```
bouthilx, line 231: So far the algorithms are not cached to be reused across the tests for a given phase right?

src/orion/testing/algo.py
```diff
@@ -416,84 +400,83 @@ def get_id(point, ignore_fidelity=False, exp_id=None):
...
+        new_algo = self.create_algo(seed=seed)
+        assert new_algo.n_observed == algo.n_observed
+        trial_c = new_algo.suggest(1)[0]
+        assert trial_c == trial_a
 
-    @phase
-    def test_seed_rng_init(self, mocker, num, attr):
+    def test_seed_rng_init(self):
         """Test that the seeding gives reproducibile results."""
         algo = self.create_algo(seed=1)
```
bouthilx, line 419: I think this test will not work properly now since the new `create_algo` implementation is executing `algo.seed_rng`. Because of this we cannot test that the algo was seeded properly at instantiation.

src/orion/testing/algo.py
```diff
@@ -416,84 +400,83 @@ def get_id(point, ignore_fidelity=False, exp_id=None):
...
+        algo = self.create_algo(seed=seed)
 
-        seed = numpy.random.randint(10000)
-        algo.seed_rng(seed)
-        spy = self.spy_phase(mocker, max(num, 1), algo, attr)
         state = algo.state_dict
         a = algo.suggest(1)[0]
 
+        # NOTE: This is not necessarily true for all algorithms. For instance, if the algo doesn't
+        # have any RNG (e.g. GridSearch), this will fail.
```
bouthilx, line 448: But the algo should be reproducible anyhow, so the first suggested trial should be the same. Along the same lines, we assume the initial state to be the same, and to the be same for both after sampling the first trial.

tests/unittests/algo/test_evolution_es.py
bouthilx: (Also this should probably be another issue/PR, not work for this one)

bouthilx:
Thanks a lot for this PR! This is a great improvement on the parametrization of the phases for the algorithm test-suite. My only main concern is with the automatic parametrization of all tests which can lead to lots of skips (configuration test, is_done test, cardinality test, convergence test). If we could why a simple way to mark methods so that they are parametrized for first phase only that would be great.

src/orion/algo/hyperband.py
lebrice: Yes!

bouthilx: Ok great, good catch! :)

src/orion/algo/pbt/pbt.py
lebrice: Yes, the if/else fixes it.
Can I leave this note ?

tests/unittests/algo/test_evolution_es.py
lebrice: Created a new issue for this: https://github.com/Epistimio/orion/issues/890

I did not find out why, but I know that the trial's parents has a different ID as well (IDK if that's useful).

tests/unittests/algo/test_evolution_es.py
lebrice: I agree: What I was suggesting was to move this testing logic after the `yield` statement inside a fixture, in other words, in the "cleanup" phase, not that this is cleanup code.

src/orion/testing/algo.py
lebrice: Hmm. Not sure I agree.
I think all tests of a BaseMethodTests subclass should be run in all phases by default.
I see marking a test as `first_phase_only` a bit like adding an `xfail` on all the other phases. It's not something you should do often.

src/orion/testing/algo.py
lebrice: Yeah, they aren't. There would be a problem with the ordering of the parameterization from the fixture. I think I left a note somewhere about this.
I'll remove this TODO, it's outdated anyhow.

src/orion/testing/algo.py
lebrice: `create_algo` only calls `seed_rng` if a seed is passed as a parameter.

lebrice: Does that address your concern?

src/orion/testing/algo.py
lebrice: This doesn't just test the initial state anymore, it tests at each phase. This condition is correct, right?
I'll rephrase the note a bit.

src/orion/algo/pbt/pbt.py
lebrice: (since I don't think this is particularly clean, it's more like a patch for a bug)

src/orion/testing/algo.py
lebrice: Rephrased the note in ecba557eb89cddf71eeb540e0d4af43f94e1841a

src/orion/testing/algo.py
lebrice: Removed the todo in e456637ec38ea396be7f1f3713c52dbe7e65d63d

lebrice:
> If we could why a simple way to mark methods so that they are parametrized for first phase only that would be great.
 
I wouldn't know how to do this currently, since the `phase` fixture is already parameterized by the `params` when it actually gets access to the `request` fixture (where it could inspect `fixturenames` for instance).

I think it's fine to expect all tests to be parametrized to run on each phase by default (even configuration tests, for instance).

Also, The docs job is failing for a weird reason, some kind of internal pytests thing. I'll try to sort that out.

src/orion/algo/pbt/pbt.py
bouthilx: Leaving the note is fine, but it should be more clear that this is a hacky bug fix, rather than a bug to be fixed.

tests/unittests/algo/test_evolution_es.py
bouthilx: Thanks!

src/orion/testing/algo.py
bouthilx: All tests that are for argument values, configuration, convergence and is_done are not meant to be parametrized with phases. Depending on the algorithm that can be from about 1/5 to 1/3 of the tests.

src/orion/testing/algo.py
bouthilx: See in this case it will clearly call seed_rng, wheras this test is for seeding durint init, and thus seed_rng should not be called. It's problematic.

src/orion/testing/algo.py
bouthilx: Oh, I did not realize the changes to the test due to the modification of create_algo. So previously we would have `algo` updated up to the phase, but the new_algo would be at initial phase (num=0). I think it would be best to keep it this way, to make sure algorithms can have their state updated from different phases instead of always just one sample away. For this `create_algo` would need to take as argument whether it should be from scratch or at current phase.

bouthilx: Just to make sure we are not reinserting a blind spot here, do you remember which test made you find the bug in hyperband when you were modifying the test-suite?

src/orion/testing/algo.py
lebrice: I don't think that the "call `seed_rng` internally if seed is passed to constructor" behaviour is something we want to enforce, since algos don't have to take a seed as a constructor parameter.

src/orion/testing/algo.py
lebrice: > I think it would be best to keep it this way

You mean the way I have it here? I'm confused.

lebrice: > For this create_algo would need to take as argument whether it should be from scratch or at current phase.

Why do you need this, if the test is parametrized to run on all phases? Would you mind explaining again?

lebrice: > Just to make sure we are not reinserting a blind spot here, do you remember which test made you find the bug in hyperband when you were modifying the test-suite?

Which bug? Sorry I'm a bit slow this morning I guess

src/orion/testing/algo.py
lebrice: I guess I could skip the test if the algo under test doesn't have a `seed` constructor parameter..

src/orion/testing/algo.py
```diff
@@ -143,19 +95,82 @@ class BaseAlgoTests:
...
+
+    @pytest.fixture(name="first_phase")
+    @classmethod
+    def first_phase(cls, phase: TestPhase):
+        if phase != cls.phases[0]:
+            pytest.skip(reason="Test runs only on first phase.")
+
+    @classmethod
+    def duration_of(cls, phase: TestPhase) -> int:
+        """ Returns the number of trials in the given phase. """
```
bouthilx, line 162:
```suggestion
        """Returns the number of trials in the given phase. """
```

src/orion/testing/algo.py
bouthilx: For convergence and is_done it should only be last phase IMO. Arguments and configuration should be first phase only.

Skipping tests is fine in general but skipping systematically 1/3 of the tests is a lot. It's not a big issue, but it will pollute the pytest report. If you can find an easy fix that would be great but otherwise we'll just leave it like that.

lebrice: Added a first_phase_only and last_phase_only fixture (and make the mentioned tests use it) in 1ce04a9a8b14b2193000c9222880632cacf4e7b2

src/orion/testing/algo.py
lebrice: Fixed in 1ce04a9a8b14b2193000c9222880632cacf4e7b2

src/orion/testing/algo.py
lebrice: Fixed in c88d490e382d116532dd4735c7723d3479ec7aea

I still don't quite recall what bug you're referring to. :(

bouthilx:
Is this PR waiting for a final review from my side? Sorry I lost track in the past 2 weeks.

lebrice:
> Is this PR waiting for a final review from my side? Sorry I lost track in the past 2 weeks.

Yep. I'll fix the merge conflict tomorrow, but apart from that, it's ready to go IMO

src/orion/algo/pbt/pbt.py
bouthilx: Still unaddressed. It's noting critical, but it's confusing.

src/orion/testing/algo.py
bouthilx: So far yes they need to: https://github.com/Epistimio/orion/blob/develop/src/orion/algo/gridsearch.py#L116. GridSearch is the only situation where we don't use the seed so far.

src/orion/testing/algo.py
bouthilx: It's ok, I don't think you told me exactly what it was back then. Let's hope it is fixed properly. I'm it is. :)

bouthilx (approved):
LGTM, thanks!

PR #889, opened 2022-04-19
Improve prior build error message

bouthilx:
If there is an error during the attempt of building a builtin prior
(methods implemented in SpaceBuilder), the builder assumes the
expression is not a builtin prior and then tries to create a scipy.stat
prior which also fails. The error message is then that the name of the
expression is not a valid prior. We should instead explicitly verify if
the name of the prior is one that is builtin and then try to build it
without silencing error messages. This way we get the correct error
message for the prior.

### DIFF
--- a/tests/unittests/core/io/test_space_builder.py
+++ b/tests/unittests/core/io/test_space_builder.py
@@ -127,6 +127,17 @@ class TestDimensionBuilder(object):
         assert "Parameter" in str(exc.value)
         assert "size" in str(exc.value.__cause__)
 
+    def test_build_fails_because_of_invalid_arguments_for_custom_prior(
+        self, dimbuilder
+    ):
+        """Build fails because ValueError happens on init of builtin prior."""
+        with pytest.raises(ValueError) as exc:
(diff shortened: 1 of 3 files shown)


lebrice (approved):

tests/unittests/core/io/test_space_builder.py
```diff
@@ -127,6 +127,17 @@ def test_build_fails_because_of_ValueError_on_init(self, dimbuilder):
         assert "Parameter" in str(exc.value)
         assert "size" in str(exc.value.__cause__)
 
+    def test_build_fails_because_of_invalid_arguments_for_custom_prior(
+        self, dimbuilder
+    ):
+        """Build fails because ValueError happens on init of builtin prior."""
+        with pytest.raises(ValueError) as exc:
```
lebrice, line 134: You could use the `match` argument of `raises` instead of checking if the value is in the exception message, I think.
(That might be only true for `pytest.warns` though).

bouthilx: Indeed. This test is looking at two different parts of `exc.value` however.

Issue #892, opened 2022-04-21
Suggestion: add pre-commit config for black / isort / typos / linting


lebrice:
This would make the round-trip time for PRs shorter, by  having these fixed locally before being committed and pushed, instead of having to go back to the branch, fix the typo or isort issue, and then go back to the PR, then see a docs issue, then go back to the PR, etc.

This is essentially the same as adding these kinds of  'run-isort' and 'run-black' etc automatically after each commit, but only for the files that are changed by the commit. It's super quick to run.

For an example of such a config, check out the Gym repo: 
- GitHub action: https://github.com/openai/gym/blob/master/.github/workflows/pre-commit.yml
- pre-commit Config:  https://github.com/openai/gym/blob/master/.pre-commit-config.yaml

PR #895, opened 2022-04-22
Add .pre-commit-config.yaml for local setup

lebrice:
Devs can optionally turn on pre-commit locally using:
```console
$ pip install pre-commit
$ pre-commit install
```

Then, the following actions are performed on the diff before it is committed locally:
- black: Reformat the code using black
- codespell: Check for spelling mistakes
- flake8: check for various code quality issues
- isort: sort imports
- pyupgrade: Remove outdated python syntax stuff from the code (e.g. super(thing, self), etc)
- autoflake: Remove unused imports

If any of the above hooks fail, the commit is interrupted, the differences get added as modifications, and the user can add them before the commit goes through.

It's really simple and convenient.

The only thing that I'm not 100% sure about is codespell, which I think might generate a bunch of false positives. Even then, it's always possible to just add words to the ignored words list, as needed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,10 +29,15 @@ repos:
     rev: v2.31.0
     hooks:
       - id: pyupgrade
-        # TODO: remove `--keep-runtime-typing` option
-        args: ["--py37-plus", "--keep-runtime-typing"]
+        args: ["--py37-plus"]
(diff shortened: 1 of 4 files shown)


lebrice:
Fixes #892

bouthilx (approved):

PR #903, opened 2022-05-03
Fix PBT issues with working dir and promotion of max fidelity trials

bouthilx:
@Delaunay @lebrice This includes a fugly fix to working dir paths of trials that depends on the trial id which is different in transformed space and original space. To makes things worse, the ID of the trial depends on the experiment ID which is only set outsite of the algorithm by the experiment object after the trial was suggested. Because of this we need to rename directories or create sym-links at the different levels (In algo, in algo-wrapper, in experiment) to make sure that the working dir of the trial is corresponding to the one communicated to the user through `trial.working_dir` or env var `ORION_WORKING_DIR`.

We could get rid of 1 level by making the trial id not depend on experiment. In the database the index of trials should depend on trial.id and trial.experiment however to make sure trials are not duplicated within a given experiment. That would require updates of database from the side of users after next release however.

I don't see a simple and clean fix however for the different working_dir inside the algorithm and inside the algo-wrapper. The parameters are different and thus the ID and working_dir is different.

Would you have any ideas?


src/orion/core/worker/primary_algo.py
```diff
@@ -176,6 +177,14 @@ def suggest(self, num: int) -> list[Trial]:
                     # We haven't seen this trial before. Register it.
                     self.registry.register(original)
                     trials.append(original)
+                    # If the suggested trial has an exp_working_dir it means it was forked.
+                    if transformed_trial.exp_working_dir and os.path.exists(
+                        transformed_trial.working_dir
+                    ):
+                        print("to rename", transformed_trial.working_dir)
```
lebrice, line 184: prints

tests/unittests/algo/pbt/test_pbt.py
```diff
@@ -556,6 +565,58 @@ def test_suggest_num_population_size_fork_completed(self, space: Space, mocker):
         pbt_sample_mock.assert_called_with(2)
         pbt_fork_mock.assert_called_with(2)
 
+    def test_suggest_copy_working_dir(self, space, mocker, tmp_path):
```
lebrice, line 568: The Algo only has access to the transformed space, and so it doesn't have a way of knowing which of the trials is suggests are truly new. 

I believe there should be a test that makes sure that the algo (wrapped or not) doesn't create or copy any directories when it suggests a new trial.

lebrice (commented):
To summarize the discussion we had on Slack:

The PBT algorithm shouldn deal with copying / creating directories for trials, since it doesn't have any way of knowing which trials are actually new.

This responsibility should be moved to the Experiment, which would use the lineage information and the experiment id to create the working directory of the trial before it is passed to the User, and copy the files as necessary.

Delaunay (approved):

bouthilx:
Actually it should be the Runner's responsibility considering coming changes with https://github.com/Epistimio/orion/discussions/911. The trials will be generated on a remote server and thus the directory creation/copy should not occur during the trial generation on the server but rather in the Runner which is assumed to share the same file system as workers.

src/orion/client/experiment.py
```diff
@@ -668,6 +668,7 @@ def workon(
         max_broken=None,
         trial_arg=None,
         on_error=None,
+        prepare_trial=None,
```
lebrice, line 671: I'm curious: In what kind of scenario would the user need/want to pass a different callable than the default that we already use?

Also:
- if we want to make it possible to pass a different function, its expected signature should be described here (preferably using a type hint rather than a docstring).
- Wouldn't it be more accurate to set the default value here to `prepare_trial_working_dir`, rather than `None`?

  ```suggestion
          prepare_trial: Callable[[ExperimentClient, Trial], None]=prepare_trial_working_dir,
  ```

src/orion/client/runner.py
```diff
@@ -8,6 +8,8 @@
 Executes the optimization process
 """
 import logging
```
lebrice, line 10: About type hints: You might want to add these at the start:
```python
from __future__ import annotations
import typing
if typing.TYPE_CHECKING:
    from orion.client.experiment import ExperimentClient
```

src/orion/client/runner.py
```diff
@@ -151,7 +195,8 @@ def __init__(
         max_trials_per_worker,
         max_broken,
         trial_arg,
-        on_error,
+        on_error=None,
+        prepare_trial=None,
```
lebrice, line 199: Same comment about the `prepare_trial` argument here.

src/orion/core/worker/primary_algo.py
```diff
@@ -152,6 +152,19 @@ def suggest(self, num: int) -> list[Trial]:
                         f"Space: {self.transformed_space}"
                     )
                 original = self.transformed_space.reverse(transformed_trial)
+                if transformed_trial.parent is not None:
```
lebrice, line 155: Assuming that `trial.parent: Optional[str]`:
```suggestion
                if transformed_trial.parent:
```

src/orion/core/worker/primary_algo.py
```diff
@@ -348,9 +361,33 @@ def _verify_trial(self, trial: Trial, space: Optional[Space] = None) -> None:
             )
 
 
+def get_original_parent(registry, transformed_space, trial_parent_id):
```
lebrice, line 364: Type hints are sort-of required here given the context.
Also,  IMO any new function that gets added to a file without type hints should also have valid type hints.

```suggestion
def get_original_parent(registry: Registry, transformed_space: TransformedSpace, trial_parent_id: str) -> Trial:
```

src/orion/core/worker/trial.py
```diff
@@ -527,7 +528,7 @@ def compute_trial_hash(
             ignore_experiment = True
 
         experiment_repr = ""
-        if not ignore_experiment:
+        if not ignore_experiment and trial.experiment is not None:
```
lebrice, line 531: Assuming that `trial.experiment : Optional[str]`:

```suggestion
        if not ignore_experiment and trial.experiment:
```

tests/functional/algos/test_algos.py
```diff
@@ -122,6 +186,13 @@ def call(self, x: float | numpy.ndarray, noise: float | None = None) -> list[dic
 multidim_rosenbrock = MultiDimRosenbrock(max_trials=30, with_fidelity=False)
 
 
+def branching_rosenbrock(x, noise=None, trial=None):
```
lebrice, line 189:
```suggestion
def branching_rosenbrock(x: float, noise: float | None=None, trial: Trial | None=None) -> list[dict]:
```

tests/functional/algos/test_algos.py
```diff
@@ -396,3 +467,41 @@ def test_parallel_workers(algorithm):
...
         assert param.name == "x"
         assert param.type == "real"
+
+
+@pytest.mark.parametrize(
+    "algorithm",
+    branching_algorithm_configs.values(),
+    ids=list(branching_algorithm_configs.keys()),
+)
+def test_branching_algos(algorithm, tmp_path):
```
lebrice, line 477:
```suggestion
def test_branching_algos(algorithm: dict[str, Any], tmp_path: Path):
```

tests/functional/algos/test_algos.py
```diff
@@ -396,3 +467,41 @@ def test_parallel_workers(algorithm):
...
+    shutil.rmtree(tmp_path)
+    os.makedirs(tmp_path)
+
+    with OrionState() as cfg:  # Using PickledDB
```
lebrice, line 481: `cfg` seems unused here. 
```suggestion
    with OrionState():  # Using PickledDB
```

Also, would it be possible to use a fixture, to avoid indenting the entire body of this test?

tests/unittests/algo/pbt/test_pbt.py
```diff
@@ -1,30 +1,38 @@
...
     def test_base_1(self):
         assert compute_fidelities(10, 10, 20, 1) == list(map(float, range(10, 21)))
 
     def test_other_bases(self):
         assert compute_fidelities(9, 2, 2**10, 2) == [2**i for i in range(1, 11)]
 
-
-def _create_algo(space: Space, **pbt_kwargs) -> SpaceTransformAlgoWrapper[PBT]:
-    return create_algo(PBT, space=space, **pbt_kwargs)
+    def test_fidelity_upgrades(self, space):
```
lebrice, line 31:
```suggestion
    def test_fidelity_upgrades(self, space: Space):
```

tests/unittests/algo/pbt/test_pbt.py
```diff
@@ -549,6 +558,58 @@ def test_suggest_num_population_size_fork_completed(self, space: Space, mocker):
         pbt_sample_mock.assert_called_with(2)
         pbt_fork_mock.assert_called_with(2)
 
+    def test_suggest_copy_working_dir(self, space, mocker, tmp_path):
```
lebrice, line 561:
```suggestion
    def test_suggest_copy_working_dir(self, space: Space, mocker, tmp_path: Path):
```
It would also be nice to type the `mocker` if possible. (The type is `MockerFixture`, from `pytest_mock.plugin`)

tests/unittests/algo/pbt/test_pbt.py
```diff
@@ -549,6 +558,58 @@ def test_suggest_num_population_size_fork_completed(self, space: Space, mocker):
...
+        shutil.rmtree(tmp_path)
```
lebrice, line 562: I'm curious: Why do these tests have this `shutil.rmtree(tmp_path)` and `os.makedirs(tmp_path)`?
Shouldn't this tmp_path fixture give you an empty directory everytime?

tests/unittests/client/test_runner.py
```diff
@@ -94,6 +94,11 @@ def _free_executor(self):
             self.executor = None
             self.executor_owner = False
 
+    def get_trial(self, uid):
```
lebrice, line 97:
```suggestion
    def get_trial(self, uid: str) -> Trial:
```

tests/unittests/client/test_runner.py
```diff
@@ -112,7 +117,14 @@ def function(lhs, sleep):
     return lhs + sleep
 
 
-def new_runner(idle_timeout, n_workers=2, client=None, executor=None, backend="joblib"):
+def new_runner(
```
lebrice, line 120: Nit: Type hints.

Also, consider using a `Literal` annotation, e.g. `Literal["joblib", "dask"]` for the `backend` argument, since it's way cooler than just using `str` ;)

tests/unittests/client/test_runner.py
```diff
@@ -663,3 +676,85 @@ def test_runner_inside_dask():
     future = executor.submit(run_runner, executor=executor, reraise=True)
 
     assert future.get() == 0
+
+
+def test_custom_prepare_trial():
+    """Test that a different callback can be passed for prepare_trial"""
+
+    def test_callback(experiment_client, trial):
```
lebrice, line 684: Nit:
```suggestion
    def test_callback(experiment_client: ExperimentClient, trial: Trial) -> None:
```

tests/unittests/client/test_runner.py
```diff
@@ -663,3 +676,85 @@ def test_runner_inside_dask():
...
+
+    client = runner.client
+    client.trials.extend([new_trial(i, sleep=0.75) for i in range(count, -1, -1)])
+
+    runner.run()
+
+    assert test_callback.called == count
+
+
+def test_prepare_trial_working_dir(tmp_path):
```
lebrice, line 706:
```suggestion
def test_prepare_trial_working_dir(tmp_path: Path):
```

tests/unittests/core/test_primary_algo.py
```diff
@@ -356,6 +381,61 @@ def test_suggest_equivalent_to_existing(
         assert algo_wrapper.has_suggested(equivalent_original)
         assert algo_wrapper.algorithm.has_suggested(equivalent_transformed)
 
+    def test_suggest_non_existing_parents(self):
```
lebrice, line 384: Nit:
```suggestion
    def test_suggest_nonexistent_parents(self):
```

lebrice (commented):
Left some minor comments.

src/orion/client/experiment.py
bouthilx: > I'm curious: In what kind of scenario would the user need/want to pass a different callable than the default that we already use?

I don't know honestly. If they need to prepare something before the worker starts doing the work...

> if we want to make it possible to pass a different function, its expected signature should be described here (preferably using a type hint rather than a docstring).

Indeed

> Wouldn't it be more accurate to set the default value here to prepare_trial_working_dir, rather than None?

It would be more explicit. I was trying to avoid adding a reference to runner's default inside workon, but the advantage of making it more explicit is better in retrospect.

bouthilx: If I want to type the other arguments, what would be the correct Callable for `fct` since it takes `**kwargs`?

bouthilx: Fixed in c357931cec3b672251173861c937ef6f117f28a9

src/orion/client/runner.py
bouthilx: Fixed in 5c150f31

src/orion/core/worker/trial.py
bouthilx: Nop, it can be `str` or `int` depending on the DB backend. Thus `trial.experiment == 0` is possible.

tests/functional/algos/test_algos.py
bouthilx: Oh yeah, all OrionState are unnecessary in this module. Thanks for pointing this out! Fixed in ca6a7cdd13ae1af5af33b8d4a6e1cb3331577882

tests/unittests/algo/pbt/test_pbt.py
bouthilx: They should not indeed. I removed this tests since it is not relevant anymore. I'll check for the instance in functional tests.

bouthilx: I confirm it was not necessary. The folder exists but it is guaranteed empty. I removed them in cd2ff722.

tests/functional/algos/test_algos.py
```diff
@@ -370,138 +374,134 @@ def test_with_multidim(algorithm):
...
+    param = best_trial._params[1]
+    assert param.name == "x"
+    assert param.type == "real"
 
 
 @pytest.mark.parametrize(
     "algorithm", algorithm_configs.values(), ids=list(algorithm_configs.keys())
 )
-def test_parallel_workers(algorithm):
+def test_parallel_workers(algorithm, storage):
```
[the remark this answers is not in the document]
bouthilx: No, pylint is executed on the source code only, not on tests.

tests/unittests/core/utils/test_module_import.py
```diff
@@ -0,0 +1,58 @@
+"""Example usage and tests for :mod:`orion.core.utils.module_import`."""
+import pytest
+
+from orion.core.utils.module_import import ImportOptional
+
+
+class TestImportOptional:
+    def test_package_name_default_extra(self):
```
lebrice, line 8: Nitpick, I guess, but I'd be in favour of enabling pylint (and other code quality metrics/hooks) in tests as well, since we'd like those to be just as readable (if not even more) than our source code.
For instance, here, having a docstring on each test might be a little bit extra, but in general, I think it's a good idea to have some explanation of the desired behaviour.

I'm of the opinion that clear, well-written tests are the best form of documentation, but others might not agree.

bouthilx: I am hesitant. We used to have linters enabled on `tests/` too, but the tests often required patterns that outraged the linters. The tests docstrings also often ended up being a uninformative rewording of the test name, so we ended up removing the constraint and only using docstrings when we considered the test name not explicit enough.

I guess your comment implies that some of these test names are not clear enough. :) Which ones do you think would deserve a clearer explanation (if not all)?

lebrice: Nah it's not a big deal. Keep them as-is.

src/orion/core/utils/random_state.py
```diff
@@ -77,3 +78,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
             # Add the base_seed property in this case.
             random_state = replace(random_state, base_seed=base_seed)
         return random_state
+
+
+@contextmanager
+def control_randomness(random_state: RandomState | None):
+    """Seeds the randomness inside the indented block of code using `self.random_state`.
+
+    NOTE: This only has an effect if `seed_rng` was called previously, i.e. if
```
lebrice, line 87:
```suggestion
```

src/orion/core/utils/random_state.py
```diff
@@ -77,3 +78,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
+    `self.random_state` is not None.
```
lebrice, line 88:
```suggestion
```

src/orion/core/utils/random_state.py
```diff
@@ -77,3 +78,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
+    """
+    if random_state is None:
```
lebrice, line 90: Could probably remove this case, and mark the argument as required instead of optional.
It doesn't really make sense to call `control_randomness()` and for it to not do anything.

tests/functional/algos/test_algos.py
```diff
@@ -50,18 +54,98 @@
...
+                ],
+                "of_type": "PipelineExploit",
+            },
+            "fork_timeout": 60,
+        }
+    },
 }
 
+
+def xfail_if_not_installed(value, import_optional):
```
lebrice, line 117:
```suggestion
def xfail_if_not_installed(value: dict, import_optional: ImportOptional):
```

lebrice (approved):

src/orion/algo/axoptimizer.py
```diff
@@ -203,9 +204,14 @@ def suggest(self, num):
                 if self.fidelity_index is not None:
                     # Convert 0-dim arrays into python numbers so their type can
                     # be validated by Ax
-                    parameters[self.fidelity_index] = float(
-                        self.space[self.fidelity_index].high
+                    assert isinstance(
```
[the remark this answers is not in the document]
bouthilx: For now all dimensions are always wrapped in a transformedspace/transformeddimension even if the algo does not need any transformation. When I fixed the Space converter for ConfigSpace this hack here won't be necessary and I'll remove. We have the same hack in many algos that well need to be removed.

src/orion/algo/pbt/pb2.py
```diff
@@ -98,6 +106,29 @@ def configuration(self):
...
+    def state_dict(self) -> dict:
+        """Return a state dict that can be used to reset the state of the algorithm."""
+        state_dict: dict[str, Any] = super().state_dict
+        state_dict["random_state"] = self.random_state or RandomState.current()
+        return state_dict
+
+    def set_state(self, state_dict: dict) -> None:
+        """Reset the state of the algorithm based on the given state_dict"""
+        super().set_state(state_dict)
+        self.random_state = state_dict["random_state"]
```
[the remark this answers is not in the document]
bouthilx: Yes, the dict returned by the algo is pickled to a binary stream before being saved in the DB with the algo lock.

src/orion/core/utils/random_state.py
bouthilx: If the algo has no seed. We could have the if-else out-side control-randomness though.

src/orion/core/utils/random_state.py
```diff
@@ -77,3 +78,22 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
+        return
+
+    # Save the initial random state.
+    initial_rng_state = RandomState.current()
+    # Set the random state.
+    random_state.set()
+    yield
+    # Update the random state stored on `self`, so that the changes inside the block are
+    # reflected in the RandomState object.
+    random_state = RandomState.current()
```
lebrice, line 97: This line isn't used, now that I look at it.

src/orion/core/utils/random_state.py
```diff
@@ -77,3 +78,22 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
+def control_randomness(random_state: RandomState | None):
+    """Seeds the randomness inside the indented block of code using `self.random_state`."""
+    if random_state is None:
+        yield
+        return
+
+    # Save the initial random state.
+    initial_rng_state = RandomState.current()
+    # Set the random state.
+    random_state.set()
```
lebrice, line 93: Could move the if statement from above down here:
```suggestion
    if random_state:
        random_state.set()
```

src/orion/core/utils/random_state.py
bouthilx: Oh, you're right! Does it really make sense to have the RandomState be an immutable object? If we want to track the RNG states that's by definition tracking mutable objects.

bouthilx: Fixed in d3e4935e

src/orion/core/utils/random_state.py
bouthilx: Would also need to add it after the yield. I think it's simpler to have it as a first check.

src/orion/core/utils/random_state.py
bouthilx: Related discussion: https://github.com/Epistimio/orion/pull/903#discussion_r931059989

tests/unittests/core/worker/test_consumer.py
```diff
@@ -59,7 +59,7 @@ def mock_popen(self, *args, **kwargs):
     shutil.rmtree(trial.working_dir)
 
 
-def setup_code_change_mock(config, monkeypatch, ignore_code_changes):
+def setup_code_change_mock(storage, config, monkeypatch, ignore_code_changes):
```
lebrice, line 62:
```suggestion
def setup_code_change_mock(storage: dict, config: dict, monkeypatch, ignore_code_changes: bool):
```

src/orion/core/utils/random_state.py
```diff
@@ -80,20 +80,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
         return random_state
 
 
+class Algo(Protocol):
```
lebrice, line 83: Perhaps a more explicit name? Since this might eventually be used on things other than just algorithms?
```suggestion
class HasRandomState(Protocol):
```

lebrice (approved):

tests/unittests/core/worker/test_consumer.py
bouthilx: Storage is not a dict here. If you don't mind I'd skip on this one to wrap up this PR.

src/orion/core/utils/random_state.py
bouthilx: Agreed, done in a39c818.

src/orion/core/utils/random_state.py
```diff
@@ -80,26 +81,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
 
     random_state: RandomState | None
 
 
 @contextmanager
-def control_randomness(algo: Algo):
+def control_randomness(has_random_state: HasRandomState):
     """Seeds the randomness inside the indented block of code using `self.random_state`."""
-    if algo.random_state is None:
+    if has_random_state.random_state is None:
```
lebrice, line 93: One last thing. Wouldn't we want to set the property, if the object doesn't already have one? (for example, by just doing the same thing as below, but without the initial set()?
Or do we want to keep that object completely unseeded, if its `random_state` attribute is None?

src/orion/core/utils/random_state.py
```diff
@@ -80,26 +81,26 @@ def seed(cls, base_seed: int | None) -> RandomState:
...
         yield
         return
 
     # Save the initial random state.
     initial_rng_state = RandomState.current()
     # Set the random state.
-    algo.random_state.set()
+    has_random_state.random_state.set()
     yield
     # Update the random state stored on `self`, so that the changes inside the block are
```
lebrice, line 102:
```suggestion
    # Update the stored random state, so that the changes inside the block are
```

src/orion/algo/axoptimizer.py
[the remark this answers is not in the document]
lebrice: This assert here doesn't seem to have any value, typing-wise. (The while loop narrows the type to TransformedDimension within itself, and there is also an assert for Fidelity below it).
Plus, the code might change at some point, which will make this assertion false.

Removing the assert seems to me like a good idea, since it's unnecessary, and might prevent an error later on.

src/orion/core/utils/random_state.py
bouthilx: I'd say keep the object completely unseeded.

src/orion/algo/axoptimizer.py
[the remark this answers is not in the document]
bouthilx: Oh right, the type is verified in the while loop anyway.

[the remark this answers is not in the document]
bouthilx: Fixed in 72301433ce5ff6d53d8b7b13b09b2f13b730520e

PR #913, opened 2022-05-13
Hotfix for nevergrad tests

lebrice:
- Missing `from __future__ import annotations` at the top of `integration_test.py`
- Remove the Generic base for the test class, which prevented tests from being  collected in python 3.7


bouthilx (approved):
Thanks!

PR #915, opened 2022-05-13
Add Hebo

lebrice:
Migration of the HEBO algorithm from a plugin repo on GitHub into the Orion repo.


tests/unittests/algo/hebo/test_hebo.py
```diff
@@ -0,0 +1,225 @@
...
+        evolutionary_strategy=evolutionary_strategy,
+    )
+    # Replace the value in the Config.
+    TestHEBO.config["parameters"] = new_params
+    yield
+    TestHEBO.config["parameters"] = backup
+
+
+@pytest.fixture(autouse=True)
+def xfail_if_unseeded_model_chosen(request: SubRequest):
```
bouthilx, line 120: Can this be simplified following your PR improving the parametrization of algo test suite phases?

bouthilx: (Could be a follow up PR)

Delaunay:
If you move the hebo tests to the `long/` folder you can add `hebo` to the `test-long` matrix CI job and it will run `hebo` in parallel to the unit tests. You might also increase the max-parallel job in the test-long CI so make sure all the test-long run in parallel

tests/unittests/algo/hebo/test_hebo.py
lebrice: Fixed in 2a9d245fe1c404ed60cf859fbc23cf09d03a8e42

bouthilx:
HEBO tests take a significant amount of time to run, so it should be moved in long folder and run in the test-long matrix as pointed out by @Delaunay .

lebrice:
Oops, I forgot to increase the max parallel jobs value (as suggested by @Delaunay ) I'll do that tomorrow.

bouthilx:
> @bouthilx agreed! I moved the tests to the "long" folder, which I then renamed to "extras". (I also adjusted the workflow names to reflect this).
I suggest we place all tests related to algos that are installed through extras-require in that folder, regardless of how quick they run.

I disagree. The number of algorithms will continue growing and this will increase significantly the number of jobs. I believe we should try to reduce the number of jobs, and the criterion to do so is whether the algorithm takes to much time to run sequentially or not.

src/orion/algo/hebo/random_state.py
```diff
@@ -0,0 +1,79 @@
...
+try:
+    import torch
+except ImportError as err:
+    TORCH_INSTALLED = False
+else:
+    TORCH_INSTALLED = True
+
+
+@dataclass(frozen=True)
+class RandomState:
```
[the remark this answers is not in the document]
lebrice: Done: #919

bouthilx (approved):

PR #918, opened 2022-05-18
Replace (part of) pretest of CI with Pre-commit equivalent

lebrice:
Addresses #918 

Also applies the formatting, spelling mistakes, and python upgrade hooks to all files.


lebrice:
Hmm this doesnt quite work: seems like it's only running the pre-commit workflow.at the moment.

bouthilx:
Is there a corrector is the pre-commit? :hushed: That's great!

It seems the pre-commit is missing the doc compilation. That's to long to be included in the pre-commit indeed but should still be part of the pre-tests.

bouthilx:
There also seems to be no equivalent for pylint in the pre-commit.

lebrice:
Ok @bouthilx  I left the `doc` and `pylint` steps as part of pre-test, and moved the rest to pre-commit. Pylint is a bit annoying to use as a pre-commit hook, since it requires having pylint installed locally. It also seems to take a really long time, from my limited experience.

.github/workflows/build.yml
```diff
@@ -13,11 +13,23 @@ on:
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install pre-commit
+      - run: pre-commit --version
+      - run: pre-commit install
+      - run: pre-commit run black --all-files
+      - run: pre-commit run isort --all-files
+      - run: pre-commit run doc8 --all-files
   pretest:
+    needs: pre-commit
```
bouthilx, line 28: Can we not wait for pre-commit? If we could set the execution of pre-commit in tox we could have `toxenv: [pylint, docs, pre-commit]` so that the main tests still depends on the same pretests.

bouthilx: Running them through tox would make it possible to include the pre-commit tests in the current pre-test matrix. Having the pre-commit tests available in the tox config would also make it possible to run these tests locally if necessary outside of the pre-commit hook.

lebrice: Note: The tests can be run locally with pre-commit run <hook_name>, and either `--all-files`, or `--files <files>`.

bouthilx:
What is blocking this PR at the moment?

lebrice:
As far as I recall, I need to basically fix a few tests that got broken by removing unused imports in all modules.

src/orion/benchmark/task/profet/model_utils.py
```diff
@@ -219,7 +219,7 @@ def load_data(
                 raise RuntimeError(
                     f"Download finished, but file {file} still doesn't exist!"
                 )
-        with open(file, "r") as f:
+        with open(file) as f:
```
bouthilx, line 222: Doesn't pylint complain about not specifying the encoding? :thinking:

bouthilx (approved):
This will be awesome to use, thanks a lot!

PR #925, opened 2022-05-26
[WIP] HPO remote service

Delaunay:
Base implementation to get some ideas


lebrice:
What do you think of FastAPI? https://fastapi.tiangolo.com/#example

I don't know these tools personally, but this looks intersting to me

Delaunay:
It does look good, I used falcon because that is what is currently used in the repo.
The falcon way looks a bit verbose indeed

bouthilx:
I don't know them very well either. We would need to make sure it is worth converting current code to FastAPI if we want to adopt it. Now would be the best time to take a look before the webapi becomes too large to be worth converting.

PR #933, opened 2022-05-31
Knowledge Base + Multi-Task Warm-starting

lebrice:


src/orion/core/worker/warm_start/experiment_config.py
```diff
@@ -0,0 +1,127 @@
...
+    # user_vcs : str, optional
+    #    User's version control system for this executable's code repository.
+    # user_version : str, optional
+    #    Current user's repository version.
+    # user_commit_hash : str, optional
+    #    Current `Experiment`'s commit hash for **Oríon**'s invocation.
+
+
+@dataclass(frozen=True)
+class ExperimentInfo(Config):
```
bouthilx, line 68: This could be the start of a nice refactoring for Experiment and ExperimentClient. ExperimentInfo could clearly be useful outside of warm-start. Not suggesting for this PR though, but could be described in an issue for future work.

src/orion/core/worker/algo_wrappers/algo_wrapper.py
```diff
@@ -0,0 +1,321 @@
...
+    @property
+    def configuration(self) -> dict:
+        """Return tunable elements of this algorithm in a dictionary form
+        appropriate for saving.
+
+        Subclasses should overwrite this method and add any of their state.
+        """
+        # TODO: Do we also save the algorithm wrapper's configuration here? Or only the algo's?
+        # For now, we'll just always return the algo's config.
+        return self.algorithm.configuration
```
bouthilx, line 243: If we can avoid needing to configure that wrapper that would be best.

src/orion/core/worker/algo_wrappers/algo_wrapper.py
```diff
@@ -0,0 +1,321 @@
...
+        # NOTE: This field is created automatically in the BaseAlgorithm class.
+        super().__init__(space)
+        self._algorithm = algorithm
+        assert not isinstance(algorithm, dict), algorithm
+        self.registry_mapping = RegistryMapping(
+            original_registry=self.registry,
+            transformed_registry=self.algorithm.registry,
+        )
+
+    def transform(self, trial: Trial) -> Trial:
```
bouthilx, line 40: Do you think trial transformations are general enough to be part of the base algo wrapper? I'm not sure, but I don't have a strong opinion.

src/orion/core/worker/primary_algo.py
```diff
@@ -1,356 +1,109 @@
...
-        Algorithm to be wrapped.
-    space : `orion.algo.space.Space`
-       The original definition of a problem's parameters space.
-    algorithm_config : dict
-       Configuration for the algorithm.
-
+    algo_type : Callable[..., AlgoType]
+        Type of algorithm to create and wrap.
+    space : Space
+        _description_
```
bouthilx, line 81: placeholder?

src/orion/core/worker/primary_algo.py
```diff
@@ -1,356 +1,109 @@
...
+
+    space = InsistSuggest.transform_space(space)
+    spaces.append(space)
+
+    space = SpaceTransform.transform_space(space, algo_type=algo_type)
+    spaces.append(space)
+
+    # Create the algo, using the innermost (most transformed) space.
+    # Then, create each wrapper, from the bottom up.
+    algorithm = algo_type(space=spaces.pop(), *algo_args, **algo_kwargs)
```
bouthilx, line 104: Do I understand correctly that the algorithm's suggested trial, will first go through SpaceTransform, then InsistSuggest and finally MultiTaskWrapper?

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+    @classmethod
+    def transform_space(cls, space: Space, knowledge_base: KnowledgeBase) -> Space:
+        """Transform the space, so that the algorithm that is passed to the constructor is already
+        transformed.
+        """
+        # TODO: Should we have this dimension be larger than the current number of
+        # experiments in the KB, in case some get added in the future?
+        # TODO: Should the number of tasks here only count the experiments where the spaces are
+        # compatible? (Currently only counts total number of experiments in the KB)
+        # TODO: Could we use a Catagorical over the experiment IDS instead of a Categorical with
```
bouthilx, line 66: It could help if we don't get consistent order of experiments from one execution to another.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+        # Dict mapping from task ID to list of compatible trials.
+        compatible_trials: dict[int, list[Trial]] = defaultdict(list)
+        task_ids_to_experiments: dict[int, ExperimentInfo] = {}
+
+        for i, (experiment_info, trials) in enumerate(warm_start_trials.items()):
+            # Start the task ids at 1, so the current experiment has task id 0.
+            task_id = i + 1
+            for trial in trials:
+                # Drop the point if it doesn't fit inside the current space.
+                # TODO: Do we want to 'translate' the point in this case?
```
bouthilx, line 119: Maybe this kind of feature should be implemented on the side of the KB so that the user can configure the KB instead of the algo wrapper for this. Or maybe something in-between the KB and the algo wrapper.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+        trials_from_target_task = sum(
+            get_task_id(trial) == 0 for trial in self.unwrapped.registry
+        )
+        trials_from_other_tasks = total_trials - trials_from_target_task
+        log.debug(
+            f"Trials from target task: {trials_from_target_task}, "
+            f"trials from other tasks: {trials_from_other_tasks} "
+        )
+        log.debug(f"self.n_observed: {self.n_observed}")
+        log.debug(f"self.n_observed: {self.n_suggested}")
```
bouthilx, line 199:
```suggestion
        log.debug(f"self.n_suggested: {self.n_suggested}")
```

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+        self.current_task_id = task_id
+        yield
+        self.current_task_id = previous_task_id
+
+    def _add_task_id(self, trial: Trial, task_id: int) -> Trial:
+        # NOTE: Need to do a bit of gymnastics here because the points are expected
+        # to be tuples. In order to add the task id at the right index, we convert
+        # them like so:
+        # Trial -> dict (.params) -> dict (add task_id) -> Trial.
+        params = trial.params.copy()
```
bouthilx, line 228: Maybe this could reuse/be merged in part with `_copy_status_and_results` from the base algo wrapper module?

src/orion/core/worker/warm_start/serializable.py
```diff
@@ -0,0 +1,148 @@
...
+    Mapping,
+)
+from collections.abc import ItemsView
+from logging import getLogger as get_logger
+from orion.core.utils.flatten import flatten, unflatten
+
+logger = get_logger(__name__)
+
+
+class SerializableMixin(Mapping[str, Any]):
```
bouthilx, line 18: Is this coming from another package your are working on? I'm wondering if it would be best to add dependencies instead of maintaining+testing it.

src/orion/core/worker/algo_wrappers/algo_wrapper.py
lebrice: I mean, we could have a `TransformWrapper` as an an intermediate base class, so that we'd pool the `SpaceAdapterWrapper --> TransformWrapper --> AlgoWrapper`, but that seems a bit unnecessary to me atm, since we only have a few wrappers.

src/orion/core/worker/primary_algo.py
lebrice: Indeed, that's the idea!

src/orion/core/worker/warm_start/knowledge_base.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def n_stored_experiments(self) -> int:
+        """Returns the current number of experiments registered the Knowledge base."""
+
+    @abstractmethod
+    def add_experiment(self, experiment: Experiment | ExperimentClient) -> None:
```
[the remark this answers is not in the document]
lebrice: Interesting! What do you foresee as being hard? Would you mind giving me an example, just so I can understand better?

Wouldn't it be harder to remember which experiments are in which storage objects?

E.g.
```python
knowledge_base = KnowledgeBase()
my_previous_experiment = (...)
knowledge_base.add_experiment(my_previous_experiment)
target_experiment = create_experiment("target", ..., knowledge_base=knowledge_base)
```

Seems more explicit to me than this:

```python
knowledge_base = KnowledgeBase()
knowledge_base.add_storage(PickleDb("my_previous_experiment_storage.pkl"))
target_experiment = create_experiment("target", ..., knowledge_base=knowledge_base)
```
Since it's not necessarily the case that the storage has only one experiment.

src/orion/core/worker/warm_start/knowledge_base.py
```diff
@@ -0,0 +1,79 @@
...
+
+        Parameters
+        ----------
+        experiment : Union[Experiment, ExperimentClient]
+            Experiment or experiment client to add to the KB.
+        """
+
+    # NOTE: Not making this an abstract method, since we might need to adapt this a bit.
+    # @abstractmethod
+    def add_storage(self, storage: BaseStorageProtocol) -> None:
```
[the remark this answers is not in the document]
lebrice: Wouldn't this "MetaStorage" be a subclass of "Storage"? (The same way that pytorch's `ConcatDataset` is a subclass of  `Dataset`?)

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def reverse_transform(self, trial: Trial) -> Trial:
+        return self._remove_task_id(trial)
+
+    @classmethod
+    def transform_space(cls, space: Space, knowledge_base: KnowledgeBase) -> Space:
+        """Transform the space, so that the algorithm that is passed to the constructor is already
+        transformed.
+        """
+        # TODO: Should we have this dimension be larger than the current number of
```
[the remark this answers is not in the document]
lebrice: I agree that everything should be 100% reproducible. Is this in response to the TODO here though? If so, I don't understand the link.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+                f"total, out of which {n_compatible_trials} were found to be "
+                f"compatible with the target experiment."
+            )
+
+        if not compatible_trials:
+            log.info("No compatible trials detected.")
+            return
+
+        # Only keep trials that are new.
+        new_compatible_trials = {
```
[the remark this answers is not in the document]
lebrice: Sure, technically it's multi-task learning at that point. But do we need to turn it off though? Seems like a low-hanging-fruit to me, but I'll remove it if you want.

src/orion/core/worker/primary_algo.py
lebrice: Fixed in 3ae6acb5b196cb4a3d373828ba740bf3728b0be4

src/orion/core/worker/warm_start/multi_task_wrapper.py
lebrice: Not sure I understand how though. I'm already using `_copy_status_and_results` below.
I made this a bit more explicit using the `trial_with_params` and `trial_with_info` arguments, and removed the outdated comments in 3ae6acb5b196cb4a3d373828ba740bf3728b0be4.

src/orion/core/worker/warm_start/serializable.py
lebrice: This comes from simple-parsing. I doubt that you'd want to add a dependency to it, unless that's necessary.
(Although, simple-parsing is a really lightweight package.)

lebrice: There are other (perhaps better) alternatives though (third-party libraries such as Pydantic etc). I'll take a look.

src/orion/core/worker/warm_start/experiment_config.py
bouthilx: I was thinking of more substantial changes in Experiment and ExperimentClient. Let's leave it here for now, but it would be good to move this when we are able to tackle a refactoring of Experiment and ExperimentClient.

src/orion/core/worker/algo_wrappers/algo_wrapper.py
bouthilx: So we agree we can remove the TODO?

src/orion/core/worker/algo_wrappers/algo_wrapper.py
bouthilx: Yeah that's fine. I didn't realize at that point in my review that the wrapper for warm-starting would make transformations as well.

src/orion/core/worker/warm_start/knowledge_base.py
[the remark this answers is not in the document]
bouthilx: It is more explicit indeed but then I can only reproduce the experiment using the same code (or re-implementation) to add experiments piece by piece, whereas it could be reproduced based on a configuration of storage configs. But even with the storage it can be problematic since they are mutable... There is something I don't like here for reproducibility :sweat_smile: Maybe the KB should always lead to a versionned dataset based on current status of the storages, and then we can reproduce by used the versionned dataset instead of the original storage configs.

src/orion/core/worker/warm_start/knowledge_base.py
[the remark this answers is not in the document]
bouthilx: We already have a readonlydb actually: https://github.com/Epistimio/orion/blob/develop/src/orion/core/io/database/__init__.py#L284. Yeah that would work. I leave it up to the typing master to figure out the best way to lay this out. :smile:

src/orion/core/worker/warm_start/multi_task_wrapper.py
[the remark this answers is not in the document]
bouthilx: This would make the dimension size (number of categories) fix.

src/orion/core/worker/warm_start/multi_task_wrapper.py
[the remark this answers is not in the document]
bouthilx: Not sure, it may be a low hanging fruit very strongly attached to a branch that won't come off so easily. :) Limiting to warm-start makes it much simpler for when we call the warm-starting, how to make the query efficient and easily reproducible.

src/orion/core/worker/warm_start/serializable.py
bouthilx: I don't know very well pydantic. If you think that is a good alternative than we could use it. 

Otherwise, do you think simple-parsing could help improve the cmdline parsers of Oríon? (thinking of commands like `hunt`, `status`, `db`, etc).

src/orion/algo/base.py
```diff
@@ -409,14 +404,29 @@ def configuration(self):
...
-    @space.setter
-    def space(self, space):
-        """Set space."""
-        self._space = space
+    @property
+    def unwrapped(self):
+        return self
+
+    @contextmanager
+    def warm_start_mode(self):
```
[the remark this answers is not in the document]
bouthilx: So trials are all squashed together in the algo. I am worried that this will cause issues with methods like `is_done`.

src/orion/core/worker/algo_wrappers/algo_wrapper.py
lebrice: Yeah, I made it a NOTE. Sorry I don't know which commit exactly.

src/orion/core/worker/warm_start/multi_task_wrapper.py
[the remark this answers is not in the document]
lebrice: Given the huge transformer stuff I've been reading, I'm voting towards keeping this in. We can chat about this more in our sync-up meeting if you'd like.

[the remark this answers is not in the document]
bouthilx: Ok! Looking forward to hear more about it!

src/orion/core/worker/warm_start/knowledge_base.py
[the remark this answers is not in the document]
lebrice: I think that `BaseStorageProtocol` (which should really be renamed to just `Storage` imo) is fine here.
If someone wants to add a read-only storage object, that's fine. If they want to add a writeable storage object, that's also fine. The KnowledgeBase itself won't be writeable, in any case.

src/orion/core/worker/warm_start/multi_task_wrapper.py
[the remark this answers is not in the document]
lebrice: > Perhaps we should make the KB specification so that running the experiment twice would query exactly the same experiments.

I still don't quite understand what you mean here.
The number of categories would be fixed, if and only if the user creates the same knowledge base, with the same experiments, correct? Sure, the ordering will be deterministic (e.g. sorted by name or similarity or something).

Also, about the idea of limiting the KB specification, do you mean that, after creating a KB with a storage object A, if new experiments are added to A, and then a new KB is created with Storage A, then it should have the same "view" of experiments/trials as the first KB? If so, I disagree.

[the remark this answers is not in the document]
bouthilx: If a new experiment is added, then the number of tasks will increase, thus there will be one additional category for the task_id dimension.

With respect to freezing the KB, from what I remember we agreed offline that this should not be implemented and users should rather rely on the upcoming feature to checkpoint a storage which they could assume would not change if they don't make any modifications to it. Maybe I misunderstood and we had something different in mind?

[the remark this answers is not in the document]
lebrice: No that's all good.

src/orion/core/worker/warm_start/experiment_config.py
lebrice: I moved this to `orion/core/worker/experiment_config.py` and also removed the Serializable class in 1b29a3813db57aa3f12e09cc49ca335fb5b15f4c

The `configuration` property is marked as returning this type of dictionary, although the implementation doesn't currently match (as far as I can tell). Therefore, I'm just leaving it as a type annotation, and adding a TODO/issue for later, to actually return the literal dictionary, with the right type of entries.

LMK what you think

src/orion/core/worker/warm_start/serializable.py
lebrice: Removed in 1b29a3813db57aa3f12e09cc49ca335fb5b15f4c

src/orion/algo/base.py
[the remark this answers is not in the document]
lebrice: There's now a test for this: https://github.com/lebrice/orion/blob/278bfcf4a45d942f240528f683435acb66c10b90/tests/unittests/core/worker/warm_start/test_multi_task.py#L334

src/orion/core/io/experiment_builder.py
```diff
@@ -630,21 +630,22 @@ def _instantiate_algo(
...
-        algo_name = backported_config.pop("of_type")
-        algo_type = algo_factory.get_class(algo_name)
-        algo_config = {}
-    else:
-        raise ValueError(f"Invalid algorithm configuration: {config}")
 
     try:
+        if isinstance(config, str):
+            algo_type = algo_factory.get_class(config)
+            algo_config = {}
```
bouthilx, line 637: Looks like the if-else clauses here are not tested.

src/orion/core/worker/experiment_config.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    datetime: datetime
+    """ When was this particular configuration submitted to the database. """
+
+
+class MetaData(BasicMetaData, total=False):
+    """Metadata of the experiment. Also contains some optional keys."""
+
+    # NOTE: Fields that don't appear to be set in the metadata, but that are mentioned in the
```
bouthilx, line 64: They are set in `resolve_config` when using the commandline interface.

src/orion/core/worker/experiment_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+    type: str
+
+
+class StorageConfig(TypedDict):
+    """Storage configuration."""
+
+    type: str
+    database: DatabaseConfig
+
```
bouthilx, line 26: We are these used?

src/orion/client/__init__.py
```diff
@@ -331,14 +342,15 @@ def workon(
 
     try:
         setup_storage(storage={"type": "legacy", "database": {"type": "EphemeralDB"}})
-
+        # TODO: A bit weird that we pass the kb to both the experiment and exp client...
```
bouthilx, line 345: It's only passed to the experiment here, no?

src/orion/core/worker/producer.py
```diff
@@ -55,6 +56,23 @@ def produce(self, pool_size, timeout=60, retry_interval=1):
         with self.experiment.acquire_algorithm_lock(
             timeout=timeout, retry_interval=retry_interval
         ) as algorithm:
+            if (
+                self.experiment.knowledge_base
+                and not self.warm_started
+                and is_warmstarteable(algorithm)
+            ):
+                # TODO: Ugly.
+                similar_trials = self.experiment.knowledge_base.get_related_trials(
```
bouthilx, line 65: Ugly because knowledge_base is an attribute of experiment and experiment is passed to kb's method?

src/orion/core/io/experiment_builder.py
lebrice: You mean you'd rather have those as a sequence of ifs, rather than elifs?
If so, even though you're correct that it would work with a sequence of ifs, since they don't affect each other's conditions, I like the clarity that elifs bring: It's 100% unambiguous that only one condition will be met.

bouthilx: Oh no! Elifs are better! I meant that the if-else clauses (including the elifs) are currently not covered by the unit-tests.

src/orion/core/worker/experiment_config.py
lebrice: Nowhere atm, since (afaik, but correct me if I'm wrong) the storage isn't part of the Experiment configuration dictionary.
However, if it is, or if it becomes part of it, we could then add a `storage: StorageConfig` annotation onto the `ExperimentConfig` typeddict.

src/orion/core/worker/experiment_config.py
lebrice: Okay, good to know. They are still optional, from a typing perspective.

src/orion/core/worker/experiment_config.py
bouthilx: I think we should add them when we need them.

lebrice: Removed in 8f48a0e6ef9bb451fee59dbdf7e0f07d1be2f66e

bouthilx: ok thanks!

src/orion/client/__init__.py
lebrice: Yeah outdated todo. Removed in b0f9de3a0605afaf21973f510ba160da8146cd00

src/orion/core/worker/producer.py
lebrice: Yeah sort-of!

src/orion/core/io/experiment_builder.py
lebrice: > Looks like the if-else clauses here are not tested.

I'm sorry, I misread that as "looks like the if-else clauses are not used"!

src/orion/core/worker/producer.py
bouthilx: Maybe the experiment object could pass it's ExperimentConfig instead? :thinking:

src/orion/core/io/experiment_builder.py
bouthilx: Haha, no problem!

lebrice: Ok there's something weird with the coverage I think. Last time I checked, it was saying that none of the cases were being tested.

src/orion/core/worker/producer.py
lebrice: Yeah indeed! Maybe that's nicer: 55f7652c3f0fd7535df59a71d9a6ee08539ffd1c

src/orion/core/io/experiment_builder.py
bouthilx: Yeah it should go through at least one of them. :thinking: I am pretty sure `str` and `dict` is covered. Not sure about BaseAlgorithm and the else-clause (no algo found).

src/orion/core/worker/producer.py
bouthilx: Yeah, it's narrower at least.

tests/functional/core/worker/warm_start/test_warm_starting.py
```diff
@@ -0,0 +1,225 @@
...
+        knowledge_base=knowledge_base,
+    )
+    assert len(with_warm_starting.fetch_trials()) == max_trials
+
+    best_trial_with = _get_best_trial(with_warm_starting)
+    best_trial_without = _get_best_trial(without_warm_starting)
+
+    objective_with = _get_objective(best_trial_with)
+    objective_without = _get_objective(best_trial_without)
+    # FIXME: This is still somehow not always true, weirdly enough. Perhaps the TPE algo has
```
bouthilx, line 105: Did you manage to solve this issue?

bouthilx (approved):
Looks good! I suggest we create an issue and add in a later PR  support for defining the KB using the configuration files (would need to be added in `orion.core.config`).

bouthilx:
Oh, and I suggest we wait for release v0.2.5 to be done before we merge this one in develop. So that the KB be part of release v0.2.6.

tests/functional/core/worker/warm_start/test_warm_starting.py
lebrice: Didn't change anything really, but strangely, I've been running the tests like 10 times this morning and it's passed every time.  :thinking:

src/orion/core/io/experiment_builder.py
```diff
@@ -855,11 +919,14 @@ def _default(v: T | None, default: V) -> T | V:
 
         space = _instantiate_space(space)
         max_trials = _default(max_trials, orion.core.config.experiment.max_trials)
+        if isinstance(knowledge_base, dict):
+            knowledge_base = _instantiate_knowledge_base(knowledge_base)
```
bouthilx, line 923: Is this part really not covered in tests? The KB is never defined as a dict?

src/orion/core/io/experiment_builder.py
```diff
@@ -200,34 +208,76 @@ def _instantiate_space(config):
     return SpaceBuilder().build(config)
 
 
-def _instantiate_algo(space, max_trials, config=None, ignore_unavailable=False):
+def _instantiate_knowledge_base(kb_config: dict[str, Any]) -> KnowledgeBase:
+    """Instantiate the Knowledge base from its configuration."""
+    if len(kb_config) != 1:
```
bouthilx, line 213: I think a test for these different ways of defining the DB config should be relatively simple to test.

src/orion/core/io/experiment_builder.py
```diff
@@ -200,34 +208,76 @@ def _instantiate_space(config):
...
     try:
-        backported_config = backward.port_algo_config(config)
-        algo_constructor: type[BaseAlgorithm] = algo_factory.get_class(
-            backported_config.pop("of_type")
-        )
-        # NOTE: the config doesn't have the `of_type` key anymore, it only has the algo's kwargs
+        if isinstance(config, str):
+            algo_type = algo_factory.get_class(config)
+            algo_config = {}
+        elif inspect.isclass(config) and issubclass(config, BaseAlgorithm):
```
bouthilx, line 262: Following up on https://github.com/Epistimio/orion/pull/933/files#r933572718, this elif and the last else clause are likely not covered by the tests. If you find tests covering these could you please share the links here. Thanks!

src/orion/core/worker/algo_wrappers/algo_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+        If it isn't, raises a RuntimeError.
+        """
+        node = self
+        while isinstance(node, AlgoWrapper):
+            if isinstance(node, target_type):
+                return node
+            node = node.algorithm
+        if isinstance(node, target_type):
+            return node
+        raise RuntimeError(
```
bouthilx, line 70: This also could be a simple test to add.

src/orion/core/worker/warm_start/knowledge_base.py
```diff
@@ -0,0 +1,168 @@
...
+    ) -> list[tuple[ExperimentConfig, list[Trial]]]:
+        """Takes at most `max_trials` trials from the given experiments.
+
+        If `max_trials` is None, returns all trials from all given experiments.
+        """
+        related_trials: list[tuple[ExperimentConfig, list[Trial]]] = []
+        total_trials_so_far = 0
+        for experiment_config in experiment_configs:
+            experiment_id = experiment_config["_id"]
+            if experiment_id is None:
```
bouthilx, line 108: What were these `if` supposed to process? Experiments should always have an ID if they are fetched from the storage.

src/orion/core/worker/warm_start/knowledge_base.py
```diff
@@ -0,0 +1,168 @@
...
+                continue
+
+            trials = self.storage.fetch_trials(uid=experiment_id)
+            if trials is None:
```
bouthilx, line 112: This should be tested.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,249 @@
...
+        return self._remove_task_id(trial)
+
+    # pylint: disable=arguments-differ
+    @classmethod
+    def transform_space(cls, space: Space, knowledge_base: KnowledgeBase) -> Space:
+        """Transform an (outer) space, returning the (inner) space of the wrapped algorithm.
+
+        This returns what will become `self.algorithm.space` in an instance of this class.
+        """
+        if "task_id" in space:
```
bouthilx, line 59: I think this should be tested.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,249 @@
...
+                    task_ids_to_experiments[task_id] = experiment_info
+
+            n_compatible_trials = len(compatible_trials[task_id])
+            logger.info(
+                f"Experiment {experiment_info} has {len(trials)} trials in "
+                f"total, out of which {n_compatible_trials} were found to be "
+                f"compatible with the target experiment."
+            )
+
+        if not compatible_trials:
```
bouthilx, line 134: This should be tested.

src/orion/core/worker/warm_start/multi_task_wrapper.py
```diff
@@ -0,0 +1,249 @@
...
+            logger.info("No compatible trials found!")
+            return
+
+        # Note: If there are no compatible trials found, this would allow multiple attempts at
+        # warm-starting. This isn't used though.
+        if self._total_warm_start_trials:
```
bouthilx, line 140: This should be tested.

src/orion/core/io/experiment_builder.py
lebrice: Should be fixed by 34965b402c81ab37aad27c83e034780bc2e0f162

src/orion/core/io/experiment_builder.py
lebrice: Should be fixed by 34965b402c81ab37aad27c83e034780bc2e0f162

src/orion/core/worker/warm_start/knowledge_base.py
lebrice: The type of the `_id` field of the Experiment class is currently `str | int | None`. This `if` clause is to take care of that possibility, but you're right, it will never be None in this case. However, type-correctness requires me to add this case here.

src/orion/core/io/experiment_builder.py
lebrice: Should be fixed in 5040c07f67c31da2d7bac6a74cfec256d2ae574a

src/orion/core/io/experiment_builder.py
lebrice: Should be fixed in 5040c07f67c31da2d7bac6a74cfec256d2ae574a

src/orion/core/worker/algo_wrappers/algo_wrapper.py
lebrice: Should be fixed in cb616d2f22c915a7ce5c265e281b11f325ef04cc

src/orion/core/worker/algo_wrappers/algo_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+
+        :param state_dict: Dictionary representing state of an algorithm
+
+        AlgoWrappers should overwrite this and restore any additional state they own.
+        """
+        super().set_state(state_dict)
+        self.algorithm.set_state(copy.deepcopy(state_dict["algorithm"]))
+
+    def __repr__(self) -> str:
+        algo_repr = repr(self.algorithm)
```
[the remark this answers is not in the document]
lebrice: I added this test for the InsistSuggest wrapper in cb616d2f22c915a7ce5c265e281b11f325ef04cc, and I don't quite see a way of making the test purely generic for all the algo wrappers.. LMK if this is sufficient.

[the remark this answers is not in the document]
lebrice: Ok, made the tests generic for all wrappers (except the MultiTaskWrapper) in bc029748c9d454f52e5afd8207520962ddf13f11, and then corrected / simplified things a bit in 57d561bd6eeede5eb0ba01071ff93e3ecad59765

src/orion/core/worker/algo_wrappers/insist_suggest.py
```diff
@@ -0,0 +1,60 @@
...
+        for suggest_attempt in range(1, self.max_suggest_attempts + 1):
+            trials = super().suggest(num)
+            if trials:
+                if suggest_attempt > 1:
+                    logger.debug(
+                        f"Succeeded in suggesting new trials after {suggest_attempt} attempts."
+                    )
+                return trials
+
+            if self.is_done:
```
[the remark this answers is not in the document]
lebrice: Should be fixed with cb616d2f22c915a7ce5c265e281b11f325ef04cc

src/orion/core/worker/producer.py
lebrice: Can we mark this as resolved?

src/orion/core/worker/warm_start/knowledge_base.py
lebrice: "Fixed" in 8f23392155a38400310c6e6922c7141de4b8652a

src/orion/core/worker/warm_start/knowledge_base.py
lebrice: "fixed" in 8f23392155a38400310c6e6922c7141de4b8652a

src/orion/core/worker/warm_start/multi_task_wrapper.py
lebrice: aff8c7b3d62ba5c46b1a245b80cf360497ef108b

src/orion/core/worker/warm_start/multi_task_wrapper.py
lebrice: Fixed in 3555408b49b0e27c7005543d318f0c411797b5a4

src/orion/core/worker/warm_start/multi_task_wrapper.py
lebrice: Fixed in 47d8a951b3953e4472cf5d0a7e3d4c64c1af7b72

src/orion/core/worker/warm_start/warm_starteable.py
```diff
@@ -0,0 +1,98 @@
...
+        Whether the input is a warm-starteable algorithm.
+    """
+    if isinstance(algo, AlgoWrapper):
+        # NOTE: Not going directly to algo.unwrapped, because the MultiTask wrapper makes the algo
+        # WarmStarteable. Check down the chain of wrappers instead.
+        return isinstance(algo, WarmStarteable) or is_warmstarteable(algo.algorithm)
+    if isinstance(algo, BaseAlgorithm):
+        return isinstance(algo, WarmStarteable)
+    if isinstance(algo, str):
+        # NOTE: I don't think the algorithm should ever be a string, since now we pass the algo
```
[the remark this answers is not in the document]
lebrice: Fixed in 0d25b26c2d07ae624995abb8a19871878cf85178

bouthilx:
The failing test is not related to this PR. Looks like `develop` needs a fx.

src/orion/core/io/experiment_builder.py
bouthilx: Great thanks! There's only the invalid config case missing.

src/orion/core/io/experiment_builder.py
bouthilx: Looks like there are multiple possible ways to incorrectly define the KB (3 different raises) but only 1 is covered.

src/orion/core/io/experiment_builder.py
lebrice: Actually, now that I think about it, this last "invalid config" case isn't serving any real purpose, since every declared type for `config` is already covered. Since this is an 'internal' function, I'd argue that this case can be removed. This is also a good use-case for asserts, and this saves us from having to write a useless test.

src/orion/core/io/experiment_builder.py
lebrice: Fixed in 3f35bbaf092beadd874bf86cf3d212af8d48a79d

src/orion/core/io/experiment_builder.py
lebrice: 5a61b5988c4ca8bcc66ee0586a0307511999b57c

lebrice: I read the message of that comment with an italian accent, for some reason :)

bouthilx (approved):
That was quite some work! Good job! :)

PR #942, opened 2022-06-06
no singleton test

Delaunay:


src/orion/benchmark/benchmark_client.py
```diff
@@ -11,13 +11,19 @@
...
 
 def get_or_create_benchmark(
-    name, algorithms=None, targets=None, storage=None, executor=None, debug=False
+    name,
+    algorithms=None,
+    targets=None,
+    storage=None,
+    executor=None,
+    debug=False,
+    storage_instance=None,
```
bouthilx, line 26: Documentation missing for storage_instance. Also the user must now pass at least one of them, or is setup_storage using global config otherwise?

src/orion/core/cli/serve.py
```diff
@@ -39,7 +40,8 @@ def main(args):
     """Starts an application server to serve http requests"""
     config = experiment_builder.get_cmd_config(args)
 
-    web_api = WebApi(config)
+    storage = setup_storage()
```
bouthilx, line 43: In WebApi the storage was created based on `config['config']`. It should be created the same way here before being passed to setup_storage() otherwise we cannot configure what storage to use with a local config file.

src/orion/core/io/database/pickleddb.py
```diff
@@ -302,8 +302,8 @@ def _create_lock(path):
     file_system = _get_fs(path)
 
     if _fs_support_globalflock(file_system):
-        log.debug("Using flock.")
+        # log.debug("Using flock.")
```
bouthilx, line 305:
```suggestion
        # log.debug("Using flock.")
```
```suggestion
        log.debug("Using flock.")
```

src/orion/core/io/database/pickleddb.py
```diff
@@ -302,8 +302,8 @@ def _create_lock(path):
...
         return FileLock(path)
     else:
-        log.debug("Cluster does not support flock. Falling back to softfilelock.")
+        # log.debug("Cluster does not support flock. Falling back to softfilelock.")
```
bouthilx, line 308:
```suggestion
        log.debug("Cluster does not support flock. Falling back to softfilelock.")
```

src/orion/core/cli/serve.py
Delaunay: fixed

src/orion/storage/base.py
```diff
@@ -558,27 +558,17 @@ def acquire_algorithm_lock(self, experiment, timeout=600, retry_interval=1):
...
 
 
-storage_factory = GenericSingletonFactory(BaseStorageProtocol)
+storage_factory = GenericFactory(BaseStorageProtocol)
 
 
-def get_storage():
-    """Return current storage
+@contextlib.contextmanager
+def with_storage(config):
```
bouthilx, line 565: Is this used for tests only?

src/orion/testing/state.py
```diff
@@ -179,33 +181,35 @@ def replace_file(v):
...
+            self.storage = setup_storage()
+            return self.storage
 
         try:
+            self.storage_config = copy.deepcopy(config)
             config["of_type"] = config.pop("type")
-            db = storage_factory.create(**config)
-            self.storage_config = config
+            self.storage = storage_factory.create(**config)
         except SingletonAlreadyInstantiatedError:
```
bouthilx, line 205: Does it make any sense to keep the error SingletonAlreadyInstantiatedError? I don't see why.

src/orion/testing/state.py
```diff
@@ -82,7 +81,10 @@ def __init__(
 
         self.tempfile = None
         self.tempfile_path = None
+
+        self.previous_config = copy.deepcopy(orion.core.config.storage.to_dict())
```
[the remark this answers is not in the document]
bouthilx: Ah, it swaps the global config. The class docstring should be adjusted.

tests/functional/backward_compatibility/test_versions.py
```diff
@@ -237,7 +237,7 @@ def build_storage():
     null_db_instances()
     experiment_builder.setup_storage()
```
bouthilx, line 238: Why do we need to call setup_storage twice?

tests/functional/client/test_cli_client.py
```diff
@@ -49,79 +49,82 @@ def test_interrupt(monkeypatch, capsys):
         assert trials[0].status == "interrupted"
 
 
-def test_interrupt_diff_code(storage, monkeypatch, capsys):
+def test_interrupt_diff_code(monkeypatch, capsys):
     """Test interruption from within user script with custom int code"""
-    monkeypatch.chdir(os.path.dirname(os.path.abspath(__file__)))
```
bouthilx, line 54: Do you remember why the fixture `storage` would not work correctly anymore when getting rid of the singleton?

bouthilx: If the issue is because global config is not set properly, that I think it would be best to fix this inside the fixture so we don't need to add back the OrionState boilerplate everywhere.

Delaunay: I don't remember if it was an actual issue.

But I also anted to reduce the number times storage are instantiated, since it made debugging easier to track down.
Some test bugged because they ended up using the wrong storage, so having less instantiation to go through helped a lot.

src/orion/storage/base.py
Delaunay: I dont think it is used right now, it was a helper, could become useful. want it removed ?

src/orion/testing/state.py
Delaunay: done

src/orion/testing/state.py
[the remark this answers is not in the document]
Delaunay: done

src/orion/benchmark/benchmark_client.py
Delaunay: I don't like it much because it makes it hard to track the number of storage instantiated.
I think the storage should be the core of the Orion client instead of the experiment

src/orion/storage/base.py
Delaunay: it is removed

tests/functional/client/test_cli_client.py
Delaunay: works, done

src/orion/benchmark/benchmark_client.py
bouthilx: You mean that it is hard to figure out how many storage objects have been created through `create_experiment/benchmark` calls with storage options? 

Yeah, it's not much more cumbersome to just wrap the config in another function to create the storage. 
```
benchmark = create_benchmark(storage=setup_storage(storage_config))
```
I would agree to keep storage only. We'd need to support storage as a configuration too during a deprecation phase however.

src/orion/storage/base.py
bouthilx: Yeah, less maintenance. Thanks!

tests/functional/client/test_cli_client.py
bouthilx: Yeah, that's what I meant. :) Thanks!

src/orion/client/__init__.py
```diff
@@ -209,10 +205,10 @@ def build_experiment(
             "max_idle_time is deprecated. Use experiment.workon(reservation_timeout) instead."
         )
 
-    setup_storage(storage=storage, debug=debug)
+    builder = experiment_builder.ExperimentBuilder(storage, debug)
```
bouthilx, line 208: Docstring must be adjusted accordingly.

bouthilx: Ah, no, you are still passing the configuration. Sorry, forget about my comment above.

src/orion/core/io/config.py
```diff
@@ -451,3 +451,35 @@ def to_dict(self):
...
+        """Set the configuration from a dictionary"""
+
+        logger.debug("Setting config to %s", config)
+        logger.debug("Config was %s", repr(self))
+
+        with _disable_logger():
+            for key in self._config:  # pylint: disable=consider-using-dict-items
+                value = config.get(key)
+
+                if value:
```
bouthilx, line 465: Seems like this would miss 0 and '' items.

src/orion/core/io/config.py
```diff
@@ -451,3 +451,35 @@ def to_dict(self):
...
+                    self[key] = value
+                else:
+                    self._config[key].pop("value", None)
```
bouthilx, line 468: This will reset to default config?

Delaunay: Yes

src/orion/core/io/config.py
Delaunay: fixed

src/orion/benchmark/__init__.py
```diff
@@ -20,8 +21,12 @@ class Benchmark:
 
     Parameters
     ----------
+    storage: Storage
```
lebrice, line 24: Do we have a `Storage` class now?
Perhaps we could rename `BaseStorageProtocol` -> `Storage` in another PR, what do you think?

src/orion/core/io/config.py
```diff
@@ -451,3 +451,35 @@ def to_dict(self):
...
+        """Set the configuration from a dictionary"""
+
+        logger.debug("Setting config to %s", config)
+        logger.debug("Config was %s", repr(self))
+
+        with _disable_logger():
+            for key in self._config:  # pylint: disable=consider-using-dict-items
+                value = config.get(key)
+
+                if value is not None:
```
bouthilx, line 465: Maybe they want to set this to None? :man_shrugging:

src/orion/core/io/config.py
```diff
@@ -451,3 +451,35 @@ def to_dict(self):
                 config[key] = self[key].to_dict()
 
         return config
+
+    def from_dict(self, config):
```
bouthilx, line 455: I think this new method would deserve a couple of tests in `tests/unittests/core/io/test_config.py`. Among other things, what if the config contains additional keys? What if it lacks some keys? What if a value is not valid? What if it's a deprecated key in the config? Are subconfiguration handled properly? Even values that are dict?

src/orion/core/io/config.py
```diff
@@ -451,3 +451,35 @@ def to_dict(self):
...
+    def __repr__(self) -> str:
+        confs = []
+        for k, v in self._config.items():
+            if "value" in v:
+                confs.append(f'{k}: {repr(v["value"])}')
+
+        for k, v in self._subconfigs.items():
+            confs.append(f"{k}: {repr(v)}")
+
+        return "{" + ", ".join(confs) + "}"
```
bouthilx, line 485: Would `pprint.pformat(self.to_dict())` do the job? I would require less testing.

src/orion/benchmark/__init__.py
Delaunay: updated, yes, no problem with renaming it later

src/orion/core/worker/experiment.py
```diff
@@ -153,7 +152,10 @@ def __init__(
         working_dir: str | None = None,
         metadata: dict | None = None,
         refers: dict | None = None,
+        storage=None,
```
bouthilx, line 155:
```suggestion
        storage: BaseStorageProtocol | None = None,
```
That's a bit weird actually, followed by the assert. Making it a positional argument would broken many parts of the code I guess?

src/orion/core/worker/experiment.py
```diff
@@ -163,12 +165,19 @@ def __init__(
         self.metadata = metadata or {}
         self.max_trials = max_trials
         self.max_broken = max_broken
-
         self.algorithms = algorithms
         self.working_dir = working_dir
 
-        self._storage = get_storage()
-        self._node = ExperimentNode(self.name, self.version, experiment=self)
+        self._storage = storage or setup_storage()
```
bouthilx, line 171: The `or` should never be executed here now, otherwise the assert above would raise.

src/orion/core/worker/experiment.py
```diff
@@ -190,20 +199,12 @@ def __getstate__(self):
         for entry in self.__slots__:
             state[entry] = getattr(self, entry)
 
-        # TODO: This should be removed when singletons and `get_storage()` are removed.
```
bouthilx, line 193: :tada: :tada: :tada: 

Sorry for spamming. This change makes me very happy. :smile:

src/orion/core/worker/experiment.py
Delaunay: yeah that is the main problem

src/orion/testing/__init__.py
```diff
@@ -147,7 +150,7 @@ def generate_benchmark_experiments_trials(
 
 @contextmanager
 def create_study_experiments(
-    exp_config, trial_config, algorithms, task_number, max_trial, n_workers=(1,)
+    state, exp_config, trial_config, algorithms, task_number, max_trial, n_workers=(1,)
```
bouthilx, line 153: This function would not need to be a contextmanager anymore right? But in any case, let's leave turning it into a standard function for another PR otherwise it will break all the test diffs.

src/orion/testing/__init__.py
```diff
@@ -221,6 +231,38 @@ def create_experiment(exp_config=None, trial_config=None, statuses=None):
     client.close()
 
 
+@contextmanager
+def falcon_client(exp_config=None, trial_config=None, statuses=None):
```
bouthilx, line 235: Why do we need to duplicate much of the code of `create_experiment`? Could we instead re-use `create_experiment` within this function?

src/orion/core/io/config.py
Delaunay: * what if the config contains additional keys

we iterate over the _config keys not the provided config so those are ignored

* What if it lacks some keys

we iterate over the `_config` keys  not `config` so we iterate over everything that is defined and pop what is not

* What if a value is not valid

it uses ``Configuration.__setitem__`` so the value gets validated and the expected TypeError exception is raised

* Are subconfiguration handled properly?

Yes, storage has the Database subconfig.

bouthilx: Thanks for the answers, I could see it from the code. :grinning: These questions were meant as potential test-case for the unittests.

tests/conftest.py
```diff
@@ -392,3 +370,20 @@ def random_dt(monkeypatch):
     """Make ``datetime.datetime.utcnow()`` return an arbitrary date."""
     with mocked_datetime(monkeypatch) as datetime:
         yield datetime.utcnow()
+
+
+@pytest.fixture(scope="function")
+def setup_pickleddb_database():
+    """Configure the database"""
+    with OrionState() as cfg:
```
bouthilx, line 378: It this still relying on a tmp file for the DB?

tests/conftest.py
```diff
@@ -392,3 +370,20 @@ def random_dt(monkeypatch):
...
+        yield cfg
+
+
+@pytest.fixture(scope="function")
+def storage(setup_pickleddb_database):
+    yield setup_pickleddb_database.storage
+
+
+@pytest.fixture(scope="function")
+def orionstate(setup_pickleddb_database):
```
bouthilx, line 388: Maybe the use of setup_pickleddb_database should be renamed to orionstate instead. There are many instances of `cfg = setup_pickleddb_database`.

src/orion/core/io/config.py
Delaunay: I am going to add the tests

tests/conftest.py
Delaunay: yes

Delaunay: `OrionState.make_config`

tests/functional/configuration/test_all_options.py
```diff
@@ -37,14 +37,17 @@ def with_storage_fork(func):
     def call(*args, **kwargs):
 
         with tempfile.NamedTemporaryFile(delete=True) as tmp_file:
-            storage = get_storage()
+
+            storage = setup_storage()
             old_path = storage._db.host
-            storage._db.host = tmp_file.name
+
+            # storage._db.host = tmp_file.name
```
bouthilx, line 44:
```suggestion
```

tests/functional/demo/test_demo.py
```diff
@@ -243,30 +241,84 @@ def test_demo_inexecutable_script(storage, monkeypatch, capsys):
     assert "User script is not executable" in captured
 
 
-def test_demo_four_workers(storage, monkeypatch):
+@contextmanager
+def generate_config(template):
+    """Generate a configuration file inside a temporary directory with the current storage config"""
+    tmp_path = tempfile.mkdtemp("orion_test")
```
bouthilx, line 247: If it relied on `tmp_path` as argument, passed from Pytest's `tmp_path`, that would ensure the folder is cleared even when the test fails.

tests/functional/demo/test_demo.py
```diff
@@ -243,30 +241,84 @@ def test_demo_inexecutable_script(storage, monkeypatch, capsys):
...
+
+    shutil.rmtree(tmp_path)
+
+
+def logging_directory():
+    base_repo = os.path.dirname(os.path.abspath(orion.core.__file__))
+    logdir = os.path.abspath(os.path.join(base_repo, "..", "..", "..", "logdir"))
+
+    try:
+        shutil.rmtree(logdir)
```
bouthilx, line 270: Maybe with `ignore_errors=True` instead?

tests/unittests/core/io/test_experiment_builder.py
```diff
@@ -249,11 +252,17 @@ def test_get_from_args_no_hit(config_file):
...
     """Try building experiment view when in db"""
-    cmdargs = {"name": "supernaekei", "config": config_file}
+    cmdargs = {"name": "supernaekei", "config": raw_config}
+
+    with OrionState(experiments=[new_config], trials=[]) as cfg:
+
+        def get_storage(*args, **kwargs):
+            return cfg.storage
+
+        monkeypatch.setattr(experiment_builder, "setup_storage", get_storage)
```
bouthilx, line 264: Why do you need to replace `setup_storage` in the next couple tests? Isn't the global `storage` config set with `OrionState`?

tests/unittests/core/worker/test_trial_pacemaker.py
```diff
@@ -35,15 +35,17 @@ def trial(exp):
...
 
     return trial
 
 
-@pytest.mark.usefixtures("storage")
 def test_trial_update_heartbeat(exp, trial):
     """Test that the heartbeat of a trial has been updated."""
-    trial_monitor = TrialPacemaker(trial, wait_time=1)
+
+    storage = setup_storage()
```
bouthilx, line 46: Maybe it would be best to use orionstate as a fixture instead to avoid polluting local DB.

tests/unittests/core/worker/test_trial_pacemaker.py
```diff
@@ -62,10 +64,11 @@ def test_trial_update_heartbeat(exp, trial):
     trial_monitor.stop()
 
 
-@pytest.mark.usefixtures("storage")
 def test_trial_heartbeat_not_updated(exp, trial):
     """Test that the heartbeat of a trial is not updated when trial is not longer reserved."""
-    trial_monitor = TrialPacemaker(trial, wait_time=1)
+    storage = setup_storage()
```
bouthilx, line 69: Same here

tests/unittests/core/worker/test_trial_pacemaker.py
bouthilx: Ah, it's set in `exp` fixture. Alright.

tests/unittests/storage/test_legacy.py
```diff
@@ -86,51 +79,38 @@ def test_setup_database_bad():
...
+
+    database = setup_database({"type": "pickleddb", "host": "test.pkl"})
     assert isinstance(database, PickledDB)
-    with pytest.raises(SingletonAlreadyInstantiatedError) as exc:
-        setup_database({"type": "mongodb"})
 
-    assert exc.match(r"A singleton instance of \(type: Database\)")
+    # with pytest.raises(SingletonAlreadyInstantiatedError) as exc:
+    #     setup_database({"type": "mongodb"})
+    # assert exc.match(r"A singleton instance of \(type: Database\)")
```
bouthilx, line 97:
```suggestion
```

tests/unittests/storage/test_legacy.py
```diff
@@ -86,51 +79,38 @@ def test_setup_database_bad():
...
-    """Test that get database fails if no database singleton exist"""
-    update_singletons()
-    with pytest.raises(SingletonNotInstantiatedError) as exc:
-        get_database()
 
-    assert exc.match(r"No singleton instance of \(type: Database\) was created")
+    database = setup_database({"type": "pickleddb", "host": "test.pkl"})
+    assert isinstance(database, PickledDB)
+    # with pytest.raises(SingletonAlreadyInstantiatedError):
+    #     setup_database({"type": "pickleddb", "host": "other.pkl"})
```
bouthilx, line 106:
```suggestion
```

src/orion/core/io/config.py
Delaunay: Updated using NOT_SET

src/orion/core/io/config.py
Delaunay: Tests added

src/orion/core/io/config.py
Delaunay: added

src/orion/core/worker/experiment.py
Delaunay: done

tests/functional/demo/test_demo.py
Delaunay: done

tests/conftest.py
Delaunay: done

tests/functional/demo/test_demo.py
Delaunay: done

src/orion/testing/__init__.py
Delaunay: done

src/orion/testing/__init__.py
Delaunay: done

src/orion/testing/__init__.py
bouthilx: You adjusted all the `with create_study_experiments` as well? :open_mouth:

tests/unittests/core/io/test_experiment_builder.py
Delaunay: because of `get_from_args` which will call `setup_storage` creating its own instance of the storage with the config coming form the `get_from_args` not the global config

src/orion/benchmark/benchmark_client.py
bouthilx: Done, thanks!

tests/unittests/core/io/test_experiment_builder.py
bouthilx: But it will use the global config if nothing is defined in a local config file passed to the arguments.

Delaunay: I added comments for clarity so people dont lose time on that :)

src/orion/testing/__init__.py
```diff
@@ -233,31 +233,12 @@ def create_experiment(exp_config=None, trial_config=None, statuses=None):
...
-        trials=generate_trials(trial_config, statuses, exp_config),
-    ) as cfg:
-
-        experiment = experiment_builder.build(
-            name=exp_config["name"], storage=cfg.storage_config
-        )
-
-        if cfg.trials:
-            experiment._id = cfg.trials[0]["experiment"]
+    cfg, experiment, exp_client = create_experiment(exp_config, trial_config, statuses)
```
bouthilx, line 237: I think this will cause issues. It should use `with create_experiment()` and then yield the falcon_client within the with-clause.

tests/unittests/core/io/test_experiment_builder.py
Delaunay: and `get_from_args` gets a different storage because it is configured to do so from ``orion_config.yaml``

bouthilx: Ah I see, previous version of the tests were not using the local storage config as well, since the singleton was already instantiated by OrionState.

Perhaps using a config file without a config for the storage would be simpler and less confusing than a mock.

src/orion/testing/__init__.py
Delaunay: right, done

tests/unittests/core/io/test_experiment_builder.py
Delaunay: done

src/orion/testing/__init__.py
Delaunay: yes

bouthilx (approved):
No more singletons. At last! :smile:

PR #943, opened 2022-06-07
Add a dashboard launcher

notoraptor:
# Description

Hi @bouthilx ! This is a PR to add a dashboard launcher.

# Changes

- Add a compiled code of dashboard into `dashboard/build`
- Add a script `orion frontend` to run dashboard build.
- Add a Github Actions to test dashboard build through `orion frontend` script, using `selenium` framework.

# Checklist

## Tests

## Documentation

## Quality


dashboard/src/public/index.html
```diff
@@ -22,7 +22,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Oríon Dashboard</title>
+    <title>Orion Dashboard</title>
```
bouthilx, line 25: Why did these change? Encoding issues?

src/orion/core/cli/frontend.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+import logging
+
+from gunicorn.app.base import BaseApplication
+
+log = logging.getLogger(__name__)
+DESCRIPTION = "Starts Oríon Dashboard"
+
+
+import mimetypes
```
bouthilx, line 19: Could these be moved above?

dashboard/src/public/index.html
notoraptor: No, I thought it was a typo error. I will revert it!

notoraptor:
Hi @bouthilx PR updated. Dashboard build test CI is moved to `build.yml` and makes a `yarn build` before testing. Testing script is also now splitted into unit tests and runs with pytest in CI.

In last commit, I try to include `dashboard/build` folder into Python package. I tested it on my computed. It is indeed included in `tar.gz` generated file using `python setup.py sdist`, but it is not included in orion package when package file is installed. At most, it created a `dashboard` folder in `sys.prefix` path (because of `data_files` added in `setup.py`, see commit), but it is outside the orion installed package folder.

It seems the best way to correctly include `dashboard/build` is to move it inside the `src/orion` folder. What do you think?

bouthilx:
Thanks for the follow up! Do you think using `sys.prefix` could allow pointing to the dashboard folder?

src/orion/core/cli/frontend.py
```diff
@@ -0,0 +1,117 @@
...
+
+import falcon
+
+
+class StaticResource:
+    """Resource class to serve frontend files."""
+
+    STATIC_DIR = os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__), "..", "..", "..", "..", "dashboard", "build"
```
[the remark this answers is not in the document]
notoraptor: Indeed, it may lead to issues. I added commits to:
- install dashboard build using setup.py `data_files` ption
- locate dashboard build at runtime, either in installation or in source code

.github/workflows/build.yml
```diff
@@ -31,6 +31,44 @@ jobs:
           pip install tox tox-gh-actions
     - name: Run pre-tests with Tox
       run: tox -e ${{ matrix.toxenv }}
+
+  test-dashboard-build:
```
bouthilx, line 35: This should be added into the list of `needs` of PyPI and conda below.

setup.py
```diff
@@ -66,6 +66,13 @@
 }
 extras_require["all"] = sorted(set(sum(extras_require.values(), [])))
 
+dashboard_files = []
+for root, sub_directories, files in os.walk("dashboard/build"):
```
bouthilx, line 70: Does it need to

src/orion/core/cli/frontend.py
```diff
@@ -0,0 +1,145 @@
...
+                os.path.dirname(current_file_path),
+                "..",
+                "..",
+                "..",
+                "..",
+                "dashboard",
+                "build",
+            )
+        )
+    assert os.path.isdir(dashboard_build_path)
```
bouthilx, line 51: This should be if-raise. What message could we provide the user to explain this error? That the folder is not present?

src/orion/core/cli/frontend.py
```diff
@@ -0,0 +1,145 @@
...
+    assert os.path.isfile(os.path.join(dashboard_build_path, "index.html"))
```
bouthilx, line 52: Why should we verify this? Is it likely to occur? Do you think this would be best as a part of the tests instead?

.github/workflows/build.yml
notoraptor: Ok!

src/orion/core/cli/frontend.py
notoraptor: Ok, now it raises a runtime error with message `"Cannot find dashboard static files to run frontend"`.

notoraptor:
@bouthilx all tests passed !

src/orion/core/cli/frontend.py
notoraptor: Ok!

notoraptor:
Tests succeed !

notoraptor:
Rebased!

PR #964, opened 2022-07-07
Small fixes

abergeron:
# Description
This is a collection of small fixes for bugs I found while investigating the code.

If one of the fixes is questionable, I can remove the commit in question from the PR on request.

# Checklist

## Documentation
- [x] I have updated the relevant documentation related to my changes

## Quality
- [x] I have read the [CONTRIBUTING](https://github.com/Epistimio/orion/blob/develop/CONTRIBUTING.md) doc
- [x] My commits messages follow [this format](https://chris.beams.io/posts/git-commit/)
- [x] My code follows the style guidelines (`$ tox -e lint`)

PR #965, opened 2022-07-08
Add a code "map" section to the developer documentation

abergeron:
This should be merged after #964 to make sure all the links will work correctly, which is why I publish it as a draft.

Otherwise it is mostly ready if you want to read and comment.

# Checklist

## Tests
The new pre-commit hook fails for unrelated files.

## Documentation

## Quality


docs/src/developer/plan.rst
```diff
@@ -0,0 +1,97 @@
...
+The code in :py:func:`orion.core.cli.main` will parse the command line
+arguments and route to :py:func:`orion.core.cli.hunt.main`.
+
+The command line arguments are passed to
+:py:func:`orion.core.io.experiment_builder.build_from_args`. This will
+massage the parsed command line arguments and merge that configuration
+with the config file and the defaults with various helpers from
+:py:mod:`orion.core.io.resolve_config` to build the final
+configuration. The result is eventually handled off to
+:py:func:`orion.core.io.experiment_builder.create_experiment` to
```
bouthilx, line 24: May be worth mentioning here, is the way the experiment configuration is fetched from the DB and compared with current one if already existing in DB. It also handles race conditions if any when multiple `hunt` calls try to create the same experiment at the same time.

docs/src/developer/plan.rst
```diff
@@ -0,0 +1,97 @@
...
+:py:meth:`orion.client.runner.Runner.gather` which will wait on all
+currently registered futures with a timeout to get some results.
+
+Once we get those results we de-register the futures and record the
+results with
+:py:meth:`orion.client.experiment.ExperimentClient.observe` or update
+the count of broken trials if they did not finish successfully.
+
+Finally we monitor the total amount of time spent waiting for trials
+to finish.
```
bouthilx, line 97: May be worth explaining the termination criteria here and how the runner may exit.

bouthilx:
Thanks a lot for documentation the flow of execution during a call to `orion hunt`! I think it is quite complete, despite maybe a short mention of experiment version check/experiment creation race condition handling and a last section on termination criteria and how runners may exit (ex: broken experiment, completed experiment, reached max trials per worker).

abergeron:
Sorry it took a lot of time, but I think I've addressed the comments now.

abergeron:
I don't know what is wrong with tests/unittests/client/test_runner.py::test_runner_inside_dask, but I'm reasonably certain I didn't touch anything related to it and it keeps failing, but only on python 3.8.

docs/src/developer/plan.rst
```diff
@@ -0,0 +1,136 @@
...
+with various helpers from :py:mod:`orion.core.io.resolve_config` to
+build the final configuration. The result is eventually handled off to
+:py:func:`orion.core.io.experiment_builder.create_experiment` to
+create an :py:class:`orion.core.worker.experiment.Experiment` and set
+its properties.
+
+If the experiment is new, meaning it has no storage id, then it will
+attempt to save it to storage, which may conflict in case another
+instance of ``orion hunt`` is doing the same thing. The storage is
+responsible for repoting conflicts and
```
bouthilx, line 35:
```suggestion
responsible for repoting conflicts and
```
```suggestion
responsible for reporting conflicts and
```

docs/src/developer/plan.rst
```diff
@@ -0,0 +1,136 @@
...
+:py:attr:`orion.client.runner.Runner.is_running`.
+
+The experiment can also stop if too many trials fail, either because
+they fail to start, they crashed, were killed (like by an external job
+scheduler) or the take too much time to complete. This is checked in
+:py:meth:`orion.client.runner.Runner.gather` with
+:py:attr:`orion.client.runner.Runner.is_broken`.
+
+If one of the workers returns an unexpected result the experiment is
+also stop immediately because it is assume that something is wrong
```
bouthilx, line 129:
```suggestion
also stoppe immediately because it is assume that something is wrong
```

docs/src/developer/plan.rst
```diff
@@ -0,0 +1,136 @@
...
+The first obvious one is the configured maximum number of trials to
+run. If this is reached, then we stop running more. This is checked at
+the beginning of the loop with
+:py:attr:`orion.client.runner.Runner.is_running`.
+
+The experiment can also stop if too many trials fail, either because
+they fail to start, they crashed, were killed (like by an external job
+scheduler) or the take too much time to complete. This is checked in
+:py:meth:`orion.client.runner.Runner.gather` with
+:py:attr:`orion.client.runner.Runner.is_broken`.
```
bouthilx, line 126: There are 2 levels of max_trials/max_broken. There is at the level of the experiment. If we reach either max_trials or max_broken, all Runners will stop. And there is at the level of the `Runner` (under the config name `worker`, that's a bit confusing since the introduction of the Runner which now control multiple workers). If max_trials or max_broken is reached within the execution of this Runner, it will stop, but the other runner working on the same experiment may continue. 

See for instance in doc: 
https://orion.readthedocs.io/en/stable/user/config.html#max-trials
vs 
https://orion.readthedocs.io/en/stable/user/config.html#config-worker-max-trials

docs/src/developer/plan.rst
abergeron: If I were to explain that it would probably go into the sample/observe parts since that is where the Experiment limits are relevant.  Since I did not go into much detail in those parts, I did not explain that.

Maybe it would be worth to go into more detail about those, but I think that this could go in as a first approximation as it is and be refined later (if there is interest).

PR #968, opened 2022-07-11
Minor refactor: Make creation of `Experiment` objects more explicit

lebrice:
- Arguments are fully specified in `Experiment.__init__`. This is so we can avoid the current pattern of setting the attributes on the instance after it is initialized.
- Adds type annotations to the arguments of `Experiment.__init__` (and the attributes of `Experiment` objects), ~~with typed dictionaries for the `refers` and `metadata` attributes.~~ (Edit: Will be in another PR).
- Add type annotations to some of the functions in `experiment_builder.py`
- Raise a `UserWarning` if `experiment_builder.create_experiment` receives extra unused keyword arguments.

### DIFF
--- a/src/orion/core/worker/experiment.py
+++ b/src/orion/core/worker/experiment.py
@@ -23,7 +23,6 @@ from orion.algo.space import Space
 from orion.core.evc.adapters import BaseAdapter
 from orion.core.evc.experiment import ExperimentNode
 from orion.core.io.database import DuplicateKeyError
-from orion.core.io.space_builder import SpaceBuilder
 from orion.core.utils.exceptions import UnsupportedOperation
 from orion.core.utils.flatten import flatten
 from orion.core.utils.singleton import update_singletons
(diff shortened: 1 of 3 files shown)


src/orion/core/worker/experiment.py
```diff
@@ -133,22 +140,37 @@ class Experiment:
...
+        max_trials: int | None = None,
+        max_broken: int | None = None,
+        algorithms: BaseAlgorithm | None = None,
+        working_dir: str | None = None,
+        metadata: dict | None = None,
+        refers: dict | None = None,
+    ):
+        self._id = _id
         self.name = name
+        self.space: Space = (
```
bouthilx, line 160: Where is the experiment created with a space config instead of with the space object?

lebrice: Nowhere atm. I can remove this if you want.

lebrice: Removed in 6da0bc9c6ab1c1f11948846551a8b09e9ff0eef6

bouthilx (approved):

PR #993, opened 2022-08-30
add hpobench wrapper

donglinjy:
# Description
Add an Orion benchmark task wrapper over [HPOBench](https://github.com/automl/HPOBench).

## Tests

## Documentation

## Quality

### DIFF
--- a/docs/src/user/benchmark.rst
+++ b/docs/src/user/benchmark.rst
@@ -27,6 +27,15 @@ Beside out of box :doc:`/code/benchmark/task` and :doc:`/code/benchmark/assessme
 you can also extend benchmark to add new ``Tasks`` with :doc:`/code/benchmark/task/base` and
 ``Assessments`` with :doc:`/code/benchmark/assessment/base`,
 
+To run benchmark with task :doc:`/code/benchmark/task/hpobench`, use ``pip install orion[hpobench]``
+to install the extra requirements. HPOBench provides local and containerized benchmarks, but different
+local benchmark could ask for total difference extra requirements. With ``pip install orion[hpobench]``,
+you will be able to run local benchmark ``benchmarks.ml.tabular_benchmark``.
+If you want to run other benchmarks in local, refer HPOBench `Run a Benchmark Locally`_. To run
+containerized benchmarks, you will need to install `singularity`_.
+
 Learn how to get start using benchmark in Orion with this `sample notebook`_.
 
+.. _Run a Benchmark Locally: https://github.com/automl/HPOBench#run-a-benchmark-locally
+.. _singularity: https://singularity.hpcng.org/admin-docs/master/installation.html
 .. _sample notebook: https://github.com/Epistimio/orion/tree/develop/examples/benchmark/benchmark_get_start.ipynb
--- a/src/orion/algo/space/configspace.py
+++ b/src/orion/algo/space/configspace.py
@@ -23,6 +23,7 @@ try:
         IntegerHyperparameter,
         NormalFloatHyperparameter,
         NormalIntegerHyperparameter,
+        OrdinalHyperparameter,
         UniformFloatHyperparameter,
         UniformIntegerHyperparameter,
     )
@@ -44,6 +45,7 @@ except ImportError as err:
     UniformIntegerHyperparameter = DummyType
     NormalIntegerHyperparameter = DummyType
     CategoricalHyperparameter = DummyType
+    OrdinalHyperparameter = DummyType
 
 
 class UnsupportedPrior(Exception):
@@ -182,6 +184,13 @@ def to_oriondim(dim: Hyperparameter) -> Dimension:
 def _from_categorical(dim: CategoricalHyperparameter) -> Categorical:
     """Builds a categorical dimension from a categorical hyperparameter"""
     choices = {k: w for k, w in zip(dim.choices, dim.probabilities)}
+    return Categorical(dim.name, choices)
+
+
+@to_oriondim.register(OrdinalHyperparameter)
+def _from_ordinal(dim: OrdinalHyperparameter) -> Categorical:
+    """Builds a categorical dimension from a categorical hyperparameter"""
+    choices = list(dim.sequence)
     return Categorical(dim.name, choices)
 
 
--- a/src/orion/benchmark/task/hpobench.py
+++ b/src/orion/benchmark/task/hpobench.py
@@ -1,0 +1,86 @@
+"""
+Task for HPOBench
+=================
+"""
+import importlib
+import subprocess
+from typing import Dict, List, Union
+
+from orion.algo.space.configspace import to_orionspace
+from orion.benchmark.task.base import BenchmarkTask
+
+
+class HPOBench(BenchmarkTask):
+    """Benchmark Task wrapper over HPOBench (https://github.com/automl/HPOBench)
+
+    For more information on HPOBench, see original paper at https://arxiv.org/abs/2109.06716.
+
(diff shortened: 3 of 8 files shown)


donglinjy:
This could be a potential way to re-use singularity container in `HPOBench`, basically create it first when create Orion benchmark task and save its socket id, then create it again with this socket_id when run the task. 

But this may not work in multiple scenario as we could create Orion benchmark task at local client node and run the task at a different node.
https://github.com/automl/HPOBench/blob/d8b45b1eca9a61c63fe79cdfbe509f77d3f5c779/hpobench/container/client_abstract_benchmark.py#L106

docs/src/user/benchmark.rst
```diff
@@ -27,6 +27,15 @@ Beside out of box :doc:`/code/benchmark/task` and :doc:`/code/benchmark/assessme
 you can also extend benchmark to add new ``Tasks`` with :doc:`/code/benchmark/task/base` and
 ``Assessments`` with :doc:`/code/benchmark/assessment/base`,
 
+To run benchmark with task :doc:`/code/benchmark/task/hpobench`, use ``pip install orion[hpobench]``
+to install the extra requirements. HPOBench provides local and containerized benchmarks, but different
+local benchmark could ask for total difference extra requirements. With ``pip install orion[hpobench]``,
+you will be able to run local benchmark ``benchmarks.ml.tabular_benchmark``.
+If you want to run other benchmarks in local, refer HPOBench `Run a Benchmark Locally`_. To run
+containerized benchmarks, you will need to install `singularity`_.
```
bouthilx, line 35: Maybe there should be a quick word on how to use it with singularity once singularity is installed. Even if the usage is exactly the same, then it should be mentioned.

src/orion/algo/space/configspace.py
```diff
@@ -185,6 +187,13 @@ def _from_categorical(dim: CategoricalHyperparameter) -> Categorical:
     return Categorical(dim.name, choices)
 
 
+@to_oriondim.register(OrdinalHyperparameter)
+def _from_ordinal(dim: OrdinalHyperparameter) -> Categorical:
```
bouthilx, line 191: Is ordinal best mapped to categorical or integer? Categorical is loosing the importance of the ordering.

src/orion/benchmark/task/hpobench.py
```diff
@@ -0,0 +1,86 @@
...
+    benchmark_kwargs: str
+        Optional parameters to create benchmark instance of class `hpo_benchmark_class`.
+    objective_function_kwargs: dict
+        Optional parameters to use when calling `objective_function` of the benchmark instance.
+    """
+
+    def __init__(
+        self,
+        max_trials: int,
+        hpo_benchmark_class: Union[str, None] = None,
```
bouthilx, line 38: Should it support `None` as default? Or maybe `''` by default to simplify typing.

src/orion/benchmark/task/hpobench.py
```diff
@@ -0,0 +1,86 @@
...
+        benchmark_kwargs: dict = dict(),
+        objective_function_kwargs: dict = dict(),
```
bouthilx, line 40: Having non-mutable objects as defaults is a bit risky.

src/orion/benchmark/task/hpobench.py
```diff
@@ -0,0 +1,86 @@
...
+            benchmark_kwargs=benchmark_kwargs,
+            objective_function_kwargs=objective_function_kwargs,
+        )
+        self._verify_benchmark(hpo_benchmark_class)
+        self.hpo_benchmark_cls = self._load_benchmark(hpo_benchmark_class)
+        self.benchmark_kwargs = benchmark_kwargs
+        self.objective_function_kwargs = objective_function_kwargs
+
+    def call(self, **kwargs) -> List[Dict]:
+        hpo_benchmark = self.hpo_benchmark_cls(**self.benchmark_kwargs)
```
bouthilx, line 54: @donglinjy Did you figure out why the singularity container is destroyed at the end of the subprocess call? If we could avoid this then we could only pay the price of building it during task instantiation.

tests/unittests/benchmark/task/test_tasks.py
```diff
@@ -1,7 +1,19 @@
...
 """Tests for :mod:`orion.benchmark.task`."""
+import inspect
 
-from orion.benchmark.task import Branin, CarromTable, EggHolder, RosenBrock
+import pytest
+
+from orion.algo.space import Space
+from orion.benchmark.task import Branin, CarromTable, EggHolder, HPOBench, RosenBrock
+
+try:
```
bouthilx, line 10: Not really necessary, but you could use the class ImportOptional like [here](https://github.com/Epistimio/orion/blob/f01220c58864f81421bb9750067e8b8593f9b2d0/src/orion/algo/dehb/dehb.py#L23) and [here](https://github.com/Epistimio/orion/blob/develop/tests/unittests/algo/test_dehb.py#L13).

bouthilx: And HPOBench will make require additional tests for singularity, and so would best be in a separate test module.

src/orion/benchmark/task/hpobench.py
donglinjy: I think this is caused by how we now run the trials, if we create the containerized benchmark during task init, we serialize the task instance and then deserialize it as a new task instance to run, after the run complete, this new instance will be destroyed which will cause the singularity container shutdown too. Although I mentioned another possible solution at https://github.com/Epistimio/orion/pull/993#issuecomment-1253803140, which will ask some additional change to make it work in remote workers scenario.

tests/unittests/benchmark/task/test_tasks.py
donglinjy: Sure, let me see If I can change it to similar stuff as ImportOptional, I actually saw such ImportOptional code and was just trying to make it simple in the test cases as `profet`.

I have not figured out an easy way to install `singularity` in our test env, maybe I can try it a little bit.

tests/unittests/benchmark/task/test_tasks.py
```diff
@@ -101,3 +113,72 @@ def test_search_space(self):
...
         assert task.get_search_space() == {"x": "uniform(-5, 10, shape=2)"}
+
+
+@pytest.mark.skipif(
+    IMPORT_ERROR is not None,
+    reason="Running without HPOBench",
+)
+class TestHPOBench:
+    """Test benchmark task HPOBenchWrapper"""
+
```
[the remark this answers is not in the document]
donglinjy: This can be true if we have `singularity` installed.  But we can also test invalid benchmark names without `singularity`, because invalid name will cause the creation of benchmark fail before it reachs the point of using `singularity`.

src/orion/algo/space/configspace.py
donglinjy: ordinal values does not have to be `int`. https://automl.github.io/ConfigSpace/main/api/hyperparameters.html#ConfigSpace.hyperparameters.OrdinalHyperparameter

src/orion/benchmark/task/hpobench.py
donglinjy: wow, I forgot what I was thinking ^^

donglinjy:
@bouthilx I already added the work-around for runner test case to succeed in https://github.com/Epistimio/orion/pull/993/commits/3afad9a78454cc5b7fb02d63887baccbe61db314. But there build will fail in python 3.10 test added recently, and `HPOBench` can only support to run up to python version 3.10.0. Any suggestion about this?

```
ERROR: Package 'hpobench' requires a different Python: 3.10.8 not in '<=3.10,>=3.6'
```

PR #1003, opened 2022-09-29
Add assert_contains to print a more precise error message

Delaunay:


bouthilx:
Merged as part of #1013

PR #1007, opened 2022-10-13
Tests for speechbrain tutorial

Simnol22:
# Description
Basic tests for the SpeechBrain model example we are adding to the doc,.

# Changes
Adding of some tests, as well as a download_data.py script.

## Tests
- [x ] All new and existing tests are passing (`$ tox -e py38`; replace `38` by your Python version if necessary)

## Documentation
- [ x] I have updated the relevant documentation related to my changes

## Quality
- [ x] I have read the [CONTRIBUTING](https://github.com/Epistimio/orion/blob/develop/CONTRIBUTING.md) doc
- [ x] My commits messages follow [this format](https://chris.beams.io/posts/git-commit/)
- [ x] My code follows the style guidelines (`$ tox -e lint`)


examples/speechbrain_tutorial/mini_librispeech_prepare.py
```diff
@@ -0,0 +1,186 @@
+"""
+Downloads and creates manifest files for speech recognition with Mini LibriSpeech.
```
bouthilx, line 2: There should be a link to the original version, and perhaps a short explanation of the changes done.

examples/speechbrain_tutorial/download_data.py
```diff
@@ -0,0 +1,36 @@
...
+    # Data preparation, to be run on only one process.
+    sb.utils.distributed.run_on_main(
+        prepare_mini_librispeech,
+        kwargs={
+            "data_folder": hparams["data_folder"],
+            "save_json_train": hparams["train_annotation"],
+            "save_json_valid": hparams["valid_annotation"],
+            "save_json_test": hparams["test_annotation"],
+        },
+    )
```
bouthilx, line 36: Seems like this is all reimplemented in `main.py`. Maybe it should be turned into a function that can be reused. Or add an argument in main.py to avoid the later training steps.

tests/functional/example/test_speechbrain_tutorial.py
```diff
@@ -0,0 +1,159 @@
...
+
+def json_clean(path, scale):
+    """Modifies json file to reduce it to some scale."""
+    f = open(path)
+    json_list = json.load(f)
+    new_list = {}
+    ctr = 0
+
+    for item in json_list:
+        if ctr < len(json_list) * scale:
```
bouthilx, line 21: Here `json_list` is a list of examples in a dataset, right?

examples/speechbrain_tutorial/mini_librispeech_prepare.py
Simnol22: There is no change here, so I will add the link to the original version.

examples/speechbrain_tutorial/download_data.py
Simnol22: True ! But I have to use `hparams`, which is defined here. I will simply make a `download` function and make it return the hparams. Then I can call the function both times instead of duplicating code !

tests/functional/example/test_speechbrain_tutorial.py
Simnol22: Yes it is

tests/functional/example/test_speechbrain_tutorial.py
```diff
@@ -0,0 +1,159 @@
...
+
+@pytest.fixture(scope="module")
+def download_data(tmp_path_factory):
+    # Creating paths
+    path = tmp_path_factory.mktemp("out")
+    data = path / "data"
+    output = data / "results"
+
+    """Calling the script, for downloading the data"""
+    script = os.path.abspath("examples/speechbrain_tutorial/download_data.py")
```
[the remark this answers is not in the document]
Simnol22: Not sure I understand what do you mean here. Were you thinking something like `str(os.path.dirname(os.path.abspath(__file__))) + "/../../speechbrain_tutorial/main.py"` ?

examples/speechbrain_tutorial/mini_librispeech_prepare.py
bouthilx: Fixed in 3c801219f00ff9ee3776b8ff4e3bcc20964ad513, thanks!

examples/speechbrain_tutorial/download_data.py
bouthilx: Done in 3c801219f00ff9ee3776b8ff4e3bcc20964ad513. Thanks!

bouthilx:
The pre-trained model is too large for github-actions' runner. We would need to have a local hosted runner for this. We'll take a look at this and then revisit this PR.

PR #1020, opened 2022-11-07
Add matplotlib dependency to pb2

Delaunay:
GPy imports it all the time but does not list it as a dependency

See https://github.com/Epistimio/hydra_orion_sweeper/actions/runs/3330026395/jobs/5507981924#step:5:248


bouthilx (approved):
Thanks!

Issue #1015, opened 2022-10-31
Trials are manually inserted with "interrupted" instead of "new" when reserved=False


Delaunay:
FIX

```
         if not reserve:
-            self.release(trial)
+            self.release(trial, status='new')
```

Delaunay:
Fixed in Orion As a Service https://github.com/Epistimio/orion/pull/1023

PR #1023, opened 2022-11-08
Orion service 2

Delaunay:

# Description

# Changes

# Checklist

## Tests

## Documentation

## Quality

# Further comments


Delaunay:
Fixes https://github.com/Epistimio/orion/issues/1015 too

src/orion/client/experiment.py
```diff
@@ -41,6 +41,8 @@
 
 log = logging.getLogger(__name__)
 
+STOPPED_STATUS = {"completed", "interrupted", "suspended"}
```
bouthilx, line 44: This should be part of worker/trial.py

src/orion/client/experiment.py
```diff
@@ -877,11 +885,28 @@ def _verify_reservation(self, trial):
             raise RuntimeError(f"Reservation for trial {trial.id} has been lost.")
 
     def _maintain_reservation(self, trial):
-        self._pacemakers[trial.id] = TrialPacemaker(trial, self.storage)
+        self._pacemakers[trial.id] = TrialPacemaker(trial, self)
         self._pacemakers[trial.id].start()
 
+    def _update_heardbeat(self, trial):
```
bouthilx, line 891: Why move this inside the experiment client instead of keeping it in the Pacemaker?

src/orion/client/__init__.py
```diff
@@ -208,6 +208,44 @@ def build_experiment(
         If the algorithm or storage specified is not properly installed.
 
     """
+    return ExperimentClient.create_experiment(
```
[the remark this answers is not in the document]
Delaunay: yes

src/orion/core/io/database/mongodb.py
```diff
@@ -249,6 +264,15 @@ def write(self, collection_name, data, query=None):
         """
         dbcollection = self._db[collection_name]
 
+        if query is not None:
```
bouthilx, line 267: Is this really using something specific to mongodb about ownership of documents or it's simply an additional uid that could also be used in the other database backends?

src/orion/core/worker/experiment.py
```diff
@@ -380,6 +381,8 @@ def duplicate_pending_trials(self):
         the current experiment will only see the new copy of the trial.
         """
         self._check_if_writable()
+        # TODO: break orion as a service
+        evc_pending_trials = []
```
bouthilx, line 385: Looks like this line isn't changing anything no?

src/orion/core/worker/trial.py
```diff
@@ -205,6 +205,7 @@ class Param(Value):
 
     def __init__(self, **kwargs):
         """See attributes of `Trial` for meaning and possible arguments for `kwargs`."""
+        kwargs.pop("owner_id", None)
```
bouthilx, line 208: Perhaps it should be the database's responsibility to remove the `owner_id` if we are not supposed to be aware of it externally anyway.

bouthilx (commented):
Uniformize storage env variable names.

src/orion/service/client/experiment.py
```diff
@@ -0,0 +1,210 @@
...
+            heartbeat = orion.core.config.worker.heartbeat
+
+        self._experiment = experiment
+        self.heartbeat = heartbeat
+        self._executor = executor
+        self._executor_owner = False
+        self.storage_client = None
+        self.plot = PlotAccessor(self)
+
+    def to_pandas(self, with_evc_tree=False):
```
bouthilx, line 118: Why would it not be supported? It's just a read operation.

src/orion/service/client/experiment.py
```diff
@@ -0,0 +1,210 @@
...
+    @property
+    def storage(self):
+        """Return the storage currently in use by this client"""
+        raise NotImplementedError("REST API does not have a storage")
+
+    #
+    # Storage REST API
+    #
+
+    def insert(self, params, results=None, reserve=False):
```
bouthilx, line 180: It's not great that insert behaves differently. I did not look at the workon module but perhaps that is where insert should be, as making `insert(params, results=None, reserve=True)` is a use-case for working on a trial.

src/orion/service/client/experiment.py
```diff
@@ -0,0 +1,210 @@
...
+        remote_trial = self.workon_client.suggest(pool_size=pool_size)
+        return remote_trial
+
+    def observe(self, trial, results, name: str = "objective"):
+        """See `~ExperimentClient.observe`"""
+        self.workon_client.observe(trial, results, name=name)
+
+    def release(self, trial, status="interrupted"):
+        """See `~ExperimentClient.release`"""
+        # we probably should not expose this
```
bouthilx, line 155: Perhaps, but there should be a way to allow interupting a trial and so far `release` is the only way using the python API.

src/orion/client/experiment.py
Delaunay: The pacemaker was using the storage directly, but storage might not be instantiated for the service,
so the pacemaker need to be using the Client and the client will do what is right

src/orion/service/client/base.py
```diff
@@ -0,0 +1,146 @@
...
+    space: dict
+    version: str
+    mode: str
+    working_dir: str
+    metadata: dict
+    max_trials: int
+
+
+@dataclass
+class RemoteTrial:
```
bouthilx, line 40: Maybe this should be turned into a standard dataclass for trials exposed to users. The current version misses some attributes such as the objectives, constraints and stats.

src/orion/service/client/workon.py
```diff
@@ -0,0 +1,178 @@
...
+    def new_experiment(self, name, **config) -> RemoteException:
+        """Create a new experiment"""
+        payload = self._post("experiment", name=name, **config)
+
+        payload.pop("experiment_name", None)
+        payload["name"] = name
+        self.experiment = RemoteExperiment(**payload)
+        return self.experiment
+
+    def suggest(self, pool_size: int = 1, experiment_name=None) -> TrialCM:
```
bouthilx, line 51: Why should it support passing the experiment name as an argument? Shouldn't the WorkonClientREST be used only together with an existing experiment? After all it's used to replace `some_existing_experiment.workon`.

src/orion/service/client/workon.py
```diff
@@ -0,0 +1,178 @@
...
+        if experiment_name is None:
+            raise ExperiementIsNotSetup("experiment_name is not set")
+
+        log.debug("client: suggest: %s", experiment_name)
+        result = self._post(
+            "suggest", experiment_name=experiment_name, pool_size=pool_size
+        )
+
+        trials = []
+        for trial in result["trials"]:
```
bouthilx, line 64: Why take the time to convert them all if only the first one is returned?

src/orion/service/client/workon.py
```diff
@@ -0,0 +1,178 @@
...
+        if isinstance(results, numbers.Number):
+            results = [dict(value=results, name=name, type="objective")]
+
+        if experiment_name is None:
+            raise ExperiementIsNotSetup("experiment_name is not set")
+
+        if not isinstance(results, list):
+            raise TypeError("Results need to be a float or a list of values")
+
+        self._post(
```
bouthilx, line 127: What happens if an exception occurs remotely during observe. Will this `_post` raise a RemoteException?

src/orion/service/client/workon.py
```diff
@@ -0,0 +1,178 @@
...
+                euid=self.experiment_id,
+                experiment_name=experiment_name,
+                trial_hash=trial.id,
+                status=status,
+            )
+        except Exception as e:
+            failed = True
+            raise e
+        finally:
+            pacemaker = self._pacemakers.pop(trial.id, None)
```
bouthilx, line 100: Looks like this class could have something like `ExperimentClient._release_reservation`. Similar code is used in `observe` below.

src/orion/service/client/workon.py
```diff
@@ -0,0 +1,178 @@
...
+            pacemaker.stop()
+        else:
+            log.warning("Trial pacemaker not found! %s", trial.id)
+
+    def is_done(self) -> bool:
+        """returns true if the experiment is done"""
+        payload = self._post(
+            "is_done", experiment_name=self.experiment_name, euid=self.experiment_id
+        )
+        return payload.get("is_done", True)
```
bouthilx, line 147: What could cause the `payload` to not contain `is_done`? Would these situations mean that the experiment is indeed done?

src/orion/service/broker/broker.py
```diff
@@ -0,0 +1,235 @@
...
+        username=request.username,
+        password=request.password,
+        owner=request.username,
+    )
+
+    storage = Legacy(
+        database_instance=db,
+        # Skip setup, this is a shared database
+        # we might not have the permissions to do the setup
+        # and the setup should already be done anyway
```
bouthilx, line 98: When would the setup be done? We would need a script or command to do this when we set in place a shared database.

src/orion/service/broker/local.py
```diff
@@ -0,0 +1,198 @@
...
+                space=client._experiment.space.configuration,
+                version=client._experiment.version,
+                mode="x",
+                max_trials=client._experiment.max_trials,
+                working_dir=client._experiment.working_dir,
+                metadata=client._experiment.metadata,
+            )
+        )
+
+    def suggest(self, request: RequestContext) -> Dict:
```
bouthilx, line 55: This should handle the kind of exceptions than may be raised by suggest: https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L550-L565

src/orion/service/broker/local.py
```diff
@@ -0,0 +1,198 @@
...
+        client = retrieve_experiment_client(request, experiment_name)
+        storage = client.storage
+
+        trial_hash = request.data.pop("trial_hash")
+        results = request.data.get("results")
+
+        # pylint: disable=protected-access
+        trial = storage.get_trial(uid=trial_hash, experiment_uid=client._experiment.id)
+
+        assert trial is not None, "Trial not found"
```
bouthilx, line 84: if: raise seems more appropriate here.

src/orion/service/broker/local.py
```diff
@@ -0,0 +1,198 @@
...
+        client.observe(trial, results)
```
bouthilx, line 85: There are a few possible errors can could occur here: https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L628-L636.

src/orion/service/broker/local.py
```diff
@@ -0,0 +1,198 @@
...
+            raise BrokenExperiment()
+
+        return success(dict(is_done=client.is_done))
+
+    def heartbeat(self, request: RequestContext) -> Dict:
+        storage = get_storage_for_user(request)
+        trial_id = request.data.get("trial_id")
+
+        # pylint: disable=protected-access
+        results = storage._db._db["trials"].update_one(
```
bouthilx, line 103: Why not use storage.heartbeat? Is it because of the username? Shouldn't it be handled by this PR's modifications in MongoDB?

src/orion/service/broker/local.py
```diff
@@ -0,0 +1,198 @@
...
+        client = retrieve_experiment_client(request, experiment_name)
+
+        trial_hash = request.data.pop("trial_hash")
+        status = request.data.pop("status")
+        trial = client.get_trial(uid=trial_hash)
+
+        if trial is None:
+            raise ValueError(f"Trial {trial_hash} does not exist in database.")
+
+        client.release(trial, status)
```
bouthilx, line 130: Sorry, is there a mechanism that handles automatically the error and return them in a json format? I guess I missed it. There are also many potential errors that could occur here: https://github.com/Epistimio/orion/blob/develop/src/orion/client/experiment.py#L492-L499.

src/orion/service/broker/remote.py
```diff
@@ -0,0 +1,173 @@
...
+
+        trial["experiment"] = str(trial["experiment"])
+        trial.pop("heartbeat")
+        trial.pop("submit_time")
+        trial.pop("start_time")
+
+        return dict(status=0, result=dict(trials=[trial]))
+
+    def resume_from_experiment(self, experiment_name):
+        """Resume an experiment, create a new ExperimentClient and assign it to a worker"""
```
bouthilx, line 116: TODO?

src/orion/core/io/database/mongodb.py
Delaunay: It could be used for other database backends, ultimately the "database" backend is just mongodb 
and pickledb does not need the owner_id since there can only be one owner

src/orion/service/broker/broker.py
```diff
@@ -0,0 +1,235 @@
...
+    auth = AuthenticationService = None
+
+
+@dataclass
+class RequestContext:
+    """Request specific information useful throughout the request handling"""
+
+    service: ServiceContext  # Service config
+    username: str  # Populated on authentication
+    password: str
```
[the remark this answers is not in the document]
Delaunay: It is not, the password is looked up internally by the authentication layer using the API token

src/orion/service/broker/broker.py
Delaunay: Setup is done by IT when the database is created, `setup.sh` has a template of the code that could be executed to setup the database

src/orion/service/broker/local.py
Delaunay: They are handled at the top level by the [server](https://github.com/Epistimio/orion/pull/1023/files#diff-f6dfba39c2c9777888fc02f84c85109cdeedc4b952a29a4e0c3c2a46f6ef4178R90)

src/orion/service/broker/local.py
Delaunay: They are handled at the top level by the [server](https://github.com/Epistimio/orion/pull/1023/files#diff-f6dfba39c2c9777888fc02f84c85109cdeedc4b952a29a4e0c3c2a46f6ef4178R90)

src/orion/service/broker/local.py
Delaunay: IMO, that error should not occur in normal execution, that is why I put it as an assert.
Standard orion exception gets handled at the top level by the service itself, to provide a uniform RemoteException across the service

src/orion/service/broker/local.py
Delaunay: Yes, it is in the server, all the exceptions are handled. In fact they have to be because the RemoteClient is tested by the regular ExperimentClient unit tests to make sure the coverage is the same and avoid duplicating the tests

src/orion/service/client/experiment.py
Delaunay: Isn't the algo running on the server side generating the parameters ?
Workon is only fetching+reserving the trials.

The only use case for inserting from the API is if a user wants to manually add a trial to be worked on by which ever worker is available ?

src/orion/service/client/workon.py
Delaunay: Only one is ever returned, because that what `ExperimentClient` does.
But I think that for it might be useful to return more in the case of running a group of worker per machine.

https://github.com/Epistimio/orion/pull/1023/files#diff-eeae4d5b89cb07bd71b11d21ee0fadac1d072e7d37550e097b0264e3ab2e8b14R65

src/orion/service/client/workon.py
Delaunay: If on the server the ExperimentClient raises an exception, the server convert the exception as a json message.
The message will be read  by the client 

1. if the exception is part of the allowed exception set, then the exact the same exception that was raised by the ExperimentClient remotely is raised locally 
2. if the exception is not allowed (i.e internal server error) a RemoteException is rased 


The setup allow us to reuse the same tests for ExperimentClient and ExperimentClientREST
and provide usage consistency throughout the clients

src/orion/service/client/workon.py
Delaunay: that is just defensive programing

src/orion/service/service.py
```diff
@@ -0,0 +1,284 @@
...
+
+            with make_server(hostname, port, self.app) as httpd:
+                print(f"{os.getpid()}: Serving on port {port}...")
+                httpd.serve_forever()
+
+        except KeyboardInterrupt:
+            pass
+
+
+def main(
```
bouthilx, line 263: Why not add this in `orion.core.cli`?

src/orion/service/testing.py
```diff
@@ -0,0 +1,140 @@
+"""Integration testing between the REST service and the client"""
```
bouthilx, line 1: Could be good to add this under `orion.testing`. Perhaps in its own submodule there.

PR #1029, opened 2022-11-22
Add a tutorial for the new Warm-start feature

lebrice:
Adds an example of how to use the new warm-start feature.

Currently uses the Profet tasks, which require quite a bit of compute to run the first time (GPUS also help, but aren't required).

TODOS:
- [ ] Use more lightweight, realistic tasks.
- [ ] Make it pretty.


examples/tutorials/plot_5_warm_starting.py
```diff
@@ -0,0 +1,375 @@
...
+        f"({100.0 * correct / num_batches:.0f}%)\n"
+    )
+    return test_loss
+
+
+def main(**kwargs):
+    """Main loop. Trains and then tests a model after each epoch."""
+    # Training settings
+
+    # note: could also use simple-parsing to parse the Config from the command-line:
```
bouthilx, line 186: Sure!

examples/tutorials/plot_5_warm_starting.py
```diff
@@ -0,0 +1,375 @@
...
+
+    torch.manual_seed(args.seed)
+    device = args.device
+
+    train_kwargs = {"batch_size": args.batch_size}
+    test_kwargs = {"batch_size": args.test_batch_size}
+    if device.type == "cuda":
+        # Note: When using Orion with parallel workers (which is the case by default?),
+        # `num_workers` should be set to 0, because otherwise we get an error about daemonic
+        # processes having children, etc.
```
bouthilx, line 201: :scream: Really?? Forcing single-process data loader is terrible. We should fix this as soon as possible.

bouthilx: (But not as part of this PR, and this PR should not depend on this fix)

examples/tutorials/plot_5_warm_starting.py
```diff
@@ -0,0 +1,375 @@
...
+assert previous_experiment.storage
+assert previous_experiment.max_trials
+warm_experiments = [
+    build_experiment(
+        name=f"warm_experiment_{seed}",
+        space={"lr": "loguniform(1e-5, 1.0)"},
+        storage=current_experiment_storage,
+        executor=SingleExecutor(),
+        max_trials=current_experiment_n_runs,
+        # NOTE: This n_initial_points is changed slightly, since it also counts the trials from
```
bouthilx, line 328: It's interesting to make this fair comparison. In practice however would not need to force random points first at the previous results already provide a good basis to start bayesian optimization.

examples/tutorials/plot_5_warm_starting.py
bouthilx: Did you create an issue about this?

lebrice: #1060

examples/tutorials/plot_5_warm_starting.py
lebrice: Ok, should I keep this then? or not?

examples/tutorials/plot_5_warm_starting.py
lebrice: The note is meant for the users of the example. Enabling this would add `simple-parsing` as a dependency. I'd rather leave it as-is, if that's alrght with you.

PR #1030, opened 2022-11-22
Fix a few bugs in the Profet tasks

lebrice:


src/orion/benchmark/task/profet/model_utils.py
```diff
@@ -210,10 +210,15 @@ def load_data(
             X, Y, and C arrays.
         """
         # file = Path(input_path) / NAMES[benchmark]
-        file = Path(input_path) / self.json_file_name
+
```
bouthilx, line 213: Any idea why this path error was not caught before?

bouthilx (approved):
LGTM, thanks!

Issue #606, opened 2021-05-05
Get rid of `get_storage`


bouthilx:
Although it is _convenient_ to have a function to get the storage object from any part of the code it is not compatible with using different experiments with different storage nor is it compatible with multiprocessing where the other processes may not have access to the singleton. The storage should always be accessed through the experiment client (for now experiment object until we clearly split experiment client and experiment config objects).

We should get rid of the singletons and of all `get_storage` functions.

PR #1042, opened 2022-12-12
Remove references to get_storage #606

Delaunay:


docs/src/user/storage.rst
```diff
@@ -269,32 +269,26 @@ Although we recommend using the experiment client,
...
-There is two ways for creating the storage client. If you
-already created an experiment client, the storage
-was already created during the process of creating the
-experiment client and you can get it with
-:py:func:`orion.storage.base.get_storage`.
-Otherwise, you can create the storage client with
-:py:func:`orion.storage.base.setup_storage` before
-fetching it with
-:py:func:`get_storage() <orion.storage.base.get_storage>`.
+You should try to reuse the same storage throughout the your code.
```
lebrice, line 272: Just curious: why do we encourage reusing the same storage? (perhaps we could talk about the benefits here?)

For instance, I often pass a different Storage object to different experiments in the same script, for example when warm-starting. I know a few users (myself included) that tend to organize things with one Storage per sweep (either for scaling reasons with PickledDb, or just to keep everything nice and separated).

Delaunay: I will rephrase it; it is more preferring one instance of the same storage rather than multiple, having multiple storage doing different experiments is fine

bouthilx (approved):
LGTM, thanks!

PR #1049, opened 2022-12-19
Adding Ray Backend

Simnol22:
# Description
Adding a Ray backend for Orion ! Similar implementation of the Dask backend.

Solves github issue [ray_backend](https://github.com/Epistimio/orion/issues/1026)

# Changes
Adding a python file for the Ray class and its Future class. Modifying a little bit the executor test.

# Checklist

## Tests
- [x ] I added corresponding tests for bug fixes and new features. If possible, the tests fail without the changes

## Documentation
- [x ] I have updated the relevant documentation related to my changes

## Quality
- [ x] I have read the [CONTRIBUTING](https://github.com/Epistimio/orion/blob/develop/CONTRIBUTING.md) doc
- [ x] My commits messages follow [this format](https://chris.beams.io/posts/git-commit/)
- [ x] My code follows the style guidelines (`$ tox -e lint`)

### DIFF
--- a/src/orion/executor/ray_backend.py
+++ b/src/orion/executor/ray_backend.py
@@ -1,5 +1,6 @@
 import traceback
 import os
+import logging
 from orion.executor.base import (
     AsyncException,
     AsyncResult,
@@ -15,6 +16,7 @@ try:
 except ImportError:
     HAS_RAY = False
 
+logger = logging.getLogger(__name__)
 
 class _Future(Future):
     def __init__(self, future):
@@ -24,10 +26,10 @@ class _Future(Future):
     def get(self, timeout=None):
         if self.exception:
             raise self.exception
-        try:
+        try: 
             return ray.get(self.future,timeout=timeout)
         except ray.exceptions.GetTimeoutError as e:
-            print(e)
+            raise TimeoutError() from e
 
     def wait(self, timeout=None):
         try:
@@ -37,8 +39,10 @@ class _Future(Future):
         except Exception as e:
             self.exception = e
 
+
     def ready(self):
-        return self.future.future().done()
+        obj_ready = ray.wait([self.future])
+        return len(obj_ready[0]) == 1
 
     def successful(self):
         # Python 3.6 raise assertion error
@@ -51,6 +55,7 @@ class _Future(Future):
 class Ray(BaseExecutor):
     def __init__(self, n_workers=-1, runtime_env=None, **config,):
         super().__init__(n_workers=n_workers)
+        self.initialized = False
         if not HAS_RAY:
             raise ImportError("Ray must be installed to use Ray executor.")
         self.config = config
@@ -58,25 +63,34 @@ class Ray(BaseExecutor):
             if runtime_env is None:
                 ray.init()
             else:
-                ray.init(runtime_env=runtime_env)
-            print("Ray was initiated")
+                ray.init()#runtime_env={"working_dir": "/Users/simonolivier/DevOrion/orion/tests/unittests/executor"})
+            self.initialized = True
+            print("Ray was initiated with runtime_env : ", runtime_env)
     
     def __getstate__(self):
         return super().__getstate__()
 
     def __setstate__(self, state):
         super().__setstate__(state)
+    
+    def close(self):
+        if self.initialized:
+            self.initialized = False
+            ray.shutdown()
 
     def __del__(self):
-        ray.shutdown()
+        self.close()
+            
 
     def __enter__(self):
         return self
 
     def submit(self, function, *args, **kwargs):
         try:
-            remote_g = ray.remote(function)
-            return _Future(remote_g.remote(*args, **kwargs))
+            if (ray.is_initialized()):
+                remote_g = ray.remote(function)
+                return _Future(remote_g.remote(*args, **kwargs))
+            raise ExecutorClosed()
         except Exception as e:
             if str(e).startswith(
                 "Tried sending message after closing.  Status: closed"
@@ -84,10 +98,10 @@ class Ray(BaseExecutor):
                 raise ExecutorClosed() from e
 
             raise
-    def wait(self, futures):
+    def wait(self, futures):        
         return [future.get() for future in futures]
 
-    def async_get(self, futures, timeout=0.01):
+    def async_get(self, futures, timeout=None):
         results = []
         tobe_deleted = []
         for i, future in enumerate(futures):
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -39,6 +39,7 @@
         "sphinx_gallery",
     ],
     "dask": ["dask[complete]"],
+    "ray": ["ray[complete]"],
```
bouthilx, line 42:
```suggestion
    "ray": ["ray"],
```
I don't think the extra package `complete` exists for Ray. Maybe minimal dependencies would be sufficient? https://docs.ray.io/en/latest/ray-overview/installation.html

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+    def wait(self, timeout=None):
+        try:
+            ray.get(self.future, timeout=timeout)
+        except ray.exceptions.GetTimeoutError:
+            pass
+        except Exception as e:
+            self.exception = e
+
+    def ready(self):
+        obj_ready = ray.wait([self.future])
```
bouthilx, line 44: Would this work if we call ready twice? Can we call ray.wait twice on the same object? Also, `wait` will block until the future is done. `ready` should not be a blocking call, hence we should give a timeout arbitrarily small here (ex: 0.01 seconds).

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+        **config,
+    ):
+        super().__init__(n_workers=n_workers)
+        self.initialized = False
+        if not HAS_RAY:
+            raise ImportError("Ray must be installed to use Ray executor.")
+        self.config = config
+        if not ray.is_initialized():
+            if runtime_env is None:
+                ray.init()
```
bouthilx, line 69: `config` should be passed to `ray.init()`, and runtime_env would be contained within `config`.

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+            else:
+                ray.init(
+                    runtime_env={
+                        "working_dir": "/Users/simonolivier/DevOrion/orion/tests/unittests/executor"
+                    }
+                )
+            self.initialized = True
+            print("Ray was initiated with runtime_env : ", runtime_env)
```
bouthilx, line 77:
```suggestion
            logger.debug("Ray was initiated with runtime_env : %s", runtime_env)
```

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __getstate__(self):
+        return super().__getstate__()
+
+    def __setstate__(self, state):
+        super().__setstate__(state)
```
bouthilx, line 83: I don't think these are needed if nothing special is done.

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __enter__(self):
+        return self
+
+    def submit(self, function, *args, **kwargs):
+        try:
+            if ray.is_initialized():
+                remote_g = ray.remote(function)
+                return _Future(remote_g.remote(*args, **kwargs))
+            raise ExecutorClosed()
```
bouthilx, line 101:
```suggestion
            if not ray.is_initialized():
                raise ExecutorClosed()
                
            remote_g = ray.remote(function)
            return _Future(remote_g.remote(*args, **kwargs))
```
Nitpick :grimacing:

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+        except Exception as e:
+            if str(e).startswith(
```
bouthilx, line 103: Why catch the `ExecutorClosed()` previously raised and reraise it? Why catch it at all?

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,134 @@
...
+
+from orion.executor.base import (
+    AsyncException,
+    AsyncResult,
+    BaseExecutor,
+    ExecutorClosed,
+    Future,
+)
+
+try:
```
bouthilx, line 12: This should be implemented using the `ImportOptional` module, like [here](https://github.com/Epistimio/orion/blob/develop/src/orion/algo/hebo/hebo_algo.py#L28).

src/orion/executor/ray_backend.py
Simnol22: I first went with the try - catch inspired by the Dask backend. But with ray, it wasnt calling an ExecutorClosed error, which was required in the tests. So I manually checked if ray is initiated, but left the try - catch. Should I remove it altogether ? 
The function would then only contain : 

`if not ray.is_initialized():
     raise ExecutorClosed()
            
  remote_g = ray.remote(function)         
  return _Future(remote_g.remote(*args, **kwargs))`

src/orion/executor/ray_backend.py
Simnol22: So, here I did some tests and it can be called twice on the same object. The wait function does not block until the future is done, it simply returns the list of futures that are ready.

src/orion/executor/ray_backend.py
bouthilx: Yeah there is no need for a try-catch here.

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,125 @@
...
+
+class Ray(BaseExecutor):
+    def __init__(
+        self,
+        n_workers=-1,
+        **config,
+    ):
+        super().__init__(n_workers=n_workers)
+        self.initialized = False
+        if not HAS_RAY:
```
bouthilx, line 65:
```suggestion
        if not import_optional.failed:
```

src/orion/executor/ray_backend.py
```diff
@@ -0,0 +1,125 @@
...
+    Future,
+)
+
+with ImportOptional("ray") as import_optional:
+    import ray
+
+    HAS_RAY = True
+
+if import_optional.failed:
+    HAS_RAY = False
```
bouthilx, line 19:
```suggestion
HAS_RAY = not import_optional.failed
```

bouthilx:
There would need to be a documentation page here: https://github.com/Epistimio/orion/tree/develop/docs/src/code/executor.

bouthilx:
And a short section about Ray here: https://github.com/Epistimio/orion/blob/develop/docs/src/user/parallel.rst

src/orion/executor/ray_backend.py
Simnol22: Yes it returns them again. Just tested it.

PR #1050, opened 2022-12-20
Use list instead of array

bouthilx:
Numpy does not seam to support anymore non homogeneous arrays (for shapes at least). We don't need an array, so we should rather use lists.


Delaunay (approved):

tests/unittests/algo/test_tpe.py
```diff
@@ -403,7 +403,7 @@ def test_get_loglikelis(self, tpe: TPE):
 
 
 def _trial_to_array(trial: Trial, space: Space) -> np.ndarray:
```
lebrice, line 405: The typing would be wrong.
```suggestion
def _trial_to_array(trial: Trial, space: Space) -> list[tuple]:
```

lebrice (changes requested):

lebrice (approved):

Delaunay:
The issue here is dependencies using the deprecated np.object.
orion phased it out recently but dependencies are still using it and the object was removed in the latest numpy release
we will have to downgrade numpy on our side to keep the compatibility

PR #1067, opened 2023-01-18
Allow to spawn workers inside daemon

Delaunay:


src/orion/executor/multiprocess_backend.py
```diff
@@ -167,13 +167,18 @@ def __init__(self, n_workers=-1, backend="multiprocess", **kwargs):
         if n_workers <= 0:
             n_workers = multiprocessing.cpu_count()
 
+        self.pool_config = {"n_workers": n_workers, "backend": backend}
         self.pool = PoolExecutor.BACKENDS.get(backend, ThreadPool)(n_workers)
 
     def __setstate__(self, state):
-        self.pool = state["pool"]
+        log.warning("Nesting multiprocess executor")
```
bouthilx, line 174: So the workaround is to pass the config instead of the pool and recreate a pool with same config inside of it?

src/orion/executor/multiprocess_backend.py
```diff
@@ -167,13 +167,18 @@ def __init__(self, n_workers=-1, backend="multiprocess", **kwargs):
...
+        self.pool_config = state["pool_config"]
+        backend = self.pool_config.get("backend", ThreadPool)
```
bouthilx, line 176: Should this be 'thread' here instead?

src/orion/executor/multiprocess_backend.py
bouthilx: Unless I am getting something wrong, it seems to me we are increasing the number of workers exponentially by handling nested submits this way. Say I pass the executor to a function to be executed by submit, then `n` workers would be created inside this subprocess, in addition to the already existing `n` workers.

tests/unittests/executor/test_executor.py
```diff
@@ -246,17 +247,38 @@ def test_nested_submit(backend):
...
-            exception = NotImplementedError
-        elif backend == thread:
-            exception = TypeError
+        for r in results:
+            assert r.value == 27
 
-        with pytest.raises(exception):
-            [executor.submit(nested, executor) for i in range(5)]
+
+@pytest.mark.parametrize("backend", [multiprocess, thread])
```
bouthilx, line 278: How is this one different than `test_nested_submit`?

src/orion/executor/multiprocess_backend.py
Delaunay: actually we do not need the default the line after it will take care of the default, 
I updated the code

src/orion/executor/multiprocess_backend.py
Delaunay: No the workaround is `return _Process(*args, **kwds)`, `_Process` being a process that cannot be a daemon.
The daemon check is done inside the worker code by an assert, to bypass the assert the worker process should not be a daemon.

tests/unittests/executor/test_executor.py
Delaunay: ah right, it used to be different because of the __setstate__ I removed it

tests/unittests/executor/test_executor.py
```diff
@@ -280,3 +300,54 @@ def test_executors_del_does_not_raise(backend):
...
+
+    loader = torch.utils.data.DataLoader(dataset, num_workers=2, batch_size=64)
+
+    for i, _ in enumerate(loader):
+        pass
+
+    return i
+
+
+def has_pytorch():
```
bouthilx, line 329: :raised_eyebrow:

src/orion/executor/multiprocess_backend.py
```diff
@@ -167,13 +167,18 @@ def __init__(self, n_workers=-1, backend="multiprocess", **kwargs):
...
+        self.pool_config = {"n_workers": n_workers, "backend": backend}
         self.pool = PoolExecutor.BACKENDS.get(backend, ThreadPool)(n_workers)
 
     def __setstate__(self, state):
-        self.pool = state["pool"]
+        log.warning("Nesting multiprocess executor")
+        self.pool_config = state["pool_config"]
+        backend = self.pool_config.get("backend")
+        n_workers = self.pool_config.get("n_workers", -1)
+        self.pool = PoolExecutor.BACKENDS.get(backend, ThreadPool)(n_workers)
```
bouthilx, line 178: I'm a bit unsure about this part. If the object is serialized and passed to the subprocess, the deserialization step will have the effect or creating another pool of n_workers, no?

Delaunay: yea

Delaunay: we maybe able to pass a queue instead to avoid creating multiple pools
but nesting the executor in general is a bit of a nono

PR #1072, opened 2023-01-23
Reduce Warnings

Delaunay:


Delaunay:
#1044

tox.ini
```diff
@@ -261,8 +261,6 @@ commands =
 addopts = -ra -q --color=yes
 norecursedirs = .* *.egg* config docs dist build
 xfail_strict = True
-rsyncdirs = src tests
```
bouthilx, line 264: What were these for?

bouthilx (approved):
LGTM, Thanks!

tox.ini
Delaunay: I don't know, but they trigger deprecation warnings in pytest and the messages imply the options were removed

bouthilx: Seems to be a long deprecated argument of pytest-xdist which we are not even using. :thinking: Alright, then.

PR #1082, opened 2023-02-10
Add storage import and export features to command line and web API

### LINKED ISSUE #953: Support storage checkpointing and restore
Users need a way to checkpoint their current storage and restore it. This would also allow migrating from one database backend to another, for example checkpointing from pickleddb, and then loading the checkpoint into mongodb. The easiest way to support this would be to save the checkpoint into a pickleddb. The restoration would be the trickiest part, as we need to ensure there are no collisions, and that the id references between experiments and trials are conserved even though the ids will need to change to avoid collisions with existing data in the storage.

`orion db dump --config orion_config.yaml --exp experiment`
`orion db load --config orion_config.yaml --exp experiment dump.pkl`

`--exp` would be an optional argument to select only a subset of the data.

notoraptor:
# Description

HI @bouthilx ! This is a pull request to add export (dump) and import (load) features to Oríon, currently available via new command lines and new web API entries. New features use storage objects instead of underlying databases.

This should fix #953 

**NB: important remarks:**
- I had to add few modifications to `BaseStorageProtocol`
- Overall, import/export may be slower when using storage instead of database when we import/export whole data. That's because we must call storage method for each data (benchmark or experiment) we want to manage, making a call to database for each manipulation. While, with database object, we could group queries and make one call to write/read many data at once.

# Changes
- Add a new module `storage_backup` that provides 2 functions, `dump_database` for export, and `load_database` for import.
- Add new command lines using these functions, `orion db dump` and `orion db load`
- Add new web API entries, `/dump` for export, `/load` for import, and `/import-status` to monitor import.

# Checklist

## Tests

## Documentation

## Quality


docs/src/user/storage.rst
```diff
@@ -153,6 +153,46 @@ simply run the upgrade command.
...
+
+.. code-block:: sh
+
+   orion db dump -n exp-name -v exp-version -o backup-exp.pkl
+
+``load`` Import database content
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``load`` command allows to import database content
+from a PickledDB PKL file generated by ``dump`` command.
```
bouthilx, line 175:
```suggestion
from any PickledDB PKL file (including files generated by ``dump`` command).
```

src/orion/core/cli/db/load.py
```diff
@@ -0,0 +1,63 @@
...
+        type=str,
+        help="File to import",
+    )
+
+    load_parser.add_argument(
+        "-r",
+        "--resolve",
+        type=str,
+        choices=("ignore", "overwrite", "bump"),
+        required=True,
```
bouthilx, line 40: If there are no conflicts it would be great if the command could work without specifying the resolution method. Did we ever discuss about this? I only have vague memories about it... :thinking:

src/orion/core/cli/db/load.py
```diff
@@ -0,0 +1,63 @@
...
+
+    load_parser.set_defaults(func=main)
+
+    return load_parser
+
+
+def main(args):
+    """Script to import storage"""
+    config = experiment_builder.get_cmd_config(args)
+    storage = setup_storage(config.get("storage"))
```
bouthilx, line 55: There should be a verification here that the storage is a Legacy storage, and ideally also that it is pickleddb. Otherwise the `logging.info` below will fail.

bouthilx: Oh wait, that's the destination storage. Not the source one. Then the `main` will fail here if the destination is not Legacy. Should only print `storage` instead of `storage._db`.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+
+    # For pickled databases, make sure src is not dst
+    if hasattr(storage, "_db"):
+        orig_db = storage._db
+        if isinstance(orig_db, PickledDB) and dump_host == os.path.abspath(
+            orig_db.host
+        ):
+            raise DatabaseError("Cannot dump pickleddb to itself.")
+
+    dst_storage = setup_storage({"database": {"host": dump_host, "type": "pickleddb"}})
```
bouthilx, line 66: What do we do if the destination storage already exists? I would be in favor of raising an error if it does, and adding an option to force overwrite. If forced overwrite, then the dump should be deleted and recreated from scratch to avoid merging storages using `dump`.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+    # Dump experiment and algo
+    dst_storage.create_experiment(
+        src_exp,
+        algo_locked=algo_lock_info.locked,
+        algo_state=algo_lock_info.state,
+        algo_heartbeat=algo_lock_info.heartbeat,
+    )
+    logger.info("\tCreated exp")
+    # Dump trials
+    for trial in src_storage.fetch_trials(uid=src_exp["_id"]):
```
bouthilx, line 204: Is it possible that the experiment id changes (if `create_experiment` assigns a new one) and then the trials are pointing to the wrong experiment id? We should be able to test this easily if we dump an experiment with id > 1 as it would have id == 1 in the `dst_storage` if the id is not kept in `create_experiment`.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+            query["version"] = version
+        experiments = src_storage.fetch_experiments(query)
+        if not experiments:
+            raise DatabaseError(
+                f"No experiment found with query {query}. Nothing to dump."
+            )
+        if len(experiments) > 1:
+            exp_data = sorted(experiments, key=lambda d: d["version"])[0]
+        else:
+            (exp_data,) = experiments
```
bouthilx, line 184:
```suggestion
exp_data = sorted(experiments, key=lambda d: d["version"])[0]
```
Simpler IMHO than a non-significant optimization for n==1.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+        query = {"name": name}
+        if version is not None:
+            query["version"] = version
+        experiments = src_storage.fetch_experiments(query)
+        if not experiments:
+            raise DatabaseError(
+                f"No experiment found with query {query}. Nothing to import."
+            )
+        if len(experiments) > 1:
+            experiments = sorted(experiments, key=lambda d: d["version"])[:1]
```
bouthilx, line 128:
```suggestion
            experiments = sorted(experiments, key=lambda d: d["version"])[-2:]
```
Should use latest version by default.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+    resolve:
+        policy to resolve import conflict. Either 'ignore', 'overwrite' or 'bump'.
+        - 'ignore' will ignore imported data on conflict
+        - 'overwrite' will overwrite old data in destination database on conflict
+        - 'bump' will bump imported data version before adding it,
+          if data with same ID is found in destination
+    name:
+        (optional) name of experiment to import (by default, whole file is imported)
+    version:
+        (optional) version of experiment to import
```
bouthilx, line 92:
```suggestion
        (optional) version of experiment to import. Otherwise use latest version of provided `name`.
```

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+    Parameters
+    ----------
+    storage: BaseStorageProtocol
+        storage of database to dump
+    dump_host:
+        file path to dump into (dumped file will be a pickled file)
+    name:
+        (optional) name of experiment to dump (by default, full database is dumped)
+    version:
+        (optional) version of experiment to dump
```
bouthilx, line 54:
```suggestion
        (optional) version of experiment to dump. Otherwise use latest version of provided `name`. 
```

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+            len(src_benchmarks),
+            len(src_benchmarks),
+            progress_callback,
+        )
+
+    _describe_import_progress(STEP_CHECK_DST_EXPERIMENTS, 0, 1, progress_callback)
+    all_dst_experiments = dst_storage.fetch_experiments({})
+    # Dictionary mapping dst exp name to exp version to list of exps with same name and version
+    dst_exp_map = {}
+    last_experiment_id = max((data["_id"] for data in all_dst_experiments), default=0)
```
bouthilx, line 288: The `_id`s may not all be integers. It won't be the case if the `dst_storage` is mongodb for example.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+                for collection in EXPERIMENT_RELATED_COLLECTIONS:
+                    queries_to_delete.setdefault(collection, []).append(
+                        {"experiment": dst_experiment["_id"]}
+                    )
+                queries_to_delete.setdefault(COL_EXPERIMENTS, []).append(
+                    {"_id": dst_experiment["_id"]}
+                )
+            elif resolve == "bump":
+                old_version = experiment["version"]
+                new_version = last_versions.get(experiment["name"], 0) + 1
```
bouthilx, line 327: This will not work for all storage backends. The new ID should be inferred after calling `storage.create_experiment` as it is the storage backend that will generate it. That unfortunately means that we will need deferring the trial.experiment assignment to `_execute_import`.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,473 @@
...
+
+    for new_benchmark in data_to_add.get(COL_BENCHMARKS, ()):
+        dst_storage.create_benchmark(new_benchmark)
+        _describe_import_progress(
+            STEP_INSERT_NEW_DATA, i_data, nb_data_to_add, progress_callback
+        )
+        i_data += 1
+
+    for new_experiment in data_to_add.get(COL_EXPERIMENTS, ()):
+        new_algo = data_to_add[COL_ALGOS][new_experiment["_id"]]
```
bouthilx, line 448: So, in order to handle any type of IDs, I think a solution would be to generate a random hashing for the new experiment id as a common reference for the algo and trial documents (during `_prepare_import`). Then here, you could fetch the algo and trial documents just as you do here, but then pop the `_id` out of `new_experiments`, call `create_experiment` and then reassign the freshly created `new_experiment['_id']` by the storage backend to algo and trials documents.

src/orion/core/cli/db/load.py
notoraptor: Done!

src/orion/core/cli/db/load.py
notoraptor: Ok!

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+from queue import Empty
+
+import falcon
+from falcon import Request, Response
+
+from orion.core.io.database import DatabaseError
+from orion.core.worker.storage_backup import dump_database, load_database
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
bouthilx, line 24: This should be perhaps configurable through the webapi config. Actually, I think using `orion -vv serve [OPTIONS]` would correctly set this logger to DEBUG.

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def _gen_host_file(basename="dump"):
```
bouthilx, line 27: Maybe using [`tempfile.NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) would require less work.

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+            self.queue.put(line)
+
+    def flush(self):
+        """Placeholder to flush data"""
+
+
+class ImportTask:
+    """Wrapper to represent an import task. Used to monitor task progress.
+
+    Properties:
```
bouthilx, line 66: Should be formatted as an Attribute section in numpy docstring style.

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def _import_data(task: ImportTask, storage, load_host, resolve, name, version):
+    """Function to run import task.
+
+    Set stream handler, launch load_database and set task status.
+    """
+    try:
+        print("Import starting.", task.task_id)
+        logging.basicConfig(stream=task.notifications, force=True, level=logging.INFO)
```
bouthilx, line 110: Why do we need to set the logging level here? Is it because we lose it when running `_import_data` in a child process?

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+    """Handle requests for the dump/ REST endpoint"""
+
+    def __init__(self, storage):
+        self.storage = storage
+        self.current_task: ImportTask = None
+
+    def on_get_dump(self, req: Request, resp: Response):
+        """Handle the GET requests for dump/"""
+        name = req.get_param("name")
+        version = req.get_param_as_int("version")
```
bouthilx, line 145: Does in handle `None`?

src/orion/serving/storage_resource.py
```diff
@@ -0,0 +1,232 @@
...
+        except DatabaseError as exc:
+            raise falcon.HTTPNotFound(title=type(exc).__name__, description=str(exc))
+        finally:
+            os.unlink(dump_host)
+            os.unlink(f"{dump_host}.lock")
+
+    def on_post_load(self, req: Request, resp: Response):
+        """Handle the POST requests for load/"""
+        if self.current_task is not None and not self.current_task.is_completed():
+            raise falcon.HTTPForbidden(description="An import is already running")
```
bouthilx, line 165: We would perhaps add an option to allow cancelling an import. If it's too much work you could create an issue instead and delay this work.

tests/functional/commands/conftest.py
```diff
@@ -441,3 +447,111 @@ def three_experiments_same_name_with_trials(
         orionstate.database.write("trials", trial2.to_dict())
         orionstate.database.write("trials", trial3.to_dict())
         x_value += 1
+
+
+@pytest.fixture
+def three_experiments_family_same_name_with_trials(
```
bouthilx, line 453: To be coherent with other tests, that's the `three_experiments_branch_same_name`.

bouthilx: Maybe `benchmark` should be in the name of the test as well since this fixture is the only one adding adding a benchmark to the DB.

bouthilx:
The `*.pickled` files are `PickledDB` databases right? These should not be included in the repo and rather be generated automatically with the test fixtures.

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,142 @@
...
+        execute("db dump")
+        assert os.path.isfile("dump.pkl")
+        dumped_db = PickledDB("dump.pkl")
+        with dumped_db.locked_database(write=False) as internal_db:
+            collections = set(internal_db._db.keys())
+        assert collections == {"experiments", "algo", "trials", "benchmarks"}
+        assert len(dumped_db.read("experiments")) == 3
+        assert len(dumped_db.read("algo")) == 3
+        assert len(dumped_db.read("trials")) == 24
+        assert len(dumped_db.read("benchmarks")) == 3
```
bouthilx, line 38: I think this test is too superficial. It should verify that the links between collections are not broken. That is, algo and trials referencing experiment ids, and experiment referencing parent ids (in field `refers`).

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,142 @@
...
+    """Test dump unknown experiment"""
+    assert not os.path.exists("dump.pkl")
+    try:
+        execute("db dump -n i-dont-exist", assert_code=1)
+        captured = capsys.readouterr()
+        assert captured.err.startswith(
+            "Error: No experiment found with query {'name': 'i-dont-exist'}. Nothing to dump."
+        )
+    finally:
+        clean_dump("dump.pkl")
```
bouthilx, line 55: Should `dump.pkl` exist if the command failed?

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,142 @@
...
+        with dumped_db.locked_database(write=False) as internal_db:
+            collections = set(internal_db._db.keys())
+        assert collections == {"experiments", "algo", "trials", "benchmarks"}
+        assert len(dumped_db.read("experiments")) == 3
+        assert len(dumped_db.read("algo")) == 3
+        assert len(dumped_db.read("trials")) == 24
+        assert len(dumped_db.read("benchmarks")) == 3
+    finally:
+        clean_dump(dump_path)
+
```
bouthilx, line 77: Should test as well dumping to an already existing file path.

tests/functional/commands/test_db_dump.py
bouthilx: It should also look for trial.parent. Right now it should not matter because the trial hash does not depend on the experiment id. But if we ever make changes that affects this we would like this test to fail so that we fix the dump script accordingly.

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_dump_one_experiment(three_experiments_family_same_name_with_trials, capsys):
```
bouthilx, line 79: I think it may be possible to combine these 2 tests and use parametrization instead.

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+import orion.core.cli
+from orion.storage.base import setup_storage
+
+LOAD_DATA = os.path.join(os.path.dirname(__file__), "orion_db_load_test_data.pickled")
+LOAD_DATA_WITH_BENCHMARKS = os.path.join(
+    os.path.dirname(__file__), "orion_db_load_test_data_with_benchmarks.pickled"
+)
+
+
+# TODO: Test loading benchmarks
```
bouthilx, line 15: It's done right?

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+
+    execute(f"db load {LOAD_DATA_WITH_BENCHMARKS}")
+    with loaded_db.locked_database(write=False) as internal_db:
+        collections = set(internal_db._db.keys())
+    assert collections == {"experiments", "algo", "trials", "benchmarks"}
+
+    assert len(loaded_db.read("benchmarks")) == 3
+    assert len(loaded_db.read("experiments")) == 3
+    assert len(loaded_db.read("trials")) == 24
+    # TODO: We should expect 6 algorithms, but only 3 are returned
```
bouthilx, line 62: I don't understand. The `pickle` file contains 6 algo locks, but only 3 experiments? This should not be possible. How was the file generated?

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+    assert len(loaded_db.read("experiments")) == 0
+    assert len(loaded_db.read("trials")) == 0
+    assert len(loaded_db.read("algo")) == 0
+
+    execute(f"db load {LOAD_DATA_WITH_BENCHMARKS}")
+    with loaded_db.locked_database(write=False) as internal_db:
+        collections = set(internal_db._db.keys())
+    assert collections == {"experiments", "algo", "trials", "benchmarks"}
+
+    assert len(loaded_db.read("benchmarks")) == 3
```
bouthilx, line 59: Same thing here for testing the links between experiments, trials and algo. This could be a generic function, taking `source_storage` and `target_storage` as argument.

src/orion/core/worker/storage_backup.py
notoraptor: Ok !

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+    experiments = loaded_db.read("experiments")
+    trials = loaded_db.read("trials")
+    algos = loaded_db.read("algo")
+    assert len(benchmarks) == 3
+    assert len(experiments) == 3
+    assert len(trials) == 24
+    assert len(algos) == 3
+
+    execute(f"db load {LOAD_DATA_WITH_BENCHMARKS} -r ignore")
+    # Duplicated data should be ignored, so we must expect same number of data and same IDs.
```
bouthilx, line 127: As these tests stand, it is not possible to distinguish the effect of no resolution, ignore or overwrite. At least one document of each collection should be modified so that we can verify that the modified version was not used in case of ignore, or was used in case of overwrite.
You could modify harmless fields like `experiment.algorithm`, `trial.status` and `algo.hearthbeat` for these tests. These would have no impact of the IDs.

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+    experiments = loaded_db.read("experiments")
+    trials = loaded_db.read("trials")
+    algos = loaded_db.read("algo")
+    assert len(benchmarks) == 0
+    assert len(experiments) == 3
+    assert len(trials) == 24
+    assert len(algos) == 3
+
+    execute(f"db load {data_source} -r bump")
+    # Duplicated data should be bumped, so we must expect twice quantity of data.
```
bouthilx, line 255: Here it is particularly important that we test for the links since we changed the base references (experiment id).

src/orion/core/worker/storage_backup.py
notoraptor: Ok !

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,462 @@
...
+
+def test_load_all(empty_database):
+    """Test load all database"""
+    assert os.path.isfile(LOAD_DATA_WITH_BENCHMARKS)
+    storage = setup_storage()
+    loaded_db = storage._db
+    assert len(loaded_db.read("benchmarks")) == 0
+    assert len(loaded_db.read("experiments")) == 0
+    assert len(loaded_db.read("trials")) == 0
+    assert len(loaded_db.read("algo")) == 0
```
bouthilx, line 52: This part could be extracted as a fixture, reused by all tests. They could take `loaded_db` instead of `empty_database` as argument.

src/orion/core/worker/storage_backup.py
notoraptor: Ok!

src/orion/core/worker/storage_backup.py
notoraptor: Ok!

tests/functional/serving/test_storage_resource.py
```diff
@@ -0,0 +1,436 @@
...
+            os.unlink(path)
+
+
+def _random_string(length: int):
+    """Generate a random string with given length. Used to generate multipart form data for /load request"""
+    domain = string.ascii_lowercase + string.digits
+    return "".join(random.choice(domain) for _ in range(length))
+
+
+def _gen_multipart_form_form_load(file: str, resolve: str, name="", version=""):
```
bouthilx, line 47:
```suggestion
def _gen_multipart_from_form_load(file: str, resolve: str, name="", version=""):
```
?

tests/functional/serving/test_storage_resource.py
```diff
@@ -0,0 +1,436 @@
...
+    _load_data(ephemeral_storage)
+    response = client.simulate_get("/dump")
+    host = _gen_host_file()
+    try:
+        with open(host, "wb") as file:
+            file.write(response.content)
+        dumped_db = PickledDB(host)
+        with dumped_db.locked_database(write=False) as internal_db:
+            collections = set(internal_db._db.keys())
+        assert collections == {"experiments", "algo", "trials", "benchmarks"}
```
bouthilx, line 118: This again could reuse the generic tests looking at links in data.

tests/functional/serving/test_storage_resource.py
```diff
@@ -0,0 +1,436 @@
...
+        assert len(dumped_db.read("experiments")) == 3
+        assert len(dumped_db.read("algo")) == 3
+        assert len(dumped_db.read("trials")) == 24
+        assert len(dumped_db.read("benchmarks")) == 0
+    finally:
+        _clean_dump(host)
+
+
+def test_dump_one_experiment(client, ephemeral_storage):
```
bouthilx, line 127: Since the interface is very similar it may be possible to refactor the tests so that they reuse the same code blocks for data validation.

src/orion/serving/storage_resource.py
notoraptor: Yes! Logging level is not shared across processes by default, because it does not use shared objects (like `multiprocessing.Queue`). So, we need to set logging level each time a process starts. In this line we also set stream as a shared object here, and then collect logging from process and access it in another process.

src/orion/serving/storage_resource.py
notoraptor: Ok!

src/orion/serving/storage_resource.py
notoraptor: Yes ! By default, it returns None if parameter is not found and not required: https://falcon.readthedocs.io/en/2.0.0/api/request_and_response.html#falcon.Request.get_param_as_int

src/orion/serving/storage_resource.py
notoraptor: Ok ! I opened an issue: https://github.com/Epistimio/orion/issues/1085

tests/functional/serving/test_storage_resource.py
notoraptor: Fixed -> `_gen_multipart_form_for_load`

notoraptor:
> The `*.pickled` files are `PickledDB` databases right? These should not be included in the repo and rather be generated automatically with the test fixtures.

Hi @bouthilx ! Indeed, the `.pickled` files are PickledDB database files. I use them as src data to test storage import. If I use test fixtures instead, does it mean I will have to regenerate a PKL file for each storage import test ? In such cases, I would need to dump to the temporary PKL file before, then load, for each test. Won't it be time and resource consuming ?

tests/functional/commands/conftest.py
notoraptor: Ok!

tests/functional/commands/test_db_dump.py
notoraptor: Yes! Output file is created as soon as the destination storage object is created in `dump_database()` (PickledDB generates the PKL file immediately), and storage object is allocated before checking if experiment exists or not. So, output file will exist in any case.

src/orion/serving/storage_resource.py
notoraptor: Indeed. Plus, I just realized that `logger` is not used in this module, so I removed it.

tests/functional/commands/test_db_dump.py
notoraptor: Done in `test_db_dump` ! To do in `test_db_load`.

tests/functional/commands/test_db_load.py
notoraptor: Indeed!

tests/functional/commands/test_db_dump.py
notoraptor: Done!

tests/functional/commands/test_db_load.py
notoraptor: There are generic fonctions now!

tests/functional/commands/test_db_load.py
notoraptor: Done!

tests/functional/serving/test_storage_resource.py
notoraptor: Done!

tests/functional/serving/test_storage_resource.py
notoraptor: Done!

tests/functional/commands/test_db_load.py
notoraptor: Tested!

tests/functional/commands/test_db_load.py
bouthilx: It's likely because we update the experiment `_id` here: https://github.com/notoraptor/orion/blob/import-export-cli-and-webapi/tests/functional/commands/conftest.py#L51. And we do not fix this id in the algo lock. It would need to be fixed.

src/orion/core/worker/storage_backup.py
notoraptor: Ok ! Code does not rely on ID as integer anymore.

src/orion/serving/storage_resource.py
bouthilx: It may be good to pass the logging level so it can be the same as the global one in the main process.

tests/functional/commands/test_db_dump.py
bouthilx: We should try to delete it whenever possible when the command fails (as part of the execution of `db dump`, not in the test cleanup phase). It's confusing otherwise.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,715 @@
...
+        storage,
+        experiments,
+        resolve,
+        import_benchmarks,
+        progress_callback=progress_callback,
+    )
+    _execute_import(storage, *preparation, progress_callback=progress_callback)
+
+
+class _Graph:
```
bouthilx, line 170: I think you could reuse `orion.core.evc.tree` instead.

tests/conftest.py
```diff
@@ -9,6 +9,19 @@
 import numpy
 import pytest
 import yaml
+
+# Import pkl_* fixtures for tests in test_storage_resource
+# Need to also import parent fixtures to make fixtures work
+from functional.commands.conftest import (
```
bouthilx, line 15: I don't understand why you need to import these here. They should only be available in the scope of `tests/functional/commands`. This makes them globally available.

notoraptor: Ok ! Imports moved into `tests/functional/conftest` instead.

notoraptor:
@bouthilx About your comment here:

> Right, it's using EphemeralDB by default. It would be best to dump it as part of the fixtures for these tests yes. Each call to pickleddb leads to a full dump/load of the DB so it should not affect performance too much.

Already done! I have completed removed PKL files from Git repository. They are now generated at runtime when necessary.

notoraptor:
Trials links also checked. Now todo:
- root ID links
- Use existent tree module instead of newly added `_Graph` class.

src/orion/core/worker/storage_backup.py
notoraptor: Reused!

src/orion/core/worker/storage_backup.py
bouthilx: Oh, that's for a version, not an ID.

tests/functional/commands/test_db_load.py
notoraptor: Done: https://github.com/Epistimio/orion/pull/1082/commits/ebfece3f477c509bab501736937b1e8e816f9a72

notoraptor:
Hi @bouthilx ! I added a supplementary commit to make sure main root log level is transferred into import sub-process in web API. There are also unit tests. Let's see if they pass on CI.

Commit here: https://github.com/Epistimio/orion/pull/1082/commits/80462924b2d17d4f16e8258476116a427f3563ca

src/orion/core/cli/db/load.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
+# pylint: disable=,protected-access
```
bouthilx:
```suggestion
```
Does not look like there are any protected access anymore.

src/orion/core/worker/storage_backup.py
```diff
@@ -0,0 +1,725 @@
+# pylint: disable=,protected-access,too-many-locals,too-many-branches,too-many-statements
+"""
```
bouthilx: This module should be in `orion.storage` instead.

tests/functional/commands/conftest.py
```diff
@@ -441,3 +465,424 @@ def three_experiments_same_name_with_trials(
...
+            loaded_db, nb_exps=0, nb_algos=0, nb_trials=0, nb_benchmarks=0
+        )
+
+    @staticmethod
+    def assert_tested_db_structure(dumped_db, nb_orig_benchmarks=3, nb_duplicated=1):
+        """Check counts and experiments for database from specific fixture
+        `three_experiments_branch_same_name_trials[_benchmarks]`.
+        """
+        # NB: about nb_algos:
+        # TODO: We should expect 6 algorithms, but only 3 are returned
```
bouthilx: Is this fixed now?

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,150 @@
...
+        testing_helpers.assert_tested_db_structure(dumped_db)
+    finally:
+        clean_dump(dump_path)
+
+
+def test_dump_unknown_experiment(
+    three_experiments_branch_same_name_trials_benchmarks, capsys
+):
+    """Test dump unknown experiment"""
+    assert not os.path.exists("dump.pkl")
```
bouthilx: Most tests should use `tmp_path / 'dump.pkl'` to avoid conflicts between them.

tests/functional/commands/test_db_load.py
```diff
@@ -0,0 +1,340 @@
...
+
+import os
+
+import pytest
+
+import orion.core.cli
+from orion.core.io.database.pickleddb import PickledDB
+from orion.storage.base import setup_storage
+
+# TODO: Test trial parents links and experiment root links in loaded data
```
bouthilx: This is done now right?

tests/functional/commands/test_db_dump.py
notoraptor: Done! https://github.com/Epistimio/orion/pull/1082/commits/8cd1528cf5d450f42f8c1a235addc7f4fef2aa85

src/orion/core/cli/db/load.py
notoraptor: Done: https://github.com/Epistimio/orion/pull/1082/commits/3321d22cf30d27572845c50c8bb6a278847a8999

src/orion/core/worker/storage_backup.py
notoraptor: Done: https://github.com/Epistimio/orion/pull/1082/commits/5fe487b6be09904df125fe9e16f54d337ef7a3ab

tests/functional/commands/conftest.py
notoraptor: Yes, fixed here: https://github.com/Epistimio/orion/pull/1082/commits/ebfece3f477c509bab501736937b1e8e816f9a72

I removed this TODO: https://github.com/Epistimio/orion/pull/1082/commits/1a3928794495ba24f5939795bb3d9279cf5f78c5

tests/functional/commands/test_db_load.py
notoraptor: Indeed, TODO removed: https://github.com/Epistimio/orion/pull/1082/commits/73ff556367863c88efafd942fd744f3797b3e42a

notoraptor:
@bouthilx Supplementary commit to fix tests: https://github.com/Epistimio/orion/pull/1082/commits/a8406e7ffb7572840b5e72a3cb8bcc1932af02f3

tests/functional/commands/test_db_dump.py
notoraptor: Some tests need to not specify an output file, to verify that everything goes well with default output file `dump.pkl`. How should I manage that ?

src/orion/storage/backup.py
```diff
@@ -0,0 +1,734 @@
...
+                f"Export output already exists (specify `--force` to overwrite) at {dump_host}"
+            )
+
+    try:
+        dst_storage = setup_storage(
+            {"database": {"host": dump_host, "type": "pickleddb"}}
+        )
+        logger.info(f"Dump to {dump_host}")
+        _dump(storage, dst_storage, name, version)
+    except Exception as exc:
```
bouthilx, line 97: If the path already exists and --force is used, this would have the effect of deleting the target db. I think the safest way would be to work on a tmp file-path and then move it to the target db path only if the process was successful. This case should be tested as well (db exists and --force is used but the process crash, then the db should have been left intact).

tests/functional/commands/test_db_dump.py
bouthilx: Indeed, we need at least one test for the default output. I think the others should use `tmp_path/'dump.pkl'` so at least only the test for default output will fail if we are not cleaning up the tests properly.

notoraptor: Done here! https://github.com/Epistimio/orion/pull/1082/commits/e6b23fc06a28303219ba887bd9b5c83abecd5541

One more commit is coming for latest commentary.

src/orion/storage/backup.py
notoraptor: Done ! https://github.com/Epistimio/orion/pull/1082/commits/922744d70f10152a03a8a78cacee04421d260ec2

All comments addressed!

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,262 @@
...
+        testing_helpers.assert_tested_db_structure(dumped_db)
+    finally:
+        clean_dump("dump.pkl")
+
+
+def test_dump_overwrite(
+    three_experiments_branch_same_name_trials_benchmarks, capsys, testing_helpers
+):
+    """Test dump with overwrite argument"""
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
bouthilx: You can avoid this line by using the fixture `tmp_path`: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,262 @@
...
+            testing_helpers.assert_tested_db_structure(dumped_db)
+        finally:
+            clean_dump(dump_path)
+
+
+def test_dump_to_specified_output(
+    three_experiments_branch_same_name_trials_benchmarks, capsys, testing_helpers
+):
+    """Test dump to a specified output file"""
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
bouthilx: Same here

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,262 @@
...
+            "Nothing to dump.",
+        ),
+    ],
+)
+def test_dump_post_clean_on_error(
+    output_already_exists, output_specified, overwrite, error_message, capsys
+):
+    """Test how dumped file is cleaned if dump fails."""
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
bouthilx: idem

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,262 @@
...
+            # Output should not exist after error.
+            assert not os.path.exists(expected_output)
+            assert not os.path.exists(f"{expected_output}.lock")
+
+
+def test_dump_unknown_experiment(
+    three_experiments_branch_same_name_trials_benchmarks, capsys
+):
+    """Test dump unknown experiment"""
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
bouthilx: idem

tests/functional/commands/test_db_dump.py
```diff
@@ -0,0 +1,262 @@
...
+    three_experiments_branch_same_name_trials_benchmarks,
+    testing_helpers,
+    given_version,
+    expected_version,
+    nb_trials,
+    nb_child_trials,
+    algo_state,
+):
+    """Test dump experiment test_single_exp"""
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
bouthilx: idem

tests/functional/commands/test_db_dump.py
notoraptor: Done (here and for all comments below): https://github.com/Epistimio/orion/pull/1082/commits/91a10d9c2d70ebc27f73a2b8ba5a1ecfcf4634ea

PR #1094, opened 2023-03-17
Handle cmdline config with status cmd

bouthilx:
The storage is not a singleton anymore and must be passed down in functions if they all need to use the same storage. The status command was not adjusted properly when the whole code base was refactored to remove singletons.

Now uses ExperimentBuilder always with the same storage.


notoraptor (approved):
Looks good to me ! I hope Dashboard Playwright tests won't fail here ...

PR #1096, opened 2023-04-17
Fix dashboard tests

notoraptor:
# Description

Try to fix dashboard tests.

# Changes

- Update playwright
- Handle new pandas version 2.0.0

# Checklist

## Tests

## Documentation

## Quality

### DIFF
--- a/.github/workflows/dashboard-src-new-with-playwright.yml
+++ b/.github/workflows/dashboard-src-new-with-playwright.yml
@@ -28,9 +28,12 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
+    - name: Setup kernel for react native, increase watchers
+      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
     - name: Check ulimit
       run: ulimit -a
-    - run: sudo ulimit -i 126832
+    - run: ulimit -i 126832
     - run: ulimit -l 65536
     - run: ulimit -u 126832
     - run: ulimit -a
--- a/.github/workflows/dashboard-src-new-with-playwright.yml
+++ b/.github/workflows/dashboard-src-new-with-playwright.yml
@@ -41,17 +41,7 @@ jobs:
       uses: supercharge/mongodb-github-action@1.8.0
       with:
         mongodb-version: 6.0
-
-    - name: Install MongoShell
-      run: |
-        wget -qO - https://www.mongodb.org/static/pgp/server-6.0.asc | sudo apt-key add -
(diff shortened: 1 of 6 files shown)


notoraptor:
Hi @bouthilx ! It seems I have fixed the issue related to failing dashboard tests on CI!

Ultimately, the solution was to handle both new and old versions of pandas by specify parameter `group_keys=False` when using `groupby()`: https://github.com/Epistimio/orion/pull/1096/files#diff-ad4314cde86cb004bc4eeacc08ea818db0887a0e74888f4064ad384ddd9a48d7

Before:
```python
return trials.groupby(group_by).apply(rank)
```

After:
```python
return trials.groupby(group_by, group_keys=False).apply(rank)
```

`group_keys` argument is already available in older versions of pandas, but it seems pandas 2.0.0 changes its behavior. So, to keep old behavior, a warning from pandas 1.3.5 (version just before 2.0.0) suggests to use `group_keys=False`:

```
/home/stevenbocco/mila/dev/git/epistimio/orion/src/orion/analysis/base.py:107: FutureWarning:

Not prepending group keys to the result index of transform-like apply. In the future, the group keys will be included in the index, regardless of whether the applied function returns a like-indexed object.
To preserve the previous behavior, use

	>>> .groupby(..., group_keys=False)

To adopt the future behavior and silence this warning, use 

	>>> .groupby(..., group_keys=True)
```

This seems sufficient to fix tests, no need to change ulimit.

So, dashboard tests pass again! But now I have issues with build tests and long algos.

- For build tests, I don't yet know how to fix them: https://github.com/Epistimio/orion/actions/runs/4737274539/jobs/8409954043?pr=1096
- For long algos, errors occur on Python 3.7: https://github.com/Epistimio/orion/actions/runs/4737274539/jobs/8409954604?pr=1096

About long algo issues, it seems related to package `ax-platform` which uses `typeguard` with a new version that causes backward incompatibilities. I found this ticket: https://github.com/facebook/Ax/pull/1510

On Python 3.8, new ax package seems to have fixed the issue by forcing a typeguard version in dependencies: https://github.com/facebook/Ax/blob/0.3.1/setup.py#L31

But, on Python 3.7, latest supported ax package is 0.2.6, from August 2022, and I guess it is not updated and relies on `typeguard` by default, without hardcoded version: https://github.com/facebook/Ax/blob/0.2.6/setup.py#L29

So, I don't know what strategy to adopt. Should I try to "force" orion to install `typeguard=2.13.3` for Python <= 3.7 ?

bouthilx:
Python 3.7 end of life is scheduled for 2023-06-27. I think we can safely drop support for 3.7 in next release, so now in `develop`.

bouthilx:
The failed test in the build tests look like a glitch. I relaunched them.

bouthilx (approved):

PR #1108, opened 2023-07-12
Remove the profet import and warning

lebrice:
# Description
Removes the warning about needing the profet extra to run profet tasks when importing Orion.
Also removes the `profet` module from the `__all__` of `orion.benchmark.task`, it will have to be imported directly instead.

# Changes
`from orion.benchmark.task import *` will no longer include `profet` when the dependencies are installed.

An alternative would be to remove the warning and keep the import. However needing to load PyTorch when importing Orion is not great. It's probably better for it to stay a lazy import.


bouthilx (approved):
Thanks!

PR #1109, opened 2023-07-12
Fix Kaleido version to work around release error

NeilGirdhar:
Due to https://github.com/plotly/Kaleido/issues/156, Kaleido doesn't work with package managers like Poetry.  It also hasn't seen a release in 2 years and the package manager seems to have forgotten about it.  Could we pin the version to solve this problem?


bouthilx (approved):
Seems ok. We may need to find an alternative at some point. :(

PR #1110, opened 2023-07-13
Add py.typed as per PEP 561

NeilGirdhar:


NeilGirdhar:
Without this, MyPy complains that there are no stubs for orion, which is a shame because orion has plenty of type annotations.

PR #1121, opened 2023-11-03
Pin pre-commit to py3.11

bouthilx:
Some dependencies (namely autoflake) are not compatible with py3.12 yet.

Pinning pre-commit to py3.11 until we can move to py3.12. There is no reason for having a pre-commit on a later version than what orion supports anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3
+        with:
+          python-version: 3.11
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
You may want to consider moving from flake to ruff, which has become hugely popular.  It has very friendly maintainers, does almost everything flake does, is faster, and easier to configure. It also uses the exact same noqa commands, so it should be a drop in replacement.

bouthilx:
@NeilGirdhar Thanks for sharing, I created another PR based on my fork's branch: https://github.com/Epistimio/orion/pull/1124

PR #1136, opened 2025-11-10
Fix Github CI, versionning and unit tests to prepare a release for latest Python versions

notoraptor:
# Description

This PR is intended to:
- fix unit tests and Github CI to have a fully-working code
- prepare code for a future release compatible will latest Python versions, including 3.14
  - Make sure all tests pass with all Python versions we want to support
- Make sure versionning works

# Changes

## Fix, improve and extend Github / CI

- use pytest xdist parameter `-n 4` to speed up tests
- Add tests for Python 3.9 to Python 3.14
- Fix dashboard tests: update dependencies, and now run tests only on node 18.x. Older nodejs versions are not supported by new dependencies.

## Add some conditions in `setup.py`:

- extra `nevergrad`: `fcmaes` requires numba, not yet compatible with python 3.14. So, add a constrainst `python_version < "3.14"` to `fcmaes` only.
- extra `configspace`:
  - cannot support newest versions `1.+` because some dependencies (`sample-space`) still use a parameter `q` which was removed from `ConfigSpace 0.7` to `ConfigSpace 1`.
  - `ConfigSpace 0.7`cannot support newest versions of Numpy `2.+`
  - Cannot install older versions of Numpy (typically `1.24`) on `Python >= 3.12`
  - So, this extra is installed only on `Python < 3.12`
- extras ` dehb, bohb, hebo`: requires either `ConfigSpace < 1` or `numpy < 1.25`, and all of these dependencies cannot be installed on `Python >= 3.12`. So, these extras are installed only on `Python < 3.12`.
- extra `profet`: requires `emukit` and `GPy`, which cannot be installed on Python `>= 3.13`. So this extra is installed only on `Python < 3.13`.
- extra `ray`: right now, cannot be installed on `Python 3.14`. So, this extra is installed only on `Python < 3.14`

## Small fixes in `src`:

- algo `mofa`: use `numpy.random.default_rng()` instead of `RandomState` on newer Numpy versions.
- other small changes

## Fix unit tests:

- Deactivate some tests based on Python versions
- remove `pytest-lazy-fixture` and adapt tests which used this module
- in `test_hebo`: replace `from pymoo.factory import get_algorithm_options` with a manual algorithm finder function to support pymoo `0,6`
- other text fix changes

## Update versioneer

`versioneer` is updated to newest version.


src/orion/algo/mofa/mofa.py
```diff
@@ -28,6 +28,9 @@
 logger = logging.getLogger(__name__)
 
 
+_NUMPY_IS_OLD = version.parse(np.__version__) < version.parse("1.17.0")
```
bouthilx: Do we need backwardcompatibility or is default_rng supported in past versions?

If not, 1.17.0 is 5yrs old so I think it would be fair to remove support of it if necessary.

tests/unittests/algo/test_dehb.py
```diff
@@ -109,16 +109,24 @@ def test_is_done_max_trials(self, phase: TestPhase):
...
-                self.observe_trials(trials, algo, objective)
+                self.observe_trials(trials, algo, rng)
                 objective += len(trials)
 
         # Hyperband should ignore max trials.
         assert algo.n_observed > MAX_TRIALS
         assert algo.is_done
+
+    @pytest.mark.skip(reason="fail on: assert algo.n_observed == n_observed_trials")
+    def test_state_dict(self, seed: int, phase: TestPhase):
```
bouthilx, line 128: There should be an issue for this.

tests/unittests/client/test_runner.py
```diff
@@ -606,22 +607,29 @@ def set_is_done():
...
-    p = Process(target=get_result, args=(queue,))
+    p = Process(target=_get_result, args=(queue,))
     p.start()
     p.join()
 
     assert queue.get() == 0
     assert p.exitcode == 0
 
 
+@pytest.mark.skipif(
```
bouthilx, line 627: There should be an issue for this as well. If this leads to zombie process it's a major problem.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{38,39,310}-{linux,macos}{,-all}
+envlist = py{38,39,310,311,312,313,314}-{linux,macos}{,-all}
```
bouthilx: We can drop 3.8 and 3.9: https://devguide.python.org/versions/.

bouthilx:
For versioneer, did you make the changes manually or you updated it by doing a reinstall from the latest version of versioneer?

tox.ini
notoraptor: Done ! https://github.com/Epistimio/orion/pull/1136/commits/17558bcfccf7b400f6b085300a12d483f681c82c

src/orion/algo/mofa/mofa.py
notoraptor: I think the minimum requred version for numpy must be 1.24, when all dependencies are solved, so we could indeed drop backward compatibility.

Done here: https://github.com/Epistimio/orion/pull/1136/commits/6a83a13aeb4697ce7c0eda4a9c3401aa8cb3402f

tests/unittests/algo/test_dehb.py
notoraptor: Done here: https://github.com/Epistimio/orion/issues/1137

tests/unittests/client/test_runner.py
notoraptor: Done here: https://github.com/Epistimio/orion/issues/1138

notoraptor:
Hello @bouthilx ! Tests pass again ! There are 5 new commits.

About versioneer, I updated it with a reinstall, following instructions here: https://github.com/python-versioneer/python-versioneer (`pip install versioneer` then `versioneer install --vendor`). Note that the latter command always appends `from . import _version; __version__ = _version.get_versions()["version"]` in `src/orion/core/__init__.py`, but this code already exists in the file, so I remove the newly added lines manually.

bouthilx (approved):

bouthilx:
Looks good to me. When we'll make a new release, we'll need to state clearly which algorithms are not supported on more recent python versions.

